PR #19, opened 2023-01-02
VideoToolBox decoder/encoder for MacOS & iOS

theomonnom:
- Created `livekit::VideoEncoderFactory` to use multiple factories at the same time.
We'll add future hw encoders here
- Added VideoToolBox


theomonnom:
Compilation works fine when cross compiling from my M1 🤔

PR #22, opened 2023-01-04
CRT static: Use flag instead of .config

theomonnom:
Don't force the user to use +crt-static inside .config


theomonnom:
Need to resolve issues with duplicated symbols before merging

PR #39, opened 2023-02-07
RtpTransceiver bindings & requirements for publishing tracks

theomonnom:
- RtpSender
- RtpReceiver
- RtpTransceiver
- RtpParameters
- VideoFrameBuilder
- Interior mutability on c++ side
- Cleanup bindings ( Use #pragma once instead of include guards )
     - webrtc-sys/src/* headers are now used in only one place 
     - Avoid confusion between generated headers and our headers
- AdaptedVideoTrackSource
- Cleanup the workaround with unsupported Vec<Shared<T>>
    - Put everything inside one file


theomonnom:
This PR is becoming large, I'll finish the safe bindings and do the PublishTrack logic in another PR

PR #72, opened 2023-05-16
fix: different audio frame size than 10ms

theomonnom:
libwebrtc (audio sources) doesn't seem to work with other frame sizes?
otherwise the publication fails, seems like the packets are ignored


davidzhao (approved):
great catch! this makes sense why opus publishing from files didnt work reliably

PR #88, opened 2023-06-12
feat: add ffi datachannel & mute events

theomonnom:
- Simplify the FFI (we don't need to keep the participant and the room states on the foreign language when using streams)
- Improve FFI :
    - Publish/Receive data
    - ~~Manual subscriptions~~ Another PR?
    - Mute events
- Expose video frame timestamps to NativeVideoSource
- Move AudioSourceOptions to the AudioSource
- Move VideoResolution to the VideoSource 
- RtcVideoSource, RtcAudioSource are now enums exposing the different sources from all platforms

We should maybe expose the video sources from any RtcTrack in the future (livekit-webrtc)


livekit-ffi/protocol/track.proto
```diff
@@ -7,21 +7,10 @@ import "handle.proto";
...
-
-message AudioCaptureOptions {
-  bool echo_cancellation = 1;
-  bool noise_suppression = 2;
-  bool auto_gain_control = 3;
-}
-
 // Create a new VideoTrack from a VideoSource
 message CreateVideoTrackRequest {
   string name = 1;
```
davidzhao, line 12: is this a required field? can a user leave it blank?

davidzhao (approved):
awesome!!

livekit-ffi/protocol/track.proto
theomonnom: It is not required by the FFI server. At the moment, the name can be blank. Maybe it is better to default it to smth like `unnamed track`?

davidzhao: blank seems fine :) we have this behavior for all of our clients. if we do set something else as a default, would probably need to coordinate so they are the same

theomonnom:
Note: this PR is merged but the Mute events are not yet exposed to the FFI and is not forwarded to participant events

PR #96, opened 2023-06-24
feat: mediadevices

theomonnom:
|            | MacOS | Windows | Linux | iOS | Android |
|------------|-------|---------|-------|-----|---------|
| Camera     |       |         |       |     |         |
| Screen     |       |         |       |     |         |
| Microphone |       |         |       |     |         |


theomonnom:
Seems like we can use `audio_device_data_observer.h` instead of wrapping our own AudioDevice?

PR #99, opened 2023-06-26
feat: support webhooks for livekit-api

theomonnom:
`livekit.serde.rs` is huge 🤯

- webhooks support + example 
- fix token verifications


theomonnom:
Needs to be tested before merging

PR #125, opened 2023-07-12
feat: warn blocking callbacks

theomonnom:
- reduce lock times with scopes on some requests
- support tokio-console for tasks debugging on the ffi with the `tracing` feature flag 

This will print an error on the console if we detect a blocking callback. (Default to 2s)


davidzhao (approved):
awesome! this should make it much easier to debug

PR #126, opened 2023-07-12
ci: android & ios builds

theomonnom:
- android & ios builds
- fix win-aarch64 builds
- fix linux aarch64 builds. using QEMU, so this is slow :( 

EDIT: Now using` buildjet-4vcpu-ubuntu-2204-arm `for arm builds, (still using docker so this is still compatible with default GHA runners)


davidzhao:
WOOHOO 🚀

PR #135, opened 2023-07-18
fix: undefined symbols for simulcast adapter

theomonnom:
I initially decided to go with:
`call lib.exe /OUT:"%ARTIFACTS_DIR%\lib\webrtc.lib" %OUTPUT_DIR%\obj\webrtc.lib %OUTPUT_DIR%\obj\media\rtc_simulcast_encoder_adapter\simulcast_encoder_adapter.obj`

but decided to pick a cleaner patch from https://github.com/shiguredo-webrtc-build/webrtc-build/blob/master/patches/add_deps.patch

my previous [commit](https://github.com/livekit/client-sdk-rust/commit/6f5b45690efc7bbdba4d1ae973f67ff17fcbafa8) was not sufficient


davidzhao (approved):
lg!

PR #136, opened 2023-07-22
feat: debounce publisher negotiations

theomonnom:
- debounce publisher negotiations
- fix panics  on livekit-webrtc 
   - for e,g, it can happens if the PeerConnection is closed, so the receivers of async functions get dropped/cancelled)
   - so it was a wrong assumption to unwrap the receivers


livekit/src/room/participant/local_participant.rs
```diff
@@ -161,7 +161,7 @@ impl LocalParticipant {
         tokio::spawn({
             let rtc_engine = self.inner.rtc_engine.clone();
             async move {
-                let _ = rtc_engine.negotiate_publisher().await;
+                let _ = rtc_engine.publisher_negotiation_needed().await;
```
davidzhao: is this notification also async? would this wait until the negotiation completes?

davidzhao (approved):
lg! though I don't think I follow the nuance of the locking changes. will have to read up on this.

livekit/src/room/participant/local_participant.rs
theomonnom: No this function doesn't wait the negotiation, the negotiation is always async because it is now debounced.
Should all calls to this function wait for the result?

theomonnom: `publisher_negotiation_needed` is used to notify the rtc_engine that a negotiation is required to apply new changes, and we can't wait for it, everything happens transparently inside the rtc_session.

EDIT: `ensure_publisher_connected` is used if we really need to know if we're connected/negotiation is finished

theomonnom: Just pushed a new commit, maybe this is easier to understand now

davidzhao: yeah I don't think it should wait for the result. in the publishing track, that's the only case it should wait for the negotiation to be completed.

PR #141, opened 2023-07-24
feat: ffi improvements

theomonnom:
- the ffi client can now communicate more with the provided handles (This avoid to maintain states on the client side)
   - e.g: we can just pass track_handle instead of room_sid + participant_sid + track_sid
   - the client implementation can now be more "stateless"
- added more type safety
   - also refactored livekit id types (ParticipantSid, ParticipantIdentity, RoomSid, TrackSid)
   - improved safety by checking the prefixes
- FfiHandle trait (only the types that implement it can be stored inside the server)
   - good practice because it reduces implementation mistakes
- new proto message: FfiOwnedHandle, we now use u64 vs FfiOwnedHandle to know which handle the client owns (this was confusing before)
   - for good practice, the FfiOwnedHandle must be explicitly created on the Rust side 
- added more comments on the protocol
- reduced code duplication when retrieving an handle
- added more "Ffi" wrappers over livekit objects


theomonnom:
FYI: This PR contains breaking changes on the livekit API
(Only about some type changes)

PR #148, opened 2023-08-02
fix: more ffi client synchronisation

theomonnom:


livekit/src/room/mod.rs
```diff
@@ -109,7 +109,12 @@ pub enum RoomEvent {
         participant: RemoteParticipant,
     },
     ConnectionStateChanged(ConnectionState),
-    Connected,
+    Connected {
+        /// Initial participants & their tracks prior to joining the room
```
theomonnom, line 113: relevant code is here

theomonnom:
I guess the server-sdk-go should have a similar issue?

davidzhao:
yes, room.participants must be populated via the FFI layer.. so that FFI clients are able to access it.

there is some tricky synchronization issues in terms of when the event fires. what do you see the issue to be with the Go SDK?

PR #156, opened 2023-08-16
ci: improved ffi builds 

theomonnom:
- explicit deployment target for macos
- better library naming
- use [manylinux](https://github.com/pypa/manylinux) (increase linux compatibility). good practice from the Python community
  - currently using `manylinux2014`
  
  
 related PR was #147 but it was not sufficient


davidzhao (approved):
lgtm

PR #160, opened 2023-08-16
fix: portable ffi download script

theomonnom:
shell script & curl aren't ideal for Windows
let's use Python instead


davidzhao (approved):
yay! python > bash

PR #164, opened 2023-08-21
ffi protocol improvements 

theomonnom:
- allow to capture audio data without copy
- added play_from_disk example
- separate the info and owned messages 
- to_i420 destination (to avoid unnecessary allocations) 

breaking changes: 
 - capture_frames is now async and automatically handle the clock
 - NativeAudioSource::new now requires the sample_rate and the num_channels

### DIFF
--- a/examples/play_from_disk/src/main.rs
+++ b/examples/play_from_disk/src/main.rs
@@ -1,0 +1,92 @@
+use std::io;
+use thiserror::Error;
+use tokio::io::{AsyncRead, AsyncReadExt};
+
+#[derive(Debug, Error)]
+enum WavError {
+    #[error("Invalid header: {0}")]
+    InvalidHeader(&'static str),
+    #[error("IO error: {0}")]
+    Io(#[from] io::Error),
+}
+
+struct WavReader<R: AsyncRead + Unpin> {
+    reader: R,
+}
+
+#[derive(Debug)]
+struct WavHeader {
+    file_size: u32,
+    format: String,
+    format_length: u32,
+    format_type: u16,
+    num_channels: u16,
+    sample_rate: u32,
+    byte_rate: u32,
+    block_align: u16,
+    bits_per_sample: u16,
+}
+
+impl<R: AsyncRead + Unpin> WavReader<R> {
+    pub fn new(reader: R) -> Self {
+        Self { reader }
+    }
+
+    async fn read_header(&mut self) -> Result<WavHeader, WavError> {
+        let mut header = [0u8; 4];
+        let mut format = [0u8; 4];
+        let mut chunk_marker = [0u8; 4];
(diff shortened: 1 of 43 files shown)


examples/play_from_disk/src/main.rs
```diff
@@ -0,0 +1,190 @@
...
+            let ms = FRAME_DURATION.as_millis() as u32;
+            let num_samples = (header.sample_rate / 1000 * ms) as usize;
+
+            log::info!("sample_rate: {}", header.sample_rate);
+            log::info!("num_channels: {}", header.num_channels);
+            log::info!("max samples: {}", max_samples);
+            log::info!("chunk size: {}ms - {} samples", ms, num_samples);
+
+            let mut written_samples = 0;
+            while written_samples < max_samples {
```
davidzhao, line 162: this looks great! so much simpler to do now

livekit-webrtc/src/native/audio_source.rs
```diff
@@ -73,52 +76,195 @@ impl NativeAudioSource {
...
+
+        let mut inner = self.inner.lock().await;
+        let mut interval = interval(Duration::from_millis(10));
+
+        loop {
+            let Some(data) = self.next_frame(&mut inner, frame) else {
+                break;
             };
 
+            interval.tick().await;
```
davidzhao, line 142: would there be clock drift issues here? i.e.
* 100ns spent in processing (between on_captured_frame and next_frame)
* sleeps for 10ms
* repeat 10 times, audio would be behind by 1ms

davidzhao (commented):
Awesome work!! just one comment re: sync.

livekit-webrtc/src/native/audio_source.rs
theomonnom: So I now keep the same intervals between capture_frame calls, I think the tokio timer precision should be good enough for 10ms

examples/play_from_disk/src/main.rs
```diff
@@ -146,7 +146,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     tokio::spawn({
         let room = room.clone();
         async move {
-            const FRAME_DURATION: Duration = Duration::from_millis(1000); // Write 1s of audio at a time
+            const FRAME_DURATION: Duration = Duration::from_millis(10); // Write 1s of audio at a time
```
davidzhao: change comment as well?

livekit-webrtc/src/native/audio_source.rs
davidzhao: sorry, I misunderstood the `tick` part. understood now that you are keeping the same interval obj.

davidzhao (approved):
lgtm!

PR #165, opened 2023-08-21
fix: macosx_deployment_target to 10.15

gustavo-lighttwist:
I was getting a compilation issue on macOS 13.5 that was only resolved after changing the minimum deployment target version to >=10.13.
There seems to be some memory alignment features that are only available after 10.13.


theomonnom (approved):
Hey, thank you for the PR, really appreciate it :) 
I upgraded the version to 10.15 to match the [Swift SDK](https://github.com/livekit/client-sdk-swift/blob/2f18def0ba3053158cf42bfa130e6eac5b6a58cd/Package.swift#L9).
I'll release new webrtc builds ASAP

PR #167, opened 2023-08-29
add e2ee deps to webrtc patches

theomonnom:


theomonnom:
Just doing it so I can start new webrtc builds on the main branch

PR #178, opened 2023-09-06
fix: wrong ice_restart value

theomonnom:


livekit/src/rtc_engine/peer_transport.rs
```diff
@@ -144,24 +144,24 @@ impl PeerTransport {
...
-            if options.ice_restart {
-                if let Some(remote_description) = self.peer_connection.current_remote_description()
-                {
-                    self.peer_connection
-                        .set_remote_description(remote_description)
-                        .await?;
-                } else {
-                    error!("trying to restart ICE when the pc doesn't have remote description");
-                }
+            let remote_sdp = self.peer_connection.current_remote_description();
```
davidzhao, line 152: I'm not sure that I understand this.. why do we need to do this?

PR #180, opened 2023-09-07
fix: cancelled negotation

theomonnom:
One of the two channel was dropped, and in a wrong timing, the dropped one was selected


davidzhao (approved):
👍

PR #184, opened 2023-09-08
feat: add eos events

theomonnom:
This will allow us to make sure we're not leaking any memory by ignoring some ffi events while closing too early

### DIFF
--- a/livekit-ffi/protocol/audio_frame.proto
+++ b/livekit-ffi/protocol/audio_frame.proto
@@ -113,13 +113,18 @@ message OwnedAudioStream {
 }
 
 message AudioStreamEvent {
-  uint64 source_handle = 1;
-  oneof message { AudioFrameReceived frame_received = 2; }
+  uint64 stream_handle = 1;
+  oneof message { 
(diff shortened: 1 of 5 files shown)


livekit-ffi/protocol/audio_frame.proto
```diff
@@ -113,14 +113,19 @@ message OwnedAudioStream {
 }
 
 message AudioStreamEvent {
-  uint64 source_handle = 1;
-  oneof message { AudioFrameReceived frame_received = 2; }
+  uint64 stream_handle = 1;
+  oneof message { 
+    AudioFrameReceived frame_received = 2;
+    AudioStreamEOS eos = 3;
```
davidzhao, line 119: this works, but you can also consider to use a bool as a marker. with oneof messages, the position/field name is sufficient to mark special types.

davidzhao (approved):

livekit-ffi/protocol/audio_frame.proto
theomonnom: TIL, I think it's ok for now

PR #186, opened 2023-09-11
feat: automatically switch to ws protocol when using http 

theomonnom:


livekit-api/src/signal_client/signal_stream.rs
```diff
@@ -54,13 +54,20 @@ impl SignalStream {
...
-        url: url::Url,
+        mut url: url::Url,
     ) -> SignalResult<(
         Self,
         mpsc::UnboundedReceiver<Box<proto::signal_response::Message>>,
     )> {
         log::info!("connecting to SignalClient: {}", url);
 
+        // Automatically switch to websocket scheme when using http
+        if url.scheme() == "https" {
```
davidzhao, line 65: is it better to copy the URL instead of mutating the input?

davidzhao (approved):

livekit-api/src/signal_client/signal_stream.rs
theomonnom: the url is moved, so we can edit it without impacting the caller

theomonnom: edit* not copy

PR #187, opened 2023-09-15
fix: publisher migration failures

theomonnom:
- fix publisher offer being sent before the SyncState (This caused the migration to fail/incorrect state)
- ~~add RtcConfiguration to the FFI server~~ another PR
- wait for subscriber pc on initial connection

### DIFF
--- a/livekit/src/room/mod.rs
+++ b/livekit/src/room/mod.rs
@@ -491,12 +491,12 @@ impl RoomSession {
                     )))?;
                 }
             }
-            EngineEvent::Resuming(tx) => self.handle_resuming(tx).await,
-            EngineEvent::Resumed(tx) => self.handle_resumed(tx).await,
-            EngineEvent::SignalResumed(tx) => self.handle_signal_resumed(tx).await,
-            EngineEvent::Restarting(tx) => self.handle_restarting(tx).await,
-            EngineEvent::Restarted(tx) => self.handle_restarted(tx).await,
-            EngineEvent::SignalRestarted(tx) => self.handle_signal_restarted(tx).await,
+            EngineEvent::Resuming(tx) => self.handle_resuming(tx),
+            EngineEvent::Resumed(tx) => self.handle_resumed(tx),
+            EngineEvent::SignalResumed(tx) => self.handle_signal_resumed(tx),
+            EngineEvent::Restarting(tx) => self.handle_restarting(tx),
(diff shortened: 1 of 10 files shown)


livekit/src/rtc_engine/peer_transport.rs
```diff
@@ -59,10 +59,7 @@ impl PeerTransport {
     }
 
     pub fn is_connected(&self) -> bool {
-        matches!(
-            self.peer_connection.ice_connection_state(),
-            IceConnectionState::Connected | IceConnectionState::Completed
-        )
+        self.peer_connection.connection_state() == PeerConnectionState::Connected
```
theomonnom, line 62: Now checking directly connection_state

livekit/src/room/mod.rs
```diff
@@ -740,6 +792,8 @@ impl RoomSession {
 
         // unpublish & republish tracks
         let published_tracks = self.local_participant.tracks();
+
+        // Should I create a new task?
```
davidzhao, line 756: what's the intent behind creating a new task?

livekit/src/rtc_engine/mod.rs
```diff
@@ -375,7 +451,7 @@ impl EngineInner {
     ) -> EngineResult<()> {
         let mut running_handle = self.running_handle.write().await;
         if running_handle.is_some() {
-            unreachable!("engine is already connected");
+            panic!("engine is already connected");
```
davidzhao, line 384: is this the common way of error handling in Rust? would this crash the program?

davidzhao (approved):
lgtm!

livekit/src/rtc_engine/mod.rs
theomonnom: Yes, it's more like an assert here, this should never happens, gonna change the panic to assert

livekit/src/room/mod.rs
theomonnom: The question is more like: do we want to wait for the track to be republished before finishing the reconnection logic? (Calling the reconnect event)

theomonnom:
Now using channels instead of callbacks. Because in bad timing where the callbacks were not yet configured, we could have skipped certain events.

Channels will just buffer these events

livekit/src/room/mod.rs
theomonnom: discussed in Slack

davidzhao (approved):

PR #193, opened 2023-09-22
fix: support non-utf8 std::string which can come e.g. from Windows message errors

leanmendoza:
Fix https://github.com/livekit/client-sdk-rust/issues/192


theomonnom (approved):
Thanks for the fix!

PR #198, opened 2023-09-23
feat: room eos

theomonnom:


livekit-ffi/protocol/room.proto
```diff
@@ -210,6 +210,7 @@ message RoomEvent {
     Reconnecting reconnecting = 19;
     Reconnected reconnected = 20;
     E2eeStateChanged e2ee_state_changed = 21;
+    RoomEOS eos = 22;
```
davidzhao: what does it mean to be eos?

davidzhao (approved):

livekit-ffi/protocol/room.proto
theomonnom: that the stream of events is finished. so we can stop listening

theomonnom: happy to hear better names

PR #203, opened 2023-09-26
fix: tasks being stuck while closing the PC

theomonnom:
- It seems that SetRemoteDescription gets into a stuck state if we try to close the peer connections at the same time?

cc @cloudwebrtc


davidzhao (approved):

cloudwebrtc (approved):
lgtm

PR #218, opened 2023-10-05
feat: rtc stats

theomonnom:
expose stats to the users will be on another PR.
keeping the bridge simple by serializing the stats to JSON on cpp and using serde to deserialize on the Rust side


davidzhao (approved):
lgtm! 🚀 🚀

PR #245, opened 2023-11-14
feat: initial connection retry

theomonnom:


livekit/src/room/mod.rs
```diff
@@ -188,6 +189,7 @@ impl Default for RoomOptions {
                 continual_gathering_policy: ContinualGatheringPolicy::GatherContinually,
                 ice_transport_type: IceTransportsType::All,
             },
+            join_retries: 3,
```
theomonnom, line 192: Default to 3 retries (4 total conn)

davidzhao (approved):
lg!

Issue #250, opened 2023-11-29
Allow to create ParticipantSid from String in user module


trzeciak:
Hey, I want test/use this livekit-rust-sdk in c++ code.
To do this I create internal-rust-layer to expose some functionality, and extract it to c++.
I came across a problem to create ParticipantSid from String in user module,
therefore i purpose to add 'pub' in declaration of this type:

```rust
pub struct ParticipantSid(String);
```

```rust
pub struct ParticipantSid(pub String);
```

Do you think it is ok, or I should choose other way?

theomonnom:
Answered in https://github.com/livekit/rust-sdks/pull/251

PR #251, opened 2023-11-29
Allow to create ParticipantSid from String in user module (#250)

trzeciak:
Hey, I want test/use this livekit-rust-sdk in c++ code.
To do this I create internal-rust-layer to expose some functionality, and extract it to c++.
I came across a problem to create ParticipantSid from String in user module,
therefore i purpose to add 'pub' in declaration of ParticipantSid.
Do you think it is ok, or I should choose other way?


trzeciak:
Thanks for the explanation, I will follow it 👍

PR #253, opened 2023-11-29
allow other ffi integrations

theomonnom:
- added callback_fn, so it can be used for NodeJS integration


markgrossnickle:
@theomonnom hey, I'm working on updating the Unity LiveKit sdk and am curious about this pr where InitializeRequest was removed. The comments above still state you should start with InitializeRequest and pass in the callback method. Since that is removed here, how are we supposed to pass in the callback method?

PR #257, opened 2023-12-04
rename rgba conversions

theomonnom:
So we're not following FourCC as it is confusing for users?
But instead we follow the logic from OpenCV etc

### DIFF
--- a/livekit-ffi/src/server/requests.rs
+++ b/livekit-ffi/src/server/requests.rs
@@ -122,8 +122,10 @@ fn on_set_subscribed(
         server.retrieve_handle::<FfiPublication>(set_subscribed.publication_handle)?;
 
     let TrackPublication::Remote(publication) = &ffi_publication.publication else {
-            return Err(FfiError::InvalidRequest("publication is not a RemotePublication".into()));
-        };
+        return Err(FfiError::InvalidRequest(
+            "publication is not a RemotePublication".into(),
+        ));
+    };
 
     let _guard = server.async_runtime.enter();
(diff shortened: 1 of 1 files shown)


keepingitneil (approved):

livekit-ffi/src/server/requests.rs
```diff
@@ -470,19 +472,19 @@ unsafe fn on_to_argb(
             match rgba_format {
                 proto::VideoFormatType::FormatArgb => {
                     #[rustfmt::skip]
-                    yuv_helper::i420_to_argb(src_y, yuv.stride_y, src_u, yuv.stride_u, src_v, yuv.stride_v, argb, stride, w, h);
+                    yuv_helper::i420_to_bgra(src_y, yuv.stride_y, src_u, yuv.stride_u, src_v, yuv.stride_v, argb, stride, w, h);
```
davidzhao, line 475: I'm not understanding this, isn't this mismatched from the indicated format?

theomonnom: FourCC are defined in BigEndian https://en.wikipedia.org/wiki/FourCC#Compiler_support (It is just the naming convention)
So on little endian system ARGB is indeed BGRA in memory

theomonnom: Also more details here https://chromium.googlesource.com/libyuv/libyuv/+/refs/heads/main/docs/formats.md#the-argb-fourcc

davidzhao: ah got it. thanks!

davidzhao (approved):

PR #258, opened 2023-12-04
capture blank frames on AudioSource & VideoSource

theomonnom:


libwebrtc/src/video_source.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 LiveKit, Inc.
+// Copyright 2024 LiveKit, Inc.
```
davidzhao, line 1: damn already 2024 for you? 😉

davidzhao (approved):

libwebrtc/src/video_source.rs
theomonnom: I'm a time traveler ⏳

PR #263, opened 2023-12-08
feat: add svc codecs

theomonnom:


examples/wgpu_room/src/logo_track.rs
```diff
@@ -76,6 +77,7 @@ impl LogoTrack {
                 LocalTrack::Video(track.clone()),
                 TrackPublishOptions {
                     source: TrackSource::Camera,
+                    video_codec: VideoCodec::AV1,
```
davidzhao: maybe default to vp9? since some browsers/devices do not support AV1

davidzhao (approved):
awesome work! thanks!

PR #272, opened 2023-12-19
send silence when no data is captured

theomonnom:


libwebrtc/src/native/audio_source.rs
```diff
@@ -78,8 +78,10 @@ impl NativeAudioSource {
                     interval.tick().await;
 
                     let inner = source.inner.lock().await;
-                    if inner.captured_frames > 0 {
-                        break; // User captured something, stop injecting silence
+                    if let Some(last_capture) = inner.last_capture {
+                        if last_capture.elapsed() < Duration::from_millis(20) {
```
theomonnom: 20ms instead of 10?
Not sure if this is the best way to handle it but it avoid to inject silence if the user is 1ms late

libwebrtc/src/native/audio_source.rs
```diff
@@ -78,8 +78,10 @@ impl NativeAudioSource {
...
+                            break;
```
davidzhao: should this be continue? for the next run?

theomonnom: right, this is a mistake, will fix now

libwebrtc/src/native/audio_source.rs
davidzhao: 20ms seems fine. we would inject 10ms worth of silence? 

we should have a break condition still somewhere right? when the source is destroyed?

theomonnom: yes 10ms of silence every 10ms
blah I'm tired.. yeah I should stop the task

keepingitneil: I think it makes sense but I don't know enough about what this means for jitter buffer behavior. Basically every time you stop sending audio the pipeline will have 10ms of jitter introduced?

edit: nvm, so 10ms is introduced and then it immediately catches back up to whatever it considers healthy buffer size becuase we'll just send 10ms instantly after that.

keepingitneil (approved):

davidzhao (approved):

PR #279, opened 2023-12-31
Add rustfmt.toml configuration and re-format everything + fix basic warnings

daniel-abramov:
So that contributions from multiple people are more consistent w.r.t. code style and formatting 🙂 

Note, that this one does not include all of the `cargo clippy` fixes since there are quite a few of them that require manual intervention. I did not want to make them part of this PR (since unlike the rest of the changes these would not be automated, hence would require a more rigorous review of whether the user-written code changes anything).

Also, `cargo check` still produces 3 warnings that I did not fix as they would need some minor changes:
- `RemoteTrackPublication::on_subscription_status_changed` and `RemoteTrackPublication::on_permission_status_changed` are `pub(crate)` and never used within a crate. I was not sure if I'm allowed to remove them (since we don't use them) or if I need to make them `pub` (that would require making the rest of the things `pub` as well, for consistency).
- `RtcSession::publisher()` is `pub`, but never used (not sure why it complains about it given that it is `pub`, perhaps it's never exported or something).

Also I added a tiny simple job to check the formatting so that we can keep it consistent.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,36 @@
+name: Rust Formatting
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  formatting:
+    name: Check Formatting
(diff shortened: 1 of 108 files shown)


theomonnom (approved):
Lgtm, TIL about small heuristics

theomonnom:
Thanks 🙏 
I’ll try to address the other warnings

PR #285, opened 2024-01-13
fix: webhook body serde decode error due to missing field

### LINKED ISSUE #284: webhook body decoding error
I randomly started getting this webhook error 

code block the error is coming from :
```
let event = match webhook_receiver.receive(body, jwt) {
            Ok(i) => i,
            Err(e) => return Resp::InternalServerError().json(ServerMsg::error(e.to_string())),
        };
```

error message

```
"invalid body, failed to decode: unknown field `version`, expected one of `sid`, `name`, `empty_timeout`, `emptyTimeout`, `max_participants`, `maxParticipants`, `creation_time`, `creationTime`, `turn_password`, `turnPassword`, `enabled_codecs`, `enabledCodecs`, `metadata`, `num_participants`, `numParticipants`, `num_publishers`, `numPublishers`, `active_recording`, `activeRecording` at line 1 column 259"
```
image

Possible Place to fix error: https://github.com/livekit/rust-sdks/blob/main/livekit-protocol/src/livekit.rs#L3175

cs50victor:
Closes #284

### DIFF
--- a/livekit-protocol/src/livekit.rs
+++ b/livekit-protocol/src/livekit.rs
@@ -24,6 +24,8 @@ pub struct Room {
     pub num_publishers: u32,
     #[prost(bool, tag="10")]
     pub active_recording: bool,
+    #[prost(message, optional, tag="13")]
+    pub version: ::core::option::Option<TimedVersion>,
 }
 #[allow(clippy::derive_partial_eq_without_eq)]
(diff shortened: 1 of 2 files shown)


theomonnom:
The right solution is to ignore unknown fields. The protocol could change at anytime
Thanks for reporting the issue

PR #286, opened 2024-01-13
chore: add docs to make it easier to contribute

cs50victor:
Quick brain dump of what i had to figure out to put up this #285


davidzhao:
Thanks for putting this together! Completely agreed we need some work on making this clear.

cs50victor:
Do i merge this as a starting point and add more details in future PRs or would you want a more detailed initial doc before merging?

PR #290, opened 2024-01-15
ignore unknown fields on webhook receiver & update proto

theomonnom:


theomonnom:
Closes https://github.com/livekit/rust-sdks/issues/284 & https://github.com/livekit/rust-sdks/pull/285

PR #319, opened 2024-04-09
New FFI for data packets. Support receiving DTMF.

dennwc:
This is a breaking change, because I had to disable the `DataReceived` message. Otherwise caller will get two messages with the same data (old and new protocol). I hope removing all fields is enough to trigger an error on the target language SDK side during build, so that the author can update it to using `DataPacketReceived` message.

### DIFF
--- a/livekit-ffi/protocol/room.proto
+++ b/livekit-ffi/protocol/room.proto
@@ -260,6 +260,7 @@ message RoomEvent {
     Reconnected reconnected = 23;
     E2eeStateChanged e2ee_state_changed = 24;
     RoomEOS eos = 25; // The stream of room events has ended
+    DataPacketReceived data_packet_received = 26;
   }
 }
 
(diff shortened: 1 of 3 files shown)


livekit-ffi/protocol/room.proto
```diff
@@ -252,14 +252,15 @@ message RoomEvent {
     ParticipantMetadataChanged participant_metadata_changed = 15;
     ParticipantNameChanged participant_name_changed = 16;
     ConnectionQualityChanged connection_quality_changed = 17;
-    DataReceived data_received = 18;
+    DataReceived data_received = 18 [deprecated=true];
```
theomonnom: Feel free to delete instead of deprecating. We can't break the protocol with the FFI

dennwc: Should I just remove the field or set it as `reserved`?

theomonnom: remove is OK

theomonnom (approved):
lgtm

PR #320, opened 2024-04-11
buffer audio data before waiting inside captures

theomonnom:


libwebrtc/src/native/audio_source.rs
```diff
@@ -218,91 +190,3 @@ impl From<AudioSourceOptions> for sys_at::ffi::AudioSourceOptions {
         }
     }
 }
-
-#[cfg(test)]
-mod tests {
```
davidzhao, line 223: should we update with new tests to confirm behavior?

davidzhao (approved):

PR #339, opened 2024-05-15
Update protocol to v1.16

davidzhao:


theomonnom (approved):

theomonnom (changes requested):
not safe to merge, need to provide default values for new fields. will do asap.

if you have time, just compile the crates and check the errors, should be relatively easy to fix

PR #342, opened 2024-05-27
livekit, livekit-ffi: Update to v2 APIs

nbsp:
highly wip, went through [the migration notes](https://docs.livekit.io/recipes/migrate-from-v1/) and [the `v1.15.7..v2.0.0` commit log](https://github.com/livekit/client-sdk-js/commits/v2.0.0) for client-sdk-js and made some API changes. doesn't include any non-API-breaking change from client-sdk-js, nor any API additions from versions later than 2.0.0.

builds fine, but untested, couldn't get any of the examples to work even on main (perhaps my NixOS/Wayland system is confusing)

- [x] `DataPacket` now uses `reliable: bool` and not `kind: DataPacketKind`
- [x] `get_track` → `get_track_publication` etc
- [x] `participants` → `remote_participants`
- [x] ~~`VideoQuality::OFF` removal~~ couldn't find any mentions of `VideoQuality::OFF` here in the realtime SDK
- [x] async `sid()`
- [x] update `Cargo.toml` versions in affected crates


livekit-ffi/protocol/room.proto
```diff
@@ -89,7 +89,7 @@ message PublishDataRequest {
   uint64 data_ptr = 2;
   uint64 data_len = 3;
   DataPacketKind kind = 4;
-  repeated string destination_sids = 5; // destination
```
lukasIO, line 92: also not sure what the right approach here would be, but we could of course "only" deprecate destination_sids for now in the ffi-protocol and introduce `destination_identites=7`

nbsp: would that not be as breaking of a change, as implementations would now require `destination_identities=7` as well?

lukasIO: In my understanding it's best not to re-use the field index, see https://protobuf.dev/programming-guides/proto3/#consequences

nbsp: got it

lukasIO: you could add `repeated string destination_sids = 5 [deprecated=true];` to indicate the deprecation

nbsp:
looking into `pub fn sid(&self)` further i don't think there is much reason for having it asynchronous, as `Room::connect` already `await`s the join response that includes the SID

nbsp:
pushed an async sid() proof of concept, if you could review the general idea while i work out the kinks

- added `sid_rx` and `sid_tx` oneshot channel to RoomSession
- `sid_tx` is wrapped in an `Option` because it. works. and it didn't work without it ¯\_(ツ)_/¯
- `impl Debug for Room` and `impl From<&FfiRoom> for proto::RoomInfo` are both functions dependent on `sid()` that can't be async, dunno how to figure that out

livekit-ffi/src/conversion/room.rs
```diff
@@ -213,8 +213,9 @@ impl From<proto::AudioEncoding> for AudioEncoding {
 }
 
 impl From<&FfiRoom> for proto::RoomInfo {
-    fn from(value: &FfiRoom) -> Self {
+    #[tokio::main]
+    async fn from(value: &FfiRoom) -> Self {
         let room = &value.inner.room;
-        Self { sid: room.sid().into(), name: room.name(), metadata: room.metadata() }
+        Self { sid: room.sid().await.into(), name: room.name(), metadata: room.metadata() }
```
theomonnom: I think we should create a new FfiRequest to get the SID as it is now async

Example of an async request: https://github.com/livekit/rust-sdks/blob/8b276f9d4b98437a139e1cbe41cfda1332ce1120/livekit-ffi/protocol/room.proto#L120

livekit-ffi/src/server/room.rs
```diff
@@ -238,19 +238,21 @@ impl RoomInner {
...
-        let destination_sids = publish.destination_sids;
+        let destination_identities = publish.destination_identities;
         let async_id = server.next_id();
 
         if let Err(err) = self.data_tx.send(FfiDataPacket {
             payload: DataPacket {
                 payload: data.to_vec(), // Avoid copy?
-                kind: kind.into(),
+                reliable: match kind {
+                    proto::DataPacketKind::KindLossy => false,
```
theomonnom: Let's also change the PublishData request to take a boolean as input?
Breaking the protocol is OK across FFI version. (The binaries are shipped with the SDKs, so we can break stuff safely)

livekit/src/room/mod.rs
```diff
@@ -264,18 +263,16 @@ struct RoomInfo {
...
     options: RoomOptions,
     active_speakers: RwLock<Vec<Participant>>,
     local_participant: LocalParticipant,
-    participants: RwLock<(
-        // Keep track of participants by sid and identity
-        HashMap<ParticipantSid, RemoteParticipant>,
-        HashMap<ParticipantIdentity, RemoteParticipant>,
-    )>,
+    remote_participants: RwLock<HashMap<ParticipantIdentity, RemoteParticipant>>,
     e2ee_manager: E2eeManager,
```
theomonnom, line 273: Nice! 😀

livekit/src/room/mod.rs
```diff
@@ -1123,34 +1133,48 @@ impl RoomSession {
...
     fn get_participant_by_identity(
         &self,
         identity: &ParticipantIdentity,
     ) -> Option<RemoteParticipant> {
-        self.participants.read().1.get(identity).cloned()
+        self.remote_participants.read().get(identity).cloned()
+    }
+
+    fn set_sid(&self, sid: RoomSid) {
+        let mut guard = self.sid_tx.blocking_lock();
```
theomonnom: I don't think we should use a blocking_lock as it could easily lead to deadlock/blocking the tokio runtime.
I feel like using a channel adds too much complexity?
What about implementing a Promise-like struct inside the livekit-protocol. So we can just listen to changes to whatever variable. I think this would be cleaner

livekit/src/room/mod.rs
```diff
@@ -409,6 +409,8 @@ impl Room {
             room_task: Default::default(),
         });
 
+        inner.set_sid(room_info.sid.try_into().unwrap());
```
theomonnom: It is OK because we're still using the Protocol 9 https://github.com/livekit/rust-sdks/blob/8b276f9d4b98437a139e1cbe41cfda1332ce1120/livekit-api/src/signal_client/mod.rs#L47.

Otherwise we should keep in mind that the `async sid `method only resolves once it has a valid SID from the server (room update)

livekit/src/room/mod.rs
theomonnom: Maybe just wrapping a [JoinHandle](https://tikv.github.io/doc/tokio/task/struct.JoinHandle.html) that we await inside each call to the `sid() `method is OK.
The handle is basically just waiting for the first Room update

livekit/src/room/mod.rs
```diff
@@ -248,9 +246,10 @@ pub struct Room {
 }
 
 impl Debug for Room {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+    #[tokio::main]
```
theomonnom: This is not valid, it creates a new Tokio runtime each time Debug/Print is called.
Let's not wait for the sid here (we could use the internal state of the Room to see if we already received the sid and print it, otherwise print None)

livekit/src/room/mod.rs
nbsp: how would we detect the first Room update without a signal of some sort? `handle_room_update` needs to ping the spawned process in some way

livekit-ffi/src/conversion/room.rs
nbsp: how would that help this line? `room.sid()` is still asynchronous, so `impl From` has to be async as well

theomonnom: So I think it could also be set to None when we don't have a valid value yet. 
We can then notify the FfiClient once we have a correct SID?
So in this case, not using an async request is OK.  I think we could just send a UpdateRoomEvent to the FfiClient and we can rewrite this logic on the Python side as it seems fairly simple
 -> On NodeJS it would simply be a Promise that resolves on the first UpdateRoom event
 -> On Python we can do the same thing using asyncio.Future

livekit/src/room/mod.rs
theomonnom: the Promise utility could stil use a channel internally, but the idea is to hide the implementation as it adds complexity to the Room struct.

Inside the handle_room_update I could see something like:
```self.sid_promise.resolve(my_sid).await```

File: livekit-ffi/src/conversion/room.rs (hunk 2 above)
Comment (nbsp) [reply livekit-ffi/src/conversion/room.rs]: could we make `inner.sid` an `Option<RoomSid>`, then? `(*Room).sid()` would still return an unwrapped sid when available but i think it makes things easier if we used an option here

Comment (theomonnom) [reply livekit-ffi/src/conversion/room.rs]: Ideally the field will just look like sid: `Promise<RoomSid>`.
I think it's correct, you may need `Option<T>` internally

Comment (nbsp) [reply livekit-ffi/src/conversion/room.rs]: as far as i understand the async Rust equivalent to `Promise<T>` *is* `oneshot::channel()`, so i had to keep `sid_rx`/`sid_tx`, however i streamlined it quite a lot and made it update on `handle_room_update`.

a function i added to Room is `maybe_sid(&self)`, which just returns the raw `Option<RoomSid>`, for instances that can't be synchronous and don't care about what the value is if it can be updated later (e.g. `impl From<&FfiRoom> for proto::RoomInfo`)

Comment (theomonnom) [reply livekit-ffi/src/conversion/room.rs]: I was thinking that we could add a new Promise utility to the livekit-protocol crate:
Something similar to this Go implementation: https://github.com/livekit/protocol/blob/main/utils/promise.go
So we don't have to put sid_rx and sid_tx inside the room implementation, and we can keep things simpler/cleaner there.

Comment (nbsp) [reply livekit-ffi/src/conversion/room.rs]: added `Promise<T>` to protocol.
it doesn't work *exactly* like JS Promises since that's not what we need here, it's mostly just an async wrapper for `tokio::sync::oneshot`, but it is a lot cleaner.

File: livekit-protocol/src/promise.rs [hunk 8]
```diff
@@ -0,0 +1,49 @@
...
+    pub fn new() -> Self {
+        let (tx, rx) = oneshot::channel();
+        Self { tx: Mutex::new(Some(tx)), rx: Mutex::new(Some(rx)), done: Default::default() }
+    }
+
+    pub fn resolve(&self, result: T) -> Result<(), &'static str> {
+        let done = self.done.read().clone();
+        if !done {
+            let _ = self.tx.try_lock().unwrap().take().unwrap().send(result);
+            let mut done = self.done.write();
```
theomonnom: I don't think the `done` field is necessary (tx will be None)

livekit/src/room/mod.rs
```diff
@@ -262,9 +262,8 @@ struct RoomInfo {
 
 pub(crate) struct RoomSession {
     rtc_engine: Arc<RtcEngine>,
-    sid_tx: AsyncMutex<Option<oneshot::Sender<()>>>,
-    sid_rx: AsyncMutex<oneshot::Receiver<()>>,
-    sid: RwLock<Option<RoomSid>>,
+    sid_promise: Promise<RoomSid>,
+    sid: AsyncMutex<Option<RoomSid>>,
```
theomonnom: Can we avoid having the sid field? Ideally it is already contained inside the promise, so we keep things even simpler.

livekit-ffi/src/conversion/room.rs
theomonnom: Nice! I added a few more comments

livekit/src/room/mod.rs
nbsp: probably not, for two reasons:
- the Promise doesn't keep the value stored anywhere, it just returns from recv when ready (though this could be changed to allow for one-send-multiple-recv)
- even if we do that, though, `Promise<AsyncSid>` is too opaque for stuff like `impl Debug` or `impl From` that need to be synchronous. i could add a `try_result()` method to `Promise<T>` but that would be... hacky. might still be better though.

livekit-ffi/protocol/ffi.proto
```diff
@@ -67,17 +67,18 @@ message FfiRequest {
     UpdateLocalNameRequest update_local_name = 10;
     GetSessionStatsRequest get_session_stats = 11;
     PublishTranscriptionRequest publish_transcription = 12;
+    GetRoomSidRequest get_room_sid = 13;
```
theomonnom: Let's use this [way](https://github.com/livekit/rust-sdks/pull/342#discussion_r1619337144) instead of a request.
The Python/NodeSDK will set the sid to "" by default (in reality it defaults to what is inside the first RoomInfo, but we know it'll be ""). And as soon as they receive a UpdateRoomEvent, they will update the field manually.

The drawback I see from the GetRoomSidRequest is that the Python/NodeSDK must have called the GetRoomSidRequest at least once to have the sid info in memory. (Which isn't always the case as this is user defined, we may also have debug logs that aren't async on Node/Python)

livekit/src/room/mod.rs
nbsp: new commit addresses this, lmk if that's better

livekit-ffi/protocol/ffi.proto
nbsp: alright, should be able to just remove all FFI additions, since `maybe_sid()` sends `None` if it's empty, and the SID update triggers on Room update regardless.

theomonnom: I don't think we send an update to the FFI client yet for room updates like the SID, only for the metadata

nbsp: now it should

livekit/src/room/mod.rs
```diff
@@ -116,6 +116,9 @@ pub enum RoomEvent {
         old_metadata: String,
         metadata: String,
     },
+    RoomSidChanged {
```
theomonnom: I don't think we want to expose this event to the Rust users. I know the intent is to use it on the FfiServer and forward this update to the FfiClient.

I think you can `await room.sid()` on the FfiServer and forward the change of sid to the FfiClient once it resolves

nbsp: something like this?
```diff
diff --git a/livekit-ffi/src/server/room.rs b/livekit-ffi/src/server/room.rs
index 6a74fa1..b326710 100644
--- a/livekit-ffi/src/server/room.rs
+++ b/livekit-ffi/src/server/room.rs
@@ -166,6 +166,16 @@ impl FfiRoom {
                         },
                     ));
 
+                    let _ =
+                        server.send_event(proto::ffi_event::Message::RoomEvent(proto::RoomEvent {
+                            room_handle: inner.handle_id,
+                            message: Some(proto::room_event::Message::RoomSidChanged(
+                                proto::RoomSidChanged {
+                                    sid: ffi_room.inner.room.sid().await.into(),
+                                },
+                            )),
+                        }));
+
                     // Forward events
                     let event_handle = server.watch_panic({
                         let close_rx = close_rx.resubscribe();

theomonnom: Yes exactly, but make sure to do it in a separate task to not block the event loop

theomonnom:
Nice, lgtm, have you been able to test the changes?

nbsp:
unfortunately i wasn't able to run any of the sample code in the examples directories, even under the main branch. might have something to do with my linux distro + display server combo, but that's outside the scope of this PR

theomonnom:
Can you try an example that doesn't use any display env? Like compiling livekit_ffi and trying it on Python/Node

nbsp:
> Can you try an example that doesn't use any display env? Like compiling livekit_ffi and trying it on Python/Node

confirmed working with `python-sdks/examples/basic_room.py`, with minor changes to the Participant class (kind → reliable)

theomonnom (approved):
ok nice lgtm!

PR #343, opened 2024-05-28
Experimental example - play ogg from disk

kuba--:
just an experiment (code) - does not work, so far.


theomonnom:
Hey I think using the current AudioSource with a new preencoded parameter isn't a good solution.
 - The format of the data is a Vec<i16>, so if the size of our opus frames are not oven it'll not work. It is also hard from a devex perspective
 - This [code](https://github.com/livekit/rust-sdks/blob/8b276f9d4b98437a139e1cbe41cfda1332ce1120/libwebrtc/src/native/audio_source.rs#L68) assumes the data we're sending is PCM (We're sending data by chunks of 10ms)
 
I think we should create a new structure called EncodedAudioSource that simply takes an array of bytes per capture

PR #350, opened 2024-06-20
Update protocol

dennwc:
Update protocol. Quite a few changes:
- Transcription language moved to a segment. FFI and API updated to reflect it. This removes a field on root transcription.
- Pick up `enable_transcoding` for Ingress.
- Support SIP Inbound/Outbound Trunk API.
- Use SIP grants.
- Handle participant attributes.

I'm not sure the change for attributes is complete. I couldn't find any code that updated `name` or `metadata` based on update events. So not sure where I should add attributes updates :thinking:


keepingitneil (approved):
Not the most familiar with much of this codebase, but the transcription parts (which I am more familiar) lgtm

nbsp (commented):
make sure to bump minor on all affected crates. not familiar enough with the rest of the codebase to do a full review

theomonnom:
> make sure to bump minor on all affected crates. not familiar enough with the rest of the codebase to do a full review

let's not bump for this PR

theomonnom:
> I'm not sure the change for attributes is complete. I couldn't find any code that updated `name` or `metadata` based on update events. So not sure where I should add attributes updates 🤔

Are you talking about the events here: https://github.com/livekit/rust-sdks/blob/9c2a467751f1698532393bee95b96cba71af78b3/livekit/src/room/mod.rs#L119C5-L119C31

dennwc:
These too! Need to create a new one there :+1: 

But also, aren't these events supposed to update `ParticipantInfo` in the SDK? I haven't found any place that updates `metadata` based on received signaling updates.

livekit-api/src/services/sip.rs
```diff
@@ -116,12 +170,81 @@ impl SIPClient {
                     inbound_username: options.inbound_username.to_owned(),
                     inbound_password: options.inbound_password.to_owned(),
                 },
-                self.base.auth_header(VideoGrants { ..Default::default() })?,
+                self.base.auth_header(
+                    VideoGrants { ..Default::default() },
```
theomonnom: Could just be (I know it was already not the case lol). this is nit feel free to ignore
```suggestion
                    Default::default()
```

theomonnom (approved):

theomonnom:
Seems like we update the metadata here
https://github.com/livekit/rust-sdks/blob/9c2a467751f1698532393bee95b96cba71af78b3/livekit/src/room/participant/mod.rs#L141

dennwc:
@theomonnom I added the update handler too. Not sure if it's a good idea to clone the new metadata for the changed callback :thinking:

theomonnom (approved):
Cloning isn't ideal but this is kind of mandatory for this PR
I have some plans to make this SDK sync instead of async and calling callbacks will allow references

PR #353, opened 2024-06-24
Add enable_queue for AudioSourceOptions.

cloudwebrtc:
This parameter is used to avoid queues and silence frames causing latency accumulation, and is typically used when connecting to Audio Capture devices or other real-time sources.

### DIFF
--- a/libwebrtc/src/native/audio_source.rs
+++ b/libwebrtc/src/native/audio_source.rs
@@ -95,17 +95,16 @@ impl NativeAudioSource {
                                             num_channels,
                                             blank_data.len() / num_channels as usize,
                                         );
+
+                                        interval.tick().await;
                                     }
-                                    
-                                    interval.tick().await;
+                                    continue;
                                 }
                                 TryRecvError::Disconnected => {
                                     // channel disconnected, break out of loop
                                     break;
                                 }
                             }
-
-                            break;
                         }
                     }
                 }
--- a/libwebrtc/src/native/audio_source.rs
+++ b/libwebrtc/src/native/audio_source.rs
@@ -71,42 +71,32 @@ impl NativeAudioSource {
                 let mut interval = interval(Duration::from_millis(10));
                 interval.set_missed_tick_behavior(livekit_runtime::MissedTickBehavior::Delay);
                 let blank_data = vec![0; samples_10ms];
-                let send_silent_frames = !source.sys_handle.audio_options().disable_silent_frames;
 
                 loop {
-                    let frame = po_rx.try_recv();
-                    match frame {
-                        Ok(frame) => {
-                            source.sys_handle.on_captured_frame(
-                                &frame,
-                                sample_rate,
-                                num_channels,
-                                frame.len() / num_channels as usize,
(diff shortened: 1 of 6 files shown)


libwebrtc/src/native/audio_source.rs
```diff
@@ -71,32 +71,42 @@ impl NativeAudioSource {
                 let mut interval = interval(Duration::from_millis(10));
                 interval.set_missed_tick_behavior(livekit_runtime::MissedTickBehavior::Delay);
                 let blank_data = vec![0; samples_10ms];
+                let send_silent_frames = !source.sys_handle.audio_options().disable_silent_frames;
 
                 loop {
-                    interval.tick().await;
```
theomonnom: I think we should keep the interval when send_silent_frames is enabled?

Should we rename this feature: enable_queue instead of disable_silent_frames? Thinking about how users will understand it

cloudwebrtc: I moved the interval here:
https://github.com/livekit/rust-sdks/pull/353/files#diff-8ab7d3dca0e75d71ebc9798430e49ba444958ba064fb95ff1e00e7fde1696611R99

Makes sense. If we change the name to `enable_queue`, in the case of false, should we directly skip sending the `frame` to `po_tx`when the audio frame enters `capture_frame`? This way, we can actually skip the `queue` and send the frame to `libwebrtc`.

theomonnom: oh sorry I missed this line.
Yeah we can directly skip the queue, and directly send to libwebrtc

theomonnom (approved):

libwebrtc/src/native/audio_source.rs
```diff
@@ -134,6 +134,7 @@ impl NativeAudioSource {
 
         let mut inner = self.inner.lock().await;
         let mut samples = 0;
+        let enable_queue  = self.sys_handle.audio_options().enable_queue;
```
maxbrunsfeld, line 137: I believe this always returns `false`, because the `enable_queue` option is not initialized explicitly in the C++ implementation of this `audio_options` method.

Does the `enable_queue` bit even need to be passed into C++? Why not just store the `enable_queue` as a field on the rust `NativeAudioSource`?

maxbrunsfeld (commented):
This commit introduced a regression, causing distorted audio in our Rust app. The `enable_queue` option is not actually respected, due to a bug I mentioned above.

/cc @theomonnom @cloudwebrtc

libwebrtc/src/native/audio_source.rs
cloudwebrtc: hey, yes, there is a bug here, I created another PR to fix it
https://github.com/livekit/rust-sdks/pull/360

cloudwebrtc: After modification, `enable_queue` is an optional parameter of `NewAudioSourceRequest` instead of `AudioSourceOptions`. and store the `enable_queue` field on the rust NativeAudioSource.

PR #310, opened 2024-02-15
Wrap Tokio types in a configurable runtime

mikayla-maki:
Adds a thin compatibility layer over the rust runtime ecosystem.

TODO: 
- [x] Add support for async_std compatible runtimes
- [x] Resolve any TODOs
- [x] Get the async_std examples running
- [x] Make sure this works for Zed
- [x] Pass CI
- [ ] ~~Switch from async default to tokio default~~ N/A, CI is currently broken

PR #354, opened 2024-06-25
Avoid direct use of tokio interval in rtc video source

maxbrunsfeld:
Follow-up to https://github.com/livekit/rust-sdks/pull/310

That PR added the ability to use async runtimes other than Tokio, via the `livekit-runtime` adapter crate. But this *one* direct usage of `tokio::time::interval` got left behind.

### DIFF
--- a/libwebrtc/src/native/video_source.rs
+++ b/libwebrtc/src/native/video_source.rs
@@ -18,6 +18,7 @@ use std::{
 };
 
 use cxx::SharedPtr;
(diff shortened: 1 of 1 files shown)


theomonnom (approved):

theomonnom:
Nice, thank you, we also have plan to remove any async runtime into our SDK. To make everything sync. The reason is that the advantages of async in this SDK is very little

PR #355, opened 2024-06-26
Add APIs for converting between native VideoFrameBuffers and their underlying `CVPixelBuffer` on macOS

maxbrunsfeld:
Depends on https://github.com/livekit/rust-sdks/pull/354 (this branch contains that change as well).

### Background

In the [Zed editor](https://github.com/zed-industries/zed), we're switching over from using the LiveKit [Swift SDK](https://github.com/livekit/client-sdk-swift) to this crate.  The main motivation is that we want our LiveKit-based screen-sharing and audio features to work on Linux (and someday Windows), not just macOS.

In order to implement screen sharing efficiently (as it was before, when using the Swift  SDK), we want to pass video frames directly from the `ScreenCaptureKit` framework to this crate's `NativeVideoSource::capture_frame` method, without copying them into intermediate buffers. Similarly, when playing remote video tracks, we want to take decoded video frames directly from the `NativeVideoStream` and use them as GPU textures without performing an extra copy.

### Change

To support both of these use cases, we have added two new macOS-specific APIs to the `video_frame::NativeBuffer` type: `from_cv_pixel_buffer`, and `get_cv_pixel_buffer`. I think it makes sense to have this functionality, since AFAICT, the purpose of the `Native` variant of `VideoFrameBuffer` is to adapt a platform-specific video frame representation into the WebRTC's logic.

### Details

* We haven't attempted to make the APIs type safe. This would drag in a dependency on the macOS `CoreMedia` framework, so we decided to leave the interface in terms of `*mut c_void`.
* The `get_cv_pixel_buffer` **can** return `NULL` if the native buffer is not actually backed by a `CVPixelBuffer`. Rather than implicitly perform some potentially expensive conversion implicitly, we decided to leave that to the caller. The only point of these method is to enable a fast path in which we minimize conversions.

Let me know if there is a different solution to this problem that you'd prefer. Thanks!

### DIFF
--- a/libwebrtc/src/native/video_frame.rs
+++ b/libwebrtc/src/native/video_frame.rs
@@ -148,6 +148,17 @@ macro_rules! impl_to_argb {
 
 #[allow(unused_unsafe)]
 impl NativeBuffer {
+    #[cfg(target_os = "macos")]
+    pub unsafe fn from_cv_pixel_buffer(
+        cv_pixel_buffer: *mut std::ffi::c_void,
+    ) -> vf::native::NativeBuffer {
+        vf::native::NativeBuffer {
+            handle: NativeBuffer {
+                sys_handle: vfb_sys::ffi::new_native_buffer(cv_pixel_buffer as *mut _),
+            },
+        }
+    }
+
     pub fn sys_handle(&self) -> &vfb_sys::ffi::VideoFrameBuffer {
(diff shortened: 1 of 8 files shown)


theomonnom (approved):

theomonnom:
Seems good thanks!

PR #360, opened 2024-06-29
fix: simplified `enable_queue` field for AudioSource.

cloudwebrtc:


theomonnom (approved):

cloudwebrtc:
after several days of hard debugging, I had to remove the call of `self.sys_handle.on_captured_frame` from `captured_frame`.
I tried to break the audio frame in Unity AudioSource into 10ms frames, or directly pass all buffers into captured_frame. However, due to some call jitter, the encoded sound jittered and the delay accumulated. I guess some lock contention caused it, but I couldn't find the real reason.

So the current way that works perfectly is to use `enable_queue` to control whether to execute the `interval.tick()` timer wait inside `spawn`, and whether to insert a silent frame when `po_rx` is empty.

PR #365, opened 2024-07-10
livekit: send mute events to SFU

### LINKED ISSUE #358: Not receiving `TrackMuted` or `ActiveSpeakersChanged` events
In my simple test app, I do not seem to receive events about muting, unmuting, or active speakers changing. Is there anything I need to opt into in order to receive these events?

To reproduce in a fairly minimal test app, checkout [this branch](https://github.com/zed-industries/zed/pull/13343) of Zed, and run 

```
cargo run -p live_kit_client --example test_app
```

It connects to a live kit server on the default port, 7880.

nbsp:
fixes #358 (`ActiveSpeakersChanged` always got sent for me, this fixes the `Muted` issues)

tested with Zed's example, with rust-sdks on commit `ffi-v0.6.1~3` because some APIs changed since they made the example.

### DIFF
--- a/livekit/src/room/participant/mod.rs
+++ b/livekit/src/room/participant/mod.rs
@@ -248,8 +248,19 @@ pub(super) fn add_publication(
     publication.on_muted({
         let events = inner.events.clone();
         let participant = participant.clone();
+        let rtc_engine = inner.rtc_engine.clone();
         move |publication| {
             if let Some(cb) = events.track_muted.lock().as_ref() {
+                let rtc_engine = rtc_engine.clone();
+                let publication_cloned = publication.clone();
+                livekit_runtime::spawn(async move {
+                    let _ = rtc_engine
+                        .mute_track(proto::MuteTrackRequest {
+                            sid: publication_cloned.sid().to_string(),
+                            muted: true,
+                        })
+                        .await;
+                });
                 cb(participant.clone(), publication);
             }
         }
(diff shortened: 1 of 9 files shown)


livekit/src/room/participant/mod.rs
```diff
@@ -248,8 +248,19 @@ pub(super) fn add_publication(
     publication.on_muted({
         let events = inner.events.clone();
         let participant = participant.clone();
+        let rtc_engine = inner.rtc_engine.clone();
         move |publication| {
             if let Some(cb) = events.track_muted.lock().as_ref() {
+                let rtc_engine = rtc_engine.clone();
```
theomonnom: When unmuting or muting again, there is no guarantee that these tasks will run in order. Seems like a race could happen here?

livekit/src/room/participant/mod.rs
theomonnom, line 1: We should only send MuteTrackRequest for local tracks. This is sending mute requests even for remote tracks

livekit/src/room/participant/mod.rs
nbsp: as in the `mute_track().await` could take longer then the unmute after it?

livekit/src/room/participant/mod.rs
nbsp: added an if clause, this still runs on `Participant` just because of the way the `TrackPublication` gets passed inside a function so we can't access it until we run it

livekit/src/room/participant/mod.rs
theomonnom: not necessarily longer, but the unmute could be executed before the mute request

nbsp: new commit locks mute and unmute under the same mutex

livekit/src/room/participant/mod.rs
```diff
@@ -248,8 +248,21 @@ pub(super) fn add_publication(
...
         let events = inner.events.clone();
         let participant = participant.clone();
+        let rtc_engine = inner.rtc_engine.clone();
         move |publication| {
             if let Some(cb) = events.track_muted.lock().as_ref() {
+                if !publication.is_remote() {
+                    let rtc_engine = rtc_engine.clone();
+                    let publication_cloned = publication.clone();
+                    livekit_runtime::spawn(async move {
+                        let _ = rtc_engine
```
theomonnom: I don't think we should ignore the error here

livekit/src/room/participant/mod.rs
theomonnom: I saw you added the mutex, but this isn't fixing the issue. The race could still be there

livekit/src/room/participant/mod.rs
theomonnom: Same as the other PR. We should log instead of returning inside the tokio task

theomonnom (approved):

PR #367, opened 2024-07-12
Add participant attributes to the FFI

keepingitneil:
Also:
- rename update_name/attributes/metadata -> set_name/attributes/metadata for consistency with client-sdk-js
- add async getRoomSid to FFI

### DIFF
--- a/livekit-ffi/protocol/room.proto
+++ b/livekit-ffi/protocol/room.proto
@@ -285,6 +285,7 @@ message RoomEvent {
     E2eeStateChanged e2ee_state_changed = 24;
     RoomEOS eos = 25; // The stream of room events has ended
     DataPacketReceived data_packet_received = 26;
+    ParticipantAttributesChanged participant_attributes_changed = 27;
   }
 }
 
@@ -372,6 +373,12 @@ message RoomSidChanged {
 message ParticipantMetadataChanged { 
   string participant_sid = 1;
   string metadata = 2;
+}
+
+message ParticipantAttributesChanged {
+  string participant_sid = 1;
+  map<string, string> old_attributes = 2;
+  map<string, string> attributes = 3;
 }
 
 message ParticipantNameChanged { 
--- a/livekit-ffi/protocol/room.proto
+++ b/livekit-ffi/protocol/room.proto
@@ -117,39 +117,42 @@ message PublishTranscriptionCallback {
 }
 
 // Change the local participant's metadata
-message UpdateLocalMetadataRequest {
+message SetLocalMetadataRequest {
   uint64 local_participant_handle = 1;
   string metadata = 2;
 }
-message UpdateLocalMetadataResponse {
+message SetLocalMetadataResponse {
   uint64 async_id = 1;
 }
-message UpdateLocalMetadataCallback {
+message SetLocalMetadataCallback {
   uint64 async_id = 1;
+  optional string error = 2;
 }
 
 // Change the local participant's attributes
-message UpdateLocalAttributesRequest {
+message SetLocalAttributesRequest {
   uint64 local_participant_handle = 1;
   map<string, string> attributes = 2;
 }
-message UpdateLocalAttributesResponse {
+message SetLocalAttributesResponse {
   uint64 async_id = 1;
 }
-message UpdateLocalAttributesCallback {
+message SetLocalAttributesCallback {
   uint64 async_id = 1;
+  optional string error = 2;
 }
 
 // Change the local participant's name
-message UpdateLocalNameRequest {
+message SetLocalNameRequest {
   uint64 local_participant_handle = 1;
   string name = 2;
 }
-message UpdateLocalNameResponse {
+message SetLocalNameResponse {
   uint64 async_id = 1;
 }
-message UpdateLocalNameCallback {
+message SetLocalNameCallback {
   uint64 async_id = 1;
+  optional string error = 2;
 }
 
 // Change the "desire" to subs2ribe to a track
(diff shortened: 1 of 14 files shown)


livekit-ffi/protocol/room.proto
```diff
@@ -276,15 +288,16 @@ message RoomEvent {
     RoomSidChanged room_sid_changed = 15;
     ParticipantMetadataChanged participant_metadata_changed = 16;
     ParticipantNameChanged participant_name_changed = 17;
-    ConnectionQualityChanged connection_quality_changed = 18;
```
keepingitneil, line 279: For FFI we don't need to maintain message number consistency

livekit-ffi/src/server/room.rs
```diff
@@ -465,6 +465,28 @@ impl RoomInner {
...
+                .local_participant()
+                .update_attributes(update_local_attributes.attributes)
+                .await;
+
+            let _ = server.send_event(proto::ffi_event::Message::UpdateLocalAttributes(
+                proto::UpdateLocalAttributesCallback { async_id },
+            ));
+        });
+        server.watch_panic(handle);
+        proto::UpdateLocalAttributesResponse { async_id }
```
davidzhao: more of a question/note about request/response semantics in the FFI layer

we are planning to allow set_attributes/set_metadata to be able to return failure & error string. This could be due to the participant not having permission to update that field. How do we reflect these errors on the Rust/FFI side?

davidzhao (commented):
logic lgtm. will leave the approval to someone who's more fluent with the Rust stack

livekit-ffi/protocol/room.proto
```diff
@@ -128,6 +128,18 @@ message UpdateLocalMetadataCallback {
   uint64 async_id = 1;
 }
 
+// Change the local participant's attributes
+message UpdateLocalAttributesRequest {
```
theomonnom: nit: UpdateLocalAttributesRequest vs UpdateAttributesRequest?

livekit-ffi/protocol/room.proto
theomonnom: This is true, tho I'm thinking about changing the behaviour in the future. Safe to ignore for now

livekit-ffi/src/server/room.rs
```diff
@@ -465,6 +465,28 @@ impl RoomInner {
...
+
+    pub fn update_local_attributes(
+        self: &Arc<Self>,
+        server: &'static FfiServer,
+        update_local_attributes: proto::UpdateLocalAttributesRequest,
+    ) -> proto::UpdateLocalAttributesResponse {
+        let async_id = server.next_id();
+        let inner = self.clone();
+        let handle = server.async_runtime.spawn(async move {
+            let _ = inner
```
theomonnom: We should not ignore the error

livekit-ffi/src/server/room.rs
theomonnom: Rust FFI  is already exposing the error, but this PR ignores it above

livekit-ffi/protocol/room.proto
```diff
@@ -128,6 +128,18 @@ message UpdateLocalMetadataCallback {
...
+  uint64 local_participant_handle = 1;
+  map<string, string> attributes = 2;
+}
+message UpdateLocalAttributesResponse {
+  uint64 async_id = 1;
+}
+message UpdateLocalAttributesCallback {
+  uint64 async_id = 1;
```
theomonnom, line 141:
```suggestion
  uint64 async_id = 1;
  optional string error = 2;
```

livekit-ffi/src/server/room.rs
theomonnom: I see the errors are also not exposed for update_metadata, can you add it inside this PR?

livekit-ffi/protocol/room.proto
keepingitneil: kept consistency here with others (`UpdateLocalMetadataRequest`)

Since i'll changing Update -> Set in this PR, want me to get rid of Local?

livekit-ffi/src/server/requests.rs
```diff
@@ -54,6 +53,26 @@ fn on_connect(
...
+/// Retrieve the room sid
+fn on_room_sid(
+    server: &'static FfiServer,
+    req: proto::GetRoomSidRequest,
+) -> FfiResult<proto::GetRoomSidResponse> {
+    let async_id = server.next_id();
+    let handle = server.async_runtime.spawn(async move {
+        let ffi_room = server
+        .retrieve_handle::<FfiRoom>(req.room_handle)
+        .unwrap()
```
theomonnom: avoid unwrap here, use ?

livekit-ffi/src/server/room.rs
```diff
@@ -429,41 +429,63 @@ impl RoomInner {
...
-                proto::UpdateLocalNameCallback { async_id },
+                proto::SetLocalNameCallback { async_id, error: res.err().map(|e| e.to_string()) },
             ));
         });
         server.watch_panic(handle);
-        proto::UpdateLocalNameResponse { async_id }
+        proto::SetLocalNameResponse { async_id }
+    }
+
+    pub fn update_local_attributes(
```
theomonnom, line 469: wondering if we should remove local on the name

theomonnom (approved):
lgtm otherwise.
use cargo fmt

livekit-ffi/protocol/room.proto
```diff
@@ -49,6 +49,22 @@ message ConnectCallback {
   repeated ParticipantWithTracks participants = 5;
 }
 
+// Get Room Sid
+
+message GetRoomSidRequest {
```
theomonnom: Ah, proto is still there

PR #368, opened 2024-07-12
livekit: send UpdateTrackSettings on disable

### LINKED ISSUE #359: Can't stop a remote audio track from playing with `RtcAudioTrack::set_enabled(false)`
The `RtcAudioTrack::set_enabled` method (called by `RemoteAudioTrack::disable`) does not seem to do anything for me. In Zed, we are attempting to use those methods to implement "deafen" (temporarily turn off audio from all participants).

I'm testing on macOS.

nbsp:
fixes #359

### DIFF
--- a/livekit/src/room/participant/remote_participant.rs
+++ b/livekit/src/room/participant/remote_participant.rs
@@ -366,9 +366,7 @@ impl RemoteParticipant {
                         disabled: !enabled,
                         width,
                         height,
-                        quality: proto::VideoQuality::High.into(), // deprecated, but not marked in proto
-                        priority: 0,
-                        fps: 0,
+                        ..Default::default()
                     };
 
                     rtc_engine
--- a/livekit/src/room/participant/remote_participant.rs
+++ b/livekit/src/room/participant/remote_participant.rs
@@ -353,6 +353,32 @@ impl RemoteParticipant {
                 }
             }
         });
+
+        publication.on_enabled_status_changed({
+            let rtc_engine = self.inner.rtc_engine.clone();
+            move |publication, enabled| {
+                let rtc_engine = rtc_engine.clone();
+                livekit_runtime::spawn(async move {
+                    let tsid: String = publication.sid().into();
+                    let TrackDimension(width, height) = publication.dimension();
+                    let update_track_settings = proto::UpdateTrackSettings {
+                        track_sids: vec![tsid.clone()],
+                        disabled: !enabled,
+                        width,
+                        height,
+                        quality: proto::VideoQuality::High.into(), // deprecated, but not marked in proto
+                        priority: 0,
+                        fps: 0,
+                    };
+
+                    let _ = rtc_engine
+                        .send_request(proto::signal_request::Message::TrackSetting(
+                            update_track_settings,
+                        ))
+                        .await;
+                });
+            }
+        })
     }
 
     pub(crate) fn remove_publication(&self, sid: &TrackSid) -> Option<TrackPublication> {
--- a/livekit/src/room/participant/remote_participant.rs
+++ b/livekit/src/room/participant/remote_participant.rs
@@ -371,11 +371,11 @@ impl RemoteParticipant {
                         fps: 0,
                     };
 
-                    let _ = rtc_engine
+                    rtc_engine
                         .send_request(proto::signal_request::Message::TrackSetting(
                             update_track_settings,
                         ))
-                        .await;
+                        .await
                 });
             }
         })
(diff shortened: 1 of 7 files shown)


nbsp:
@maxbrunsfeld the test app needs some refactoring to work with this app, as it currently disables the RTC track directly, and wouldn't send any updates to the LiveKit SFU. something like this was enough for my testing:

<details>
<summary>a diff</summary>

```diff
diff --git a/crates/live_kit_client/examples/test_app.rs b/crates/live_kit_client/examples/test_app.rs
index 127f4be05..11aff6f4e 100644
--- a/crates/live_kit_client/examples/test_app.rs
+++ b/crates/live_kit_client/examples/test_app.rs
@@ -12,7 +12,7 @@ use live_kit_client::{
     options::{TrackPublishOptions, VideoCodec},
     participant::{Participant, RemoteParticipant},
     play_remote_audio_track,
-    publication::LocalTrackPublication,
+    publication::{LocalTrackPublication, RemoteTrackPublication},
     track::{LocalTrack, RemoteAudioTrack, RemoteTrack, RemoteVideoTrack, TrackSource},
     AudioStream, RemoteVideoTrackView, Room, RoomEvent, RoomOptions,
 };
@@ -91,7 +91,7 @@ struct LivekitWindow {
 
 #[derive(Default)]
 struct ParticipantState {
-    audio_output_stream: Option<(RemoteAudioTrack, AudioStream)>,
+    audio_output_stream: Option<(RemoteTrackPublication, AudioStream)>,
     muted: bool,
     screen_share_output_view: Option<(RemoteVideoTrack, View<RemoteVideoTrackView>)>,
     speaking: bool,
@@ -170,20 +170,23 @@ impl LivekitWindow {
             }
 
             RoomEvent::TrackSubscribed {
-                track, participant, ..
+                publication,
+                participant,
+                ..
             } => {
                 let output = self.remote_participant(participant);
-                match track {
-                    RemoteTrack::Audio(track) => {
+                match publication.track() {
+                    Some(RemoteTrack::Audio(track)) => {
                         output.audio_output_stream =
-                            Some((track.clone(), play_remote_audio_track(&track, cx)));
+                            Some((publication.clone(), play_remote_audio_track(&track, cx)));
                     }
-                    RemoteTrack::Video(track) => {
+                    Some(RemoteTrack::Video(track)) => {
                         output.screen_share_output_view = Some((
                             track.clone(),
                             cx.new_view(|cx| RemoteVideoTrackView::new(track, cx)),
                         ));
                     }
+                    None => todo!(),
                 }
                 cx.notify();
             }
@@ -317,8 +320,12 @@ impl LivekitWindow {
                 None
             }
         })?;
-        let track = participant.audio_output_stream.as_ref()?.0.rtc_track();
-        track.set_enabled(!track.enabled());
+        let publication = &participant.audio_output_stream.as_ref()?.0;
+        if publication.is_enabled() {
+            publication.set_enabled(false);
+        } else {
+            publication.set_enabled(true);
+        }
         cx.notify();
         Some(())
     }
@@ -390,7 +397,7 @@ impl Render for LivekitWindow {
                                 el.child(
                                     button()
                                         .id(SharedString::from(identity.0.clone()))
-                                        .child(if state.0.rtc_track().enabled() {
+                                        .child(if state.0.is_enabled() {
                                             "Deafen"
                                         } else {
                                             "Undeafen"
```
</details>

livekit/src/room/participant/remote_participant.rs
```diff
@@ -353,6 +353,32 @@ impl RemoteParticipant {
...
+                        track_sids: vec![tsid.clone()],
+                        disabled: !enabled,
+                        width,
+                        height,
+                        quality: proto::VideoQuality::High.into(), // deprecated, but not marked in proto
+                        priority: 0,
+                        fps: 0,
+                    };
+
+                    let _ = rtc_engine
```
theomonnom: I don't think we should ignore the error here?

theomonnom: This is still not ideal. Returning a value from a tokio task is still ignoring the error.
I think what you can do is to log an error when not OK

nbsp: `send_request` actually returns `()`, so there's nothing to do here, and `let _` isn't needed. there's no error value returned that we can check for. other PR does have this issue though

theomonnom (approved):

livekit/src/room/participant/remote_participant.rs
```diff
@@ -353,6 +353,32 @@ impl RemoteParticipant {
...
+                    let tsid: String = publication.sid().into();
+                    let TrackDimension(width, height) = publication.dimension();
+                    let update_track_settings = proto::UpdateTrackSettings {
+                        track_sids: vec![tsid.clone()],
+                        disabled: !enabled,
+                        width,
+                        height,
+                        quality: proto::VideoQuality::High.into(), // deprecated, but not marked in proto
+                        priority: 0,
+                        fps: 0,
```
theomonnom:
```suggestion
                        ..Default::default()
```

nbsp: i think the default for proto::VideoQuality would be `None`, which doesn't matter since it's ignored but it's still not in line with the JS SDK. should we default it as well or make it `High`?

theomonnom: I guess the SFU will ignore when this is None

PR #371, opened 2024-07-15
Add publish_dtmf

lukasIO:
implementation naming varies between `publish_dtmf` and `publish_sip_dtmf`. 
The intention was to use `publish_dtmf` on exposed APIs and `publish_sip_dtmf` for sdk-internal messaging.

Not sure anymore if that distinction makes any sense though.


dennwc (approved):
Cannot really comment on Rust code itself, but protocol parts look good to me :+1:

theomonnom (approved):
lgtm

PR #373, opened 2024-07-15
Rename "update" to "set" in overlooked places + fix broken new proto field

keepingitneil:


livekit/src/room/participant/local_participant.rs
```diff
@@ -231,6 +235,7 @@ impl LocalParticipant {
                     metadata,
                     name: self.name(),
                     attributes: Default::default(),
+                    request_id: 0,
```
theomonnom: use Default::default

theomonnom (approved):

PR #375, opened 2024-07-15
Use identity instead of sid where possible

keepingitneil:


davidzhao (commented):
I'm not sure we'd want to change it in all of the *internal* messages..

the goal is to allow end users to use identity, and not pID (because they often have no idea what the pID is). I think for certain things (i.e. disconnected messages), we would prefer to use the pID there, because it would alter the meaning. (for example, when a participant reconnects, we would want to make sure the previous one is removed, not the new one.)

keepingitneil:
Hrmm ok, @theomonnom and I spoke offline about keying all local participants off identity and removing the `participants_by_identity` mapping on the python side which would require using identity for the ffi.

Are you saying that there is logic that requires participant_sid (due to out of order events maybe?) in cases where the identity is the same?

theomonnom (approved):

theomonnom:
is the livekit crate using sid?

PR #376, opened 2024-07-17
fix {Open,Boring}SSL conflicts

nbsp:
uses `objcopy` to rename all exported BoringSSL symbols, prepending `LK_` to them, so that they don't conflict with native OpenSSL on the Rust side.

previous PR description before figuring this out below.

---

PR for review.
trying a different approach than manually patching in the list (though that's still available & commented out), since even when it finally built (90 minute patch time!) it fixed nothing.

the [BoringSSL build doc](https://boringssl.googlesource.com/boringssl/+/HEAD/BUILDING.md#building-with-prefixed-symbols) gives us an option to prefix *all* BSSL symbols with whatever we want, in our case `LK`. i've punched that into the `BUILD.gn` file, and it builds just fine in this diff, but the `basic_room` linking fails with a lot of the following error

```python
/usr/bin/ld: openssl_adapter.cc:(.text+0x7f): undefined reference to `LK_i2d_X509' # and similar symbols
```

even after i've manually gone in and updated the files to use `LK_`-prefixed symbols, they just don't link. fresh building environment. i don't know exactly how it would compile by itself (clearly no undefined symbols) and then fail at the `basic_room` link phase.


webrtc-sys/libwebrtc/build_linux.sh
```diff
@@ -67,9 +67,13 @@ then
 fi
 
 cd src
+git fetch origin m114_release
```
davidzhao, line 70: I guess this branch has not updated to m125 yet? it would be upgrade too

davidzhao (approved):
AWESOME WORK! 🚀 🚀 🚀

webrtc-sys/libwebrtc/build_linux.sh
nbsp: not yet. currently we use m114, but because m125 was released we now need to manually checkout. the `fix_m114` patch deals with a few bugs that were fixed in the new version. i think upgrading is better suited for a future PR, let's get this tagged first

davidzhao: sounds good. agreed it should be in a separate PR

kuba--:
Definitivelly awesome work!!!
And looks like it all works (also if I add extra `isahc` which uses curl+openssl).

kuba-- (approved):

PR #377, opened 2024-07-18
Calculate changed attributes + fix propagation

keepingitneil:


livekit-ffi/protocol/room.proto
```diff
@@ -407,8 +407,7 @@ message ParticipantMetadataChanged {
 
 message ParticipantAttributesChanged {
   string participant_identity = 1;
-  map<string, string> old_attributes = 2;
-  map<string, string> attributes = 3;
+  map<string, string> changed_attributes = 2;
```
lukasIO: if we only send the changed attributes, then each language implementation has to re-construct the currently set attributes from the changed_attributes, right?
Should we maybe just send both, changed + current ones?

keepingitneil: ahh yes you're right, I keep forgetting that the downstream clients keep their own state (instead of passthrough to rust).

Will update now

livekit/src/room/utils/mod.rs
```diff
@@ -0,0 +1,20 @@
+use std::collections::HashMap;
+
+pub fn calculate_changed_attributes(
+    old_attributes: HashMap<String, String>,
+    new_attributes: HashMap<String, String>,
+) -> HashMap<String, String> {
+    let old_keys: Vec<_> = old_attributes.keys().collect();
+    let new_keys: Vec<_> = new_attributes.keys().collect();
+    let all_keys: Vec<_> = old_keys.into_iter().chain(new_keys.into_iter()).collect();
```
nbsp: i think `keys()` already returns an iterator, you might not need to run `collect().into_iter()` on it

theomonnom (approved):

Issue #380, opened 2024-07-23
Failing Rust SDK Examples


edavalosanaya:
First of all, great and amazing work from the LiveKit team! Thank y'all for all the time and effort you have put into this project.

As I have started working on integrating embedded device to join a LiveKit room session, I encountered some dependency issues and failing examples.

All of this testing was done on an Ubuntu 22.04 and using a local Livekit server, following the instructions found in [LiveKit's Official Documentation](https://docs.livekit.io/home/self-hosting/local/).

All of my changes to make the examples to (somewhat) run can be found [my personal fork](https://github.com/edavalosanaya/rust-sdks).

Here are the issues in the tutorials:

1. All examples require the ``livekit`` dependencies needed to be bumped to ``0.4.1``
2. ``api`` example results in the following:
```bash
$ LIVEKIT_URL=http://127.0.0.1:7880 LIVEKIT_API_KEY=devkey LIVEKIT_API_SECRET=secret cargo run -p api
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
   Compiling livekit-protocol v0.3.4 (/home/nicole/GitHub/external/rust-sdks/livekit-protocol)
error[E0425]: cannot find function `spawn` in crate `livekit_runtime`
  --> /home/nicole/GitHub/external/rust-sdks/livekit-protocol/src/debouncer.rs:38:22
   |
38 |     livekit_runtime::spawn(debounce_task(duration, future, rx, cancel_rx));
   |                      ^^^^^ not found in `livekit_runtime`
   |
help: consider importing one of these items
   |
15 + use std::thread::spawn;
   |
15 + use tokio::spawn;
   |
help: if you import `spawn`, refer to it directly
   |
38 -     livekit_runtime::spawn(debounce_task(duration, future, rx, cancel_rx));
38 +     spawn(debounce_task(duration, future, rx, cancel_rx));
   |

error[E0425]: cannot find function `sleep` in crate `livekit_runtime`
  --> /home/nicole/GitHub/external/rust-sdks/livekit-protocol/src/debouncer.rs:54:34
   |
[...]

PR #382, opened 2024-07-24
Fix broken examples.

cloudwebrtc:
changes:
Since the examples use directories as dependencies, they should always correspond to the livekit api version, so the version field is removed from all examples' Cargo.toml

close https://github.com/livekit/rust-sdks/issues/380

Just tested, basic_room and wgpu_room are working again.


theomonnom (approved):
lgtm!

PR #385, opened 2024-08-02
livekit-ffi: add DisconnectReason

nbsp:
updates livekit-protocol to latest main


livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -518,6 +518,7 @@ impl SessionInner {
                         })
                         .unwrap(),
                         target: target as i32,
+                        r#final: false,
```
nbsp: unsure about this one, lmk if it needs to be taken from somewhere

theomonnom (approved):
awesome thanks!

livekit/src/rtc_engine/rtc_session.rs
theomonnom: seems like it is unused on client-sdk-js
what you can do is to put ..Default::default() instead of false.

PR #394, opened 2024-08-16
Update protocol to v15

nbsp:
exported as LocalTrackSubscribed in the client and FFI.

at the moment the SFU assumes Rust doesn't accept TrackSubscribed so it doesn't send it: see https://github.com/livekit/livekit/pull/2939


davidzhao:
Currently Rust stack is on protocol version 9. part of this update will require bringing it to 10+. and I think we should just bring it up to date to support protocol 15.

Here are the changes required (JS PRs)
* v10: https://github.com/livekit/client-sdk-js/pull/881
* v11: https://github.com/livekit/client-sdk-js/pull/961
* v12: https://github.com/livekit/client-sdk-js/pull/983
* v13: https://github.com/livekit/client-sdk-js/pull/1127
* v14: https://github.com/livekit/client-sdk-js/pull/1221

nbsp:
v10 through v13 added or already there. re: `RequestResponse`: we don't keep a list of pending requests and we don't care about arbitrary responses at all -- most of our signal_client requests go to `_`. should i add it regardless and make it no-op?

davidzhao:
> v10 through v13 added or already there. re: `RequestResponse`: we don't keep a list of pending requests and we don't care about arbitrary responses at all -- most of our signal_client requests go to `_`. should i add it regardless and make it no-op?

yeah no-op is fine. we can add the capability later of firing error responses for things, but not required to be protocol-compatible.

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -778,8 +798,8 @@ impl SessionInner {
     async fn simulate_scenario(&self, scenario: SimulateScenario) -> EngineResult<()> {
         let simulate_leave = || {
             self.on_signal_event(proto::signal_response::Message::Leave(proto::LeaveRequest {
+                action: proto::leave_request::Action::Reconnect.into(),
```
davidzhao, line 796: it doesn't look like this is handled? this would determine if a client should reconnect vs disconnect for good.

davidzhao (commented):
nice work! Just a couple of things:

1. see inline about handling action
2. the streamID change is a bit tricky and we would need to test e2e to be sure. the way to test is creating a room with sync-stream enabled:
  - `lk room create --name <myroom> --sync-streams`
  - then starting a publisher immediately: `lk load-test --room <myroom> --publishers 1 --audio-publishers 1`
  - then join the room with Rust and ensure both streams are retrieved correctly

livekit/src/rtc_engine/rtc_session.rs
nbsp: from what i've seen rust-sdks reconnection logic is fundamentally different, and doesn't use `LeaveRequest` for retrying at all? it attempts to restart the publisher, or completely closing it (with `Action::Disconnect`) then reconnecting and recovering? i'll see if i can make it consistent with JS

davidzhao: I think the key here is if server is not sending a reconnect action, Rust should not attempt to reconnect. CC: @theomonnom

nbsp: also can confirm streams are received and handled correctly

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -106,13 +106,20 @@ pub enum SessionEvent {
     RoomUpdate {
         room: proto::Room,
     },
+    TrackSubscribed {
```
theomonnom: Let's also keep the same naming here (LocalTrackSubscribed)

livekit/src/room/mod.rs
```diff
@@ -716,8 +722,13 @@ impl RoomSession {
         }
 
         let (participant_sid, stream_id) = lk_stream_id.unwrap();
+        let mut track_id = track.id();
+        if stream_id.starts_with("TR") {
+            track_id = stream_id.into();
+        }
```
theomonnom, line 730: I guess this logic on the JS side was just for Firefox? Which seems unneeded here?

theomonnom (approved):
otherwise lgtm

livekit/src/room/mod.rs
nbsp: should it always be stream_id or track_id, then?

theomonnom (changes requested):
nvm, seems like it is still not returning an Error when receiving a bad RequestResponse

livekit/src/room/mod.rs
davidzhao: not just for firefox.. this is required as by protocol v10: https://github.com/livekit/client-sdk-js/pull/881

davidzhao: protocol v10+, we have two ways of packing the trackID, and clients need to support both

livekit-api/src/signal_client/mod.rs
```diff
@@ -92,6 +92,7 @@ struct SignalInner {
     url: String,
     options: SignalOptions,
     join_response: proto::JoinResponse,
+    request_id: AsyncMutex<u32>,
```
theomonnom: Let's use an Atomic integer instead

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -336,6 +341,10 @@ impl RtcSession {
     pub fn data_channel(&self, target: SignalTarget, kind: DataPacketKind) -> Option<DataChannel> {
         self.inner.data_channel(target, kind)
     }
+
+    pub async fn get_response(&self, request_id: u32) -> proto::RequestResponse {
+        self.inner.get_response(request_id).await
```
lukasIO, line 346: I didn't see a timeout anywhere (maybe I missed it), would be good to have a timeout that rejects a request after a certain time in order for the request not to block indefinitely.

livekit/src/room/mod.rs
```diff
@@ -69,6 +69,8 @@ pub enum RoomError {
     TrackAlreadyPublished,
     #[error("already closed")]
     AlreadyClosed,
+    #[error("request error: {}", .0.message)]
+    Request(proto::RequestResponse),
```
theomonnom: Can you recreate another struct, basically having the same fields of the RequestResponse but removing the internal fields like the request_id

livekit/src/room/participant/local_participant.rs
```diff
@@ -236,48 +242,96 @@ impl LocalParticipant {
...
+            self.inner.rtc_engine.get_response(request_id)
+        })
+        .await
+        {
+            match response.reason() {
+                Reason::Ok => Ok(()),
+                _ => Err(RoomError::Request(response)),
+            }
+        } else {
+            log::error!("could not receive response for set_metadata request");
```
theomonnom: Logs aren't useful if we're already returning an error

theomonnom (approved):

livekit/src/room/participant/local_participant.rs
```diff
@@ -236,48 +242,96 @@ impl LocalParticipant {
...
-                    attributes,
-                    metadata: self.metadata(),
-                    name: self.name(),
-                    ..Default::default()
-                },
-            ))
-            .await;
-        Ok(())
+        if let Ok(response) = timeout(REQUEST_TIMEOUT, {
+            let request_id = self.inner.rtc_engine.session().signal_client().request_id();
```
theomonnom: ~~I think this should be moved to the Room impl? request_id doesn't necessarily needs to be on the signal client?~~
Also making it more explicit like "new_request_id"?

theomonnom (approved):
lgtm otherwise

livekit/src/room/participant/local_participant.rs
nbsp: the JS SDK puts it in the signal client, but it's a bit more cumbersome here, yeah
we can maybe move it to RtcEngine?

PR #396, opened 2024-08-20
Convenience API for iterating frames from a participant

keepingitneil:


livekit-ffi/src/server/audio_stream.rs
```diff
@@ -80,6 +82,97 @@ impl FfiAudioStream {
...
+                        c_rx,
+                    )
+                    .await;
+                    let _ = done_tx.send(());
+                });
+                tokio::select! {
+                    _ = &mut close_rx => {
+                        let _ = c_tx.send(());
+                        return
+                    }
```
nbsp: so the purpose of `c_tx` is just to close the running task if close is called?

livekit-ffi/src/server/video_stream.rs
```diff
@@ -141,4 +174,67 @@ impl FfiVideoStream {
...
+        };
+
+        let track_source = request.track_source();
+        let (track_tx, mut track_rx) = mpsc::channel::<Track>(1);
+        server.async_runtime.spawn(utils::track_changed_trigger(
+            ffi_participant,
+            track_source.into(),
+            track_tx,
+        ));
+        //  track_tx is no longer held, so the track_rx will be closed when track_changed_trigger is done
```
nbsp: nit: 
```suggestion
        // track_tx is no longer held, so the track_rx will be closed when track_changed_trigger is done
```

livekit-ffi/src/server/audio_stream.rs
```diff
@@ -80,6 +82,97 @@ impl FfiAudioStream {
...
+        };
+
+        let track_source = request.track_source();
+        let (track_tx, mut track_rx) = mpsc::channel::<Track>(1);
+        server.async_runtime.spawn(utils::track_changed_trigger(
+            ffi_participant,
+            track_source.into(),
+            track_tx,
+        ));
+        //  track_tx is no longer held, so the track_rx will be closed when track_changed_trigger is done
```
nbsp: nit:
```suggestion
        // track_tx is no longer held, so the track_rx will be closed when track_changed_trigger is done
```

livekit-ffi/src/server/utils.rs
```diff
@@ -0,0 +1,63 @@
...
+                    continue;
+                }
+                if publication.source() == track_source.into() {
+                    let track = publication.track();
+                    match track {
+                        Some(track) => {
+                            let _ = track_tx.send(track.into()).await;
+                        }
+                        _ => {}
+                    }
```
nbsp: `if let` would be better here as well as in line 15

livekit-ffi/src/server/audio_stream.rs
keepingitneil: yup that's the idea

PR #400, opened 2024-08-22
Run publish CI on ffi-v* and allow manual dispatch

nbsp:
i'll run the publish workflow once this is merged.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,9 +15,11 @@
 name: Publish crates
 
 on:
+  workflow_dispatch:
   push:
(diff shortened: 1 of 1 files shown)


theomonnom:
livekit-ffi doesn't need to be published to crates.io

nbsp:
> livekit-ffi doesn't need to be published to crates.io, this is internal? we can still do it but there is no value of doing it

`ffi-v*` versions are the ones getting tagged and released, so instead of tagging it with `vX.Y.Z ffi-vX.Y.Z` it's better to use the one tag.

PR #409, opened 2024-08-26
fixed ffi build for android.

cloudwebrtc:
Fixed compilation for android ffi
https://github.com/livekit/rust-sdks/actions/runs/12062288204


.github/workflows/builds.yml
```diff
@@ -85,5 +88,6 @@ jobs:
       - name: Build (Android)
         if: ${{ contains(matrix.target, 'android') }}
         run: |
+          ln -sf $ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/lib/${{ matrix.ndk_arch }}/{libunwind.so,libc++abi.a} $ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/lib/
```
nbsp, line 91: is there an envvar we can set instead of symlinking these?

nbsp (approved):
lg

.github/workflows/builds.yml
cloudwebrtc: When using ndk r25 above, this error will appear, but I can't find the real reason for it. It may be a bug in cargo ndk when linking `libunwind`. I observed the directory structure in ndk r27 and found that the `libunwind.so` symbolic link corresponds to `x86_64-unknown-linux-musl/libunwind.so`  by default, but this is obviously a mistake for aarch64 and armv7 compilation. I guess I should see if there is any issue/patch in the cargo-ndk upstream (our repo does not directly depend on unwind related packages).

theomonnom (approved):

PR #410, opened 2024-08-27
dynamically link webrtc

theomonnom:


libwebrtc/src/video_track.rs
```diff
@@ -12,20 +12,37 @@
...
-    pub(crate) handle: imp_vt::RtcVideoTrack,
+    pub(crate) ffi: sys::RefCounted<sys::lkRtcVideoTrack>,
 }
 
 impl RtcVideoTrack {
     media_stream_track!();
+
+    pub fn add_sink(&self, sink: Arc<NativeVideoSink>) {
+        unsafe {
+            sys::lkVideoTrackAddSink(self.ffi.as_ptr(), sink.ffi.as_ptr() as *mut std::ffi::c_void);
```
ladvoc: question: Same question as below.

libwebrtc/src/sys/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+pub use ffi::*;
+pub use refcounted::*;
+pub use utils::*;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // PeerObserver
+    #[allow(non_snake_case)]
```
ladvoc: nitpick: I think you can put `#![allow(non_snake_case)]` at the top of the tests module instead of annotating every function.

libwebrtc/src/lib.rs
```diff
@@ -59,19 +60,26 @@ pub mod rtp_sender;
 pub mod rtp_transceiver;
 pub mod session_description;
 pub mod stats;
+pub mod sys;
```
ladvoc, line 62: question: Does the sys module need to be public? Generally, FFI interface concerns should be taken care of within the crate (or corresponding _-sys_ crate) and FFI-specific types should not be in the public API.

cloudwebrtc: yeah,  `sys` should not be exposed as a public API.

cloudwebrtc: However, if we hide `sys` inside the internal, it seems we need to design an interface and implementation for each API, which seems to bring us back to square one, increasing the code size.

ladvoc: Looking at the existing implementation, we already expose the equivalent _cxx_ types in quite a few places. Ideally we would hide all FFI concerns, but that would be outside the scope of this PR.

libwebrtc/src/media_stream_track.rs
```diff
@@ -12,16 +12,30 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[cfg(not(target_arch = "wasm32"))]
```
ladvoc, line 15: question: I'm wondering if we should remove all conditional compilation for WASM.

libwebrtc/src/peer_connection.rs
```diff
@@ -68,15 +69,54 @@ pub enum SignalingState {
...
+    pub use_rtp_mux: bool,
+    pub offer_to_receive_audio: bool,
+    pub offer_to_receive_video: bool,
+}
+
+impl Default for OfferOptions {
+    fn default() -> Self {
+        Self {
+            ice_restart: false,
+            use_rtp_mux: true,
```
ladvoc, line 84: question: I noticed `use_rtp_mux` is newly added and is the only property defaulting to true. Is this intentional?

libwebrtc/src/peer_connection.rs
```diff
@@ -68,15 +69,54 @@ pub enum SignalingState {
...
     pub offer_to_receive_audio: bool,
     pub offer_to_receive_video: bool,
 }
 
-#[derive(Debug, Clone, Default)]
-pub struct AnswerOptions {}
+impl From<AnswerOptions> for lkOfferAnswerOptions {
+    fn from(_options: AnswerOptions) -> Self {
+        lkOfferAnswerOptions {
+            iceRestart: false,
```
ladvoc: question: Should the equivalent Rust struct (_AnswerOptions_) expose _iceRestart_ and _useRtpMux_? If so, the defaults should probably be set in the _Default_ implementation for _AnswerOptions_ instead of this _From_ implementation.

libwebrtc/src/peer_connection_factory.rs
```diff
@@ -67,28 +93,96 @@ impl Debug for PeerConnectionFactory {
...
+            sys::lkCreatePeer(
+                self.ffi.as_ptr(),
+                &lk_config,
+                &PEER_OBSERVER,
+                observer_ptr as *mut ::std::os::raw::c_void,
+            )
+        };
+        if sys_peer.is_null() {
+            unsafe {
+                let _ = Rc::from_raw(observer_ptr);
```
ladvoc: issue: this is UB. Since we obtain _observer_ptr_ from an _Arc_, we need to convert back to an _Arc_ in order for _Arc_'s drop implementation to perform cleanup properly.

libwebrtc/src/rtp_parameters.rs
```diff
@@ -54,18 +63,90 @@ pub struct RtcpParameters {
 pub struct RtpEncodingParameters {
     pub active: bool,
     pub max_bitrate: Option<u64>,
+    pub min_bitrate: Option<u64>,
     pub max_framerate: Option<f64>,
-    pub priority: Priority,
```
ladvoc, line 58: question: Was the removal of this field intentional? If so, the _Priority_ enum defined above should also be removed.

libwebrtc/src/rtp_parameters.rs
```diff
@@ -54,18 +63,90 @@ pub struct RtcpParameters {
...
+    pub has_message_type: bool,
+    pub message_type: RtcpFeedbackMessageType,
 }
 
 #[derive(Debug, Clone)]
 pub struct RtpCodecCapability {
     pub channels: Option<u16>,
     pub clock_rate: Option<u64>,
     pub mime_type: String,
+    pub preferred_payload_type: Option<u8>,
```
ladvoc, line 163: question: Based on the name, it seems like this should be an enum?

libwebrtc/src/rtp_transceiver.rs
```diff
@@ -15,61 +15,129 @@
...
-    SendOnly,
-    RecvOnly,
-    Inactive,
-    Stopped,
-}
-
 #[derive(Clone)]
 pub struct RtpTransceiver {
-    pub(crate) handle: imp_rt::RtpTransceiver,
+    pub ffi: sys::RefCounted<crate::sys::lkRtpTransceiver>,
```
ladvoc: suggestion: I think this can be made private. If not, it should be _pub(crate)_.

libwebrtc/src/rtp_transceiver.rs
```diff
@@ -15,61 +15,129 @@
...
+            }
+            Ok(())
+        }
     }
 
     pub fn stop(&self) -> Result<(), RtcError> {
-        self.handle.stop()
+        unsafe {
+            sys::lkRtpTransceiverStop(self.ffi.as_ptr());
+            //TODO: check for errors
```
ladvoc, line 138: question: What type of errors can occur here? Maybe it isn't worth exposing them.

libwebrtc/src/rtp_transceiver.rs
```diff
@@ -15,61 +15,129 @@
...
-        self.handle.set_codec_preferences(codecs)
+        unsafe {
+            let mut native_codecs = sys::RefCountedVector::new();
+
+            for c in codecs {
+                let mime_type_cstr = std::ffi::CString::new(c.mime_type.clone()).unwrap();
+                let cap = sys::lkRtpCodecCapabilityCreate();
+
+                if let Some(ch) = c.channels {
+                    sys::lkRtpCodecCapabilitySetChannels(cap, ch.try_into().unwrap());
```
ladvoc: nitpick: _ch_ is already _u16_.

libwebrtc/src/peer_connection.rs
cloudwebrtc: Yeah, these align with the default settings in libwebrtc.
https://github.com/webrtc-sdk/webrtc/blob/m137_release/api/peer_connection_interface.h#L730-L734

libwebrtc/src/media_stream_track.rs
cloudwebrtc: @theomonnom,  should we remove WASM support?

libwebrtc/src/rtp_parameters.rs
cloudwebrtc: yeah, The name is a bit strange, but it's actually the RTP payload ID.

PR #411, opened 2024-08-28
Cleanup media streams when finished

keepingitneil:


livekit-ffi/src/server/mod.rs
```diff
@@ -192,7 +196,13 @@ impl FfiServer {
     }
 
     pub fn drop_handle(&self, id: FfiHandleId) -> bool {
-        self.ffi_handles.remove(&id).is_some()
+        let existed = self.ffi_handles.remove(&id).is_some();
+        let _ = self.handle_dropped_tx.send(id).is_ok();
```
theomonnom: I think the overhead is going to be too big. This is executed for each event (each video/audio frame).
Is there a way to only do this when dropping a stream instead?

theomonnom: I think you can implement Drop for the specific removed handle ?

keepingitneil: Hrmm my rust isn't amazing, but Drop could be called multiple times in this case right? So there's not a good way to reliably know if it's the last drop? 

Maybe I can rework this to add the handle_id to the `watch_handle_dropped` fn

keepingitneil: Changed it to use handle id based subscription and DashMap for storing them

theomonnom: if the handle is stored inside `Arc<T>`, the Drop impl of `T` will only be called once

keepingitneil: Hrmm makes sense. Do you think that's better than this new oneshot + DashMap strategy which I would think is negligible performance-wise relative to the rest of webrtc and relative the ffi client/server messaging. 

FfiTrack.track is currently not an Arc so I'd need to change that and it would need use broadcast channel. Additionally I'd want to make sure track is not cloneable right? And Track is an enum for {local,remote}x{audio,video} so idk feels more complicated.

FWIW I will also want this on participant for my next PR.

theomonnom (approved):

PR #416, opened 2024-09-03
webrtc-sys: re-extract boringssl symbols

nbsp:
the previous list was from the bssl version m125 uses, which caused conflicts when we moved to using gclient to load m114.

this one greps for `OPENSSL_EXPORT` in the boringssl code.


webrtc-sys/libwebrtc/boringssl_prefix_symbols.txt
```diff
@@ -3441,196 +2693,45 @@ i2s_ASN1_OCTET_STRING LK_i2s_ASN1_OCTET_STRING
...
-spx_wots_sign LK_spx_wots_sign
-spx_xmss_pk_from_sig LK_spx_xmss_pk_from_sig
-spx_xmss_sign LK_spx_xmss_sign
+sk_zero LK_sk_zero
+ssl_cert_check_key_usage LK_ssl_cert_check_key_usage
+ssl_client_hello_init LK_ssl_client_hello_init
+ssl_decode_client_hello_inner LK_ssl_decode_client_hello_inner
+ssl_is_valid_ech_public_name LK_ssl_is_valid_ech_public_name
+ssl_session_serialize LK_ssl_session_serialize
+uint16_t LK_uint16_t
```
theomonnom: this doesn't seems right? how are you extracting the types?

theomonnom (approved):
otherwise lg

webrtc-sys/libwebrtc/boringssl_prefix_symbols.txt
nbsp: oops yeah, this one has a newline in the definition so it got the type. i'll have a manual comb through for those

theomonnom: can you also push the script that get the symbols, so it is reproducible by everyone

nbsp: i'll add it commented out to the top of the symbol rename list

PR #433, opened 2024-09-13
send silence frames on the AudioSource when the queue is empty

theomonnom:
old
Screenshot 2024-09-12 at 4 58 11 PM

new
Screenshot 2024-09-12 at 6 55 32 PM

also fix buffer reserving space using resize `instead` of `reserve` method


davidzhao (approved):
🎉

PR #436, opened 2024-09-20
Add chat APIs - wip

lukasIO:


livekit-ffi/src/proto.rs
```diff
@@ -1,4 +1,32 @@
 #![allow(non_snake_case)]
 #![allow(clippy::enum_variant_names)]
 
+use livekit::ChatMessage as RoomChatMessage;
+
 include!("livekit.proto.rs");
+
+impl From<ChatMessage> for RoomChatMessage {
```
lukasIO: currently defining these conversions both in livekit-ffi and in livekit, which feels wrong, but couldn't make it work otherwise.

livekit-api/src/services/sip.rs
```diff
@@ -233,6 +235,9 @@ impl SIPClient {
 
                         auth_username: options.auth_username.to_owned(),
                         auth_password: options.auth_password.to_owned(),
+
+                        headers: HashMap::new(),               // FIXME Lukas
+                        headers_to_attributes: HashMap::new(), // FIXME Lukas
```
lukasIO: @dennwc what should those header values default to?

livekit/Cargo.toml
```diff
@@ -40,3 +40,4 @@ futures-util = { version = "0.3", default-features = false, features = ["sink"]
 thiserror = "1.0"
 lazy_static = "1.4"
 log = "0.4"
+chrono = "0.4.38"
```
lukasIO, line 43: needed for timestamping chat messages

bcherry (approved):
this makes sense to me

livekit-ffi/protocol/room.proto
```diff
@@ -144,6 +144,28 @@ message SetLocalMetadataCallback {
...
+message SendChatMessageRequest {
+  uint64 local_participant_handle = 1;
+  string message = 2;
+  repeated string destination_identities = 3;
+  optional string sender_identity = 4;
+}
+message EditChatMessageRequest {
+  uint64 local_participant_handle = 1;
+  string edit_text = 2;
+  ChatMessage original_message = 3;
```
theomonnom, line 156: Do we need the whole ChatMessage? Is it possible to just use the message_id?

livekit-ffi/src/proto.rs
```diff
@@ -1,4 +1,32 @@
...
+    fn from(msg: RoomChatMessage) -> Self {
+        ChatMessage {
+            id: msg.id,
+            message: msg.message,
+            timestamp: msg.timestamp,
+            edit_timestamp: msg.edit_timestamp,
+            deleted: msg.deleted.into(),
+            generated: msg.generated.into(),
+        }
+    }
```
theomonnom: We have a module for all the conversions: https://github.com/livekit/rust-sdks/tree/main/livekit-ffi/src/conversion

theomonnom (approved):
lgtm!

livekit/src/room/participant/local_participant.rs
```diff
@@ -331,6 +333,56 @@ impl LocalParticipant {
...
+        };
+
+        let data = proto::DataPacket {
+            value: Some(proto::data_packet::Value::ChatMessage(chat_message.clone())),
+            participant_identity: sender_identity.unwrap_or_default(),
+            destination_identities: destination_identities.unwrap_or_default(),
+            ..Default::default()
+        };
+
+        let _ = self.inner.rtc_engine.publish_data(&data, DataPacketKind::Reliable).await;
```
theomonnom: We shouldn't ignore the error here

livekit/src/room/participant/local_participant.rs
```diff
@@ -331,6 +333,56 @@ impl LocalParticipant {
...
+        };
+        let proto_msg = proto::ChatMessage::from(edited_message);
+        let data = proto::DataPacket {
+            value: Some(proto::data_packet::Value::ChatMessage(proto_msg.clone())),
+            participant_identity: sender_identity.unwrap_or_default(),
+            destination_identities: destination_identities.unwrap_or_default(),
+            ..Default::default()
+        };
+
+        let _ = self.inner.rtc_engine.publish_data(&data, DataPacketKind::Reliable).await;
```
theomonnom: Same here

livekit-ffi/protocol/room.proto
lukasIO: in an ideal world, yeah, the ID would be enough. 
We don't have a central authority for chat messages though. So for an edited message we'd still want to keep the e.g. original timestamp of the message, so passing the whole message seemed like the best idea. 
But would be happy about alternative suggestions

livekit-ffi/src/conversion/room.rs
```diff
@@ -24,8 +26,7 @@ use livekit::{
         prelude::{ContinualGatheringPolicy, IceServer, IceTransportsType, RtcConfiguration},
     },
 };
-
-use crate::{proto, server::room::FfiRoom};
+use proto::ChatMessage;
```
theomonnom: can you use proto:: instead (not importing ChatMessage). 
So the code can be clearer (nit tho)

PR #439, opened 2024-09-22
feat: Try to add AEC3 for rust-core.

cloudwebrtc:


webrtc-sys/src/audio_track.cpp
```diff
@@ -163,15 +163,15 @@ AudioTrackSource::InternalSource::InternalSource(
 
         if (buffer_.size() >= samples10ms) {
           for (auto sink : sinks_)
-            sink->OnData(buffer_.data(), sizeof(int16_t), sample_rate_,
+            sink->OnData(buffer_.data(), sizeof(int16_t) * 8, sample_rate_,
```
theomonnom: ah blah, thx

theomonnom: funny that it works without it

cloudwebrtc:
yeah, assert is only triggered in debug mode.

Théo Monnom ***@***.***>于2024年9月23日 周一07:08写道：

webrtc-sys/src/audio_track.cpp
cloudwebrtc: yeah, assert is only triggered in debug mode.

PR #440, opened 2024-09-23
fix: resume/reconnection not working!

theomonnom:
- fix regression due to https://github.com/livekit/rust-sdks/pull/394 always disconnecting on signal close
- now directly trying full_reconnect if the reconnection error isn't a SignalError


keepingitneil (approved):

davidzhao (approved):
🙏

PR #449, opened 2024-09-30
avoid reconnection logs on normal close

theomonnom:


livekit/src/rtc_engine/mod.rs
```diff
@@ -402,15 +404,25 @@ impl EngineInner {
...
                 match action {
-                    proto::leave_request::Action::Resume => {
-                        self.reconnection_needed(retry_now, false)
-                    }
-                    proto::leave_request::Action::Reconnect => {
-                        self.reconnection_needed(retry_now, true)
+                    proto::leave_request::Action::Resume
+                    | proto::leave_request::Action::Reconnect => {
+                        log::warn!(
+                            "received session close: {:?} {:?} {:?}",
```
davidzhao, line 411:
```suggestion
                            "received server-initiated reconnect: {:?} {:?} {:?}",
```

davidzhao (approved):
nit on the log, otherwise lgtm

livekit/src/rtc_engine/mod.rs
theomonnom: it isn't necessarily server-initiated, this part needs cleanup but it is also called when the SignalClient is being closed unexpectedly. In this case we create our own leave_request locally with the reason Unknown.

PR #457, opened 2024-10-02
[WIP] Native RPC packet management

bcherry:
This works with https://github.com/livekit/node-sdks/pull/276 and https://github.com/livekit/livekit/pull/3067 but is likely to be superseded by https://github.com/livekit/rust-sdks/pull/461

PR #461, opened 2024-10-05
[WIP] Full RPC implementation

bcherry:
This is an alternative to https://github.com/livekit/rust-sdks/pull/457 which will implement the entire RPC feature in the Rust SDK and expose it via FFI to simplify the implementation of dependent SDKs, rather than just implementing the data packet sends and receives.  Basically its a complete reimplementation of https://github.com/livekit/node-sdks/pull/276 in Rust.


livekit/src/room/participant/local_participant.rs
```diff
@@ -490,4 +604,196 @@ impl LocalParticipant {
...
+                        Err(_) => {
+                            Err(RpcError::built_in(RpcErrorCode::RecipientDisconnected, None))
+                        }
+                    },
+                    Err(_) => {
+                        self.local.pending_responses.lock().remove(&id);
+                        Err(RpcError::built_in(RpcErrorCode::ResponseTimeout, None))
+                    }
+                }
+            }
```
theomonnom: Can you cleanup this code, all the results makes it hard to read

livekit-ffi/src/server/mod.rs
```diff
@@ -82,6 +80,7 @@ pub struct FfiServer {
     config: Mutex<Option<FfiConfig>>,
     logger: &'static logger::FfiLogger,
     handle_dropped_txs: DashMap<FfiHandleId, Vec<oneshot::Sender<()>>>,
+    rpc_method_invocation_waiters: Mutex<HashMap<u64, oneshot::Sender<Result<String, RpcError>>>>,
```
theomonnom: This shouldn't be inside the FfiServer, I think this should be inside FfiRoom

livekit/src/room/participant/rpc.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Maximum payload size in bytes
+pub const MAX_PAYLOAD_BYTES: usize = 15360; // 15 KB
+
+/// Calculate the byte length of a string
+pub(crate) fn byte_length(s: &str) -> usize {
+    s.as_bytes().len()
+}
+
+/// Truncate a string to a maximum number of bytes
+pub(crate) fn truncate_bytes(s: &str, max_bytes: usize) -> String {
```
theomonnom, line 99: Do we need this? I feel like the error doesn't need to contains the message data

bcherry: Errors need to be able to contain data, which is useful to e.g. a validation error where you list the specific fields that failed validation. I thought it's better to automatically truncate error message data rather than throwing a different error complaining that your error was malformed...

livekit-ffi/src/server/mod.rs
bcherry: done

livekit/src/room/participant/local_participant.rs
bcherry: yeah good call - i just flattened this out and its much clearer.

bcherry:
@theomonnom I believe this is ready for re-review

livekit-ffi/protocol/ffi.proto
```diff
@@ -96,9 +98,14 @@ message FfiRequest {
     NewSoxResamplerRequest new_sox_resampler = 33;
     PushSoxResamplerRequest push_sox_resampler = 34;
     FlushSoxResamplerRequest flush_sox_resampler = 35;
-
     SendChatMessageRequest send_chat_message = 36;
     EditChatMessageRequest edit_chat_message = 37;
+
+    // RPC
+    PerformRpcRequest perform_rpc = 38;
```
theomonnom, line 105: what do you think about naming this method send_rpc ?

livekit/src/room/participant/local_participant.rs
```diff
@@ -51,6 +67,9 @@ struct LocalEvents {
 struct LocalInfo {
     events: LocalEvents,
     encryption_type: EncryptionType,
+    pending_acks: Mutex<HashMap<String, oneshot::Sender<()>>>,
+    pending_responses: Mutex<HashMap<String, oneshot::Sender<Result<String, RpcError>>>>,
+    rpc_handlers: Mutex<HashMap<String, RpcHandler>>,
```
theomonnom: Can you reduce the amount of Mutex here?

livekit/src/room/participant/local_participant.rs
```diff
@@ -483,6 +506,96 @@ impl LocalParticipant {
...
+        &self,
+        destination_identity: String,
+        id: String,
+        method: String,
+        payload: String,
+        response_timeout_ms: u32,
+        version: u32,
+    ) -> RoomResult<()> {
+        let destination_identities: Vec<String> =
+            [destination_identity].into_iter().map(Into::into).collect();
```
theomonnom: seems like you could just use the vec![] macro

livekit/src/room/participant/local_participant.rs
```diff
@@ -544,4 +657,194 @@ impl LocalParticipant {
     pub fn kind(&self) -> ParticipantKind {
         self.inner.info.read().kind
     }
+
+    pub async fn perform_rpc(
+        &self,
+        destination_identity: String,
+        method: String,
+        payload: String,
+        response_timeout_ms: Option<u32>,
```
theomonnom, line 651: since this is the Rust API, I think it would be better to have a RPCData struct? Or maybe a RPCOptions?

livekit/src/room/participant/local_participant.rs
```diff
@@ -544,4 +657,194 @@ impl LocalParticipant {
...
+
+        let handler = self.local.rpc_handlers.lock().get(&method).cloned();
+
+        let caller_identity_2 = caller_identity.clone();
+        let request_id_2 = request_id.clone();
+
+        let response = match handler {
+            Some(handler) => {
+                match tokio::task::spawn(async move {
+                    match panic::catch_unwind(AssertUnwindSafe(|| {
```
theomonnom: mmh we shouldn't use a panic handler here, it is too opiniated

theomonnom: Rust doesn't raise exceptions, but rather we return Results

livekit/src/room/participant/local_participant.rs
```diff
@@ -544,4 +657,194 @@ impl LocalParticipant {
...
+        }
+
+        let handler = self.local.rpc_handlers.lock().get(&method).cloned();
+
+        let caller_identity_2 = caller_identity.clone();
+        let request_id_2 = request_id.clone();
+
+        let response = match handler {
+            Some(handler) => {
+                match tokio::task::spawn(async move {
```
theomonnom, line 801: This is blocking?

livekit/src/room/participant/local_participant.rs
```diff
@@ -15,27 +15,43 @@
...
+type RpcHandler = Arc<
+    dyn Fn(
+            String,              // request_id
+            ParticipantIdentity, // caller_identity
+            String,              // payload
+            Duration,            // response_timeout_ms
+        ) -> Pin<Box<dyn Future<Output = Result<String, RpcError>> + Send>>
+        + Send
+        + Sync,
+>;
```
theomonnom, line 47: I see that we expect it to returns a Future but where are we awaiting it?

Also Send + Sync seems like a hard requirement

livekit-ffi/protocol/ffi.proto
bcherry: I went back and forth on this but I think `perform` is clearer because it's a method invocation that will wait for a response. `send` feels more like a one-way method. Why do you think it should be called `send`?

livekit/src/room/participant/local_participant.rs
bcherry: I think it's important to ensure that the caller on the other side of the RPC gets notified when an error occurs at the destination that prevents a normal response from being returned (similar to an HTTP 500 error). To achieve that, the framework must catch any kind of unhandled error in the response handler so that it can generate the error response (called `1500 ApplicationError` in the RPC framework). Maybe we should re-throw the exception afterwards so the developer can still collect them? Of course if they have their own error handling try/catch construct within their handler, then they can catch and return any error response they'd like. This all applies only to uncaught errors coming from their handler.

That's the language agnostic above - it will differ a bit per platform - but in Rust I think this means we should be catching panics. If we wanted to implement "rethrow" then my understanding is we'd use `resume_unwind` in Rust but I don't have experience with it.

What do you think?

theomonnom: This is only OK if the panic configuration is "unwind". Not all panic will go into this function. 
Generally in Rust, you never really "try-catch" an error this way.
The right way would be to run the user function and expect it to return a Result. If it is an Err, then you can send 500.

In Rust, when a panic happens, it generally means that something much worse is happening.

livekit/src/room/participant/local_participant.rs
bcherry: yeah I moved these into a single object called RpcState and put a mutex on that

livekit/src/room/participant/local_participant.rs
bcherry: yeah the handler already is `Result<String, RpcError>`. That lets the user send custom errors through (you want them to be able to send specific errors but also you want to turn errors they didn't intend into generic errors so you don't leak private info across the channel).

take a look at the example, which [has a test case for this](https://github.com/livekit/rust-sdks/pull/461/files#diff-9584e9c9d2b409f287a1e1703a3760bb2a7b557119a86a4571c5dca6aef5f9f3R133). in this example, there is a division function exposed to RPC. but the caller sends 0 in the divisor. because the division function is poorly implemented and doesn't check for that, it attempts the divide by zero and causes a panic. this seems like a pretty normal thing to mess up - i'm fine with letting the local program crash and the dev clean it up, but I don't like the idea of the remote caller being stuck waiting for a response that will never come.

livekit/src/room/participant/local_participant.rs
bcherry: yeah makes sense - i just went ahead and cipied how sipdtmf, chat, and transcription were doing it

livekit/src/room/participant/local_participant.rs
bcherry: here https://github.com/livekit/rust-sdks/pull/461/files#diff-0e573daa2222454b9b896dc8d914c5a6bb4e1bf442f2200407531a1cb4ff7b7bR813

but i went back over the implementation and found that the catch_unwind isn't actually doing anything anyways, and the example works correctly without it. removing it makes this part clearer, the handler is now awaited directly `handler(...).await`

https://github.com/livekit/rust-sdks/pull/461/files#diff-0e573daa2222454b9b896dc8d914c5a6bb4e1bf442f2200407531a1cb4ff7b7bR804

livekit/src/room/participant/local_participant.rs
bcherry: ok so i just went back over this and found that catch_unwind was unnecessary to satisfy the example. the existing tokio::task::spawn behavior was already doing it and returning an Err.  simplified

livekit/src/room/participant/local_participant.rs
bcherry: (send + sync discussed offline and we decided to leave it as-is)

livekit-ffi/src/server/participant.rs
```diff
@@ -0,0 +1,189 @@
...
+        let local = match &self.participant {
+            Participant::Local(local) => local.clone(),
+            Participant::Remote(_) => {
+                return Err(FfiError::InvalidRequest("Expected local participant".into()))
+            }
+        };
+
+        let local_participant_handle = self.handle.clone();
+        let room: Arc<RoomInner> = self.room.clone();
+        let handle = server.async_runtime.spawn(async move {
```
theomonnom: any reason why this is async?

livekit-ffi/src/server/participant.rs
```diff
@@ -0,0 +1,189 @@
...
+                            )
+                            .await
+                        }
+                    })
+                },
+            );
+
+            let callback = proto::RegisterRpcMethodCallback { async_id };
+
+            let _ = server.send_event(proto::ffi_event::Message::RegisterRpcMethod(callback));
```
theomonnom: seems like this isn't needed?

livekit-ffi/src/server/participant.rs
bcherry: ah yeah you're right - the first one does need to be async and i just followed the same pattern but i can remove it, and you're also right (below) that the callbacks have no useful information so I can remove them.

livekit-ffi/src/server/participant.rs
bcherry: removed

livekit-ffi/src/server/requests.rs
```diff
@@ -788,6 +790,65 @@ fn on_flush_sox_resampler(
...
+    server: &'static FfiServer,
+    request: proto::RpcMethodInvocationResponseRequest,
+) -> FfiResult<proto::RpcMethodInvocationResponseResponse> {
+    let async_id = server.next_id();
+    let ffi_participant =
+        server.retrieve_handle::<FfiParticipant>(request.local_participant_handle)?.clone();
+
+    let room = ffi_participant.room;
+
+    let handle = server.async_runtime.spawn(async move {
```
theomonnom: seems like async isn't needed here as well?

livekit-ffi/src/server/requests.rs
```diff
@@ -788,6 +790,65 @@ fn on_flush_sox_resampler(
...
+            } else {
+                Ok(request.payload.unwrap_or_default())
+            };
+            let _ = waiter.send(result);
+        } else {
+            error = Some("No caller found".to_string());
+        }
+
+        let callback = proto::RpcMethodInvocationResponseCallback { async_id, error };
+        let _ = server.send_event(proto::ffi_event::Message::RpcMethodInvocationResponse(callback));
```
theomonnom: same for this message?

bcherry: you're right - i think I convinced myself that the ffi response shouldn't send anything besides async_id but that's not true, just how it worked in many of the examples i saw. i merged the error into the response itself and returned that directly with no callback.

livekit/src/room/participant/rpc.rs
theomonnom: wdyt about publish_rpc? since we already have publish_data

livekit/src/rtc_engine/mod.rs
```diff
@@ -294,6 +311,10 @@ impl RtcEngine {
     pub fn session(&self) -> Arc<RtcSession> {
         self.inner.running_handle.read().session.clone()
     }
+
+    pub fn get_latest_join_response(&self) -> proto::JoinResponse {
+        self.inner.latest_join_response.clone()
+    }
 }
```
theomonnom: Seems like this is unused,  can you remove

livekit/src/rtc_engine/mod.rs
```diff
@@ -167,6 +183,7 @@ struct EngineInner {
     lk_runtime: Arc<LkRuntime>,
     engine_tx: EngineEmitter,
     options: EngineOptions,
+    latest_join_response: proto::JoinResponse,
```
theomonnom: for this field

livekit/src/rtc_engine/mod.rs
bcherry: it's used here https://github.com/livekit/rust-sdks/pull/461/files#diff-0e573daa2222454b9b896dc8d914c5a6bb4e1bf442f2200407531a1cb4ff7b7bR660

theomonnom (approved):

livekit/src/rtc_engine/mod.rs
theomonnom: The link doesn't work for me

livekit/src/room/participant/local_participant.rs
```diff
@@ -544,4 +642,205 @@ impl LocalParticipant {
...
+        response_timeout_ms: Option<u32>,
+    ) -> Result<String, RpcError> {
+        let response_timeout = Duration::from_millis(response_timeout_ms.unwrap_or(10000) as u64);
+        let max_round_trip_latency = Duration::from_millis(2000);
+
+        if payload.len() > MAX_PAYLOAD_BYTES {
+            return Err(RpcError::built_in(RpcErrorCode::RequestPayloadTooLarge, None));
+        }
+
+        if let Some(server_info) = self.inner.rtc_engine.get_latest_join_response().server_info {
```
theomonnom: ah ok, so you should use the one inside signal_client, there is already the last join response there

theomonnom: https://github.com/livekit/rust-sdks/blob/3de6fa60f3cdafc2c106a7f370724eddf1e92fdc/livekit-api/src/signal_client/mod.rs#L164

bcherry: oh yeah perfect - i was looking for something called "latest_join_response" like in JS so i missed that. i can access it here as `self.inner.rtc_engine.session().signal_client().join_response()`

livekit/src/room/participant/rpc.rs
bcherry: gonna have a last check-in with @davidzhao on the name of this feature before merging the initial branches and publishing the forthcoming docs. will let you know where it lands!

bcherry: I went back and forth but I think rebranding this away from "RPC" isn't worth it and I think "perform" is still a pretty good verb overall for the primary action so going to leave as-is and focus on shipping it.

examples/rpc/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+        Err(e) => log::error!("[{}] RPC call failed: {:?}", elapsed_time(), e),
+    }
+    Ok(())
+}
+
+async fn perform_quantum_hypergeometric_series(
+    room: &Arc<Room>,
+) -> Result<(), Box<dyn std::error::Error>> {
+    println!("[{}] What's the quantum hypergeometric series of 42?", elapsed_time());
```
n8agrin, line 180: What the heck :) 

image

bcherry: 👋🏻

PR #463, opened 2024-10-08
Add the clippy linter to CI

bcherry:
We've got a good deal of linter churn, like unused or ungrouped imports, unnecessary casts, unnecessary name repetition etc.

This runs the clippy linter in CI and I also ran it once on the while project with `cargo clippy --fix`. This should keep project churn down going forward.

Open to other thoughts if anyone has them!

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,4 +1,4 @@
-name: Rust Formatting
+name: Rust Formatting & Linting
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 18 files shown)


nbsp:
we should probably go about this more methodically, figuring out exactly which clippy things we want
there's some stuff from clippy::pedantic that would be good to have, for example
a thing i thought of was maybe disallowing all `println!()` calls by default to avoid loose debug lines

PR #464, opened 2024-10-08
Fix enconding presets

iparaskev:
Fixes an issue with compute_appropriate_encoding, where the last preset in the list with the highest FPS wasn't considered.

Changes the order in into_rtp_encodings to have the highest quality first.

### DIFF
--- a/livekit/src/room/options.rs
+++ b/livekit/src/room/options.rs
@@ -179,7 +179,7 @@ pub fn compute_appropriate_encoding(
 
     for preset in presets {
         encoding = preset.encoding.clone();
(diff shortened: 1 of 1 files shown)


iparaskev:
Hi, first of all thanks for the project. 

I have a question regarding the "rid" when the presets are only 2 in `into_rtp_encodings`. Should they be 
1. "q" and "h"
2. "q" and "f" 
3. "h" and "f"

?

Thanks

theomonnom (approved):

theomonnom:
Thanks!

PR #470, opened 2024-10-15
Use proto2 syntax for FFI

lukasIO:


livekit-ffi/src/conversion/room.rs
```diff
@@ -156,7 +160,7 @@ impl From<proto::RoomOptions> for RoomOptions {
     fn from(value: proto::RoomOptions) -> Self {
         let e2ee = value.e2ee.and_then(|opts| {
             let encryption_type = opts.encryption_type();
-            let Some(provider_opts) = opts.key_provider_options else {
+            let provider_opts = opts.key_provider_options else {
                 return None;
             };
```
nbsp: this let-else is unnecessary

```suggestion
            let provider_opts = opts.key_provider_options;
```

livekit-ffi/src/server/audio_source.rs
```diff
@@ -75,7 +72,7 @@ impl FfiAudioSource {
         server: &'static server::FfiServer,
         capture: proto::CaptureAudioFrameRequest,
     ) -> FfiResult<proto::CaptureAudioFrameResponse> {
-        let Some(buffer) = capture.buffer else {
+        let buffer = capture.buffer else {
             return Err(FfiError::InvalidRequest("buffer is None".into()));
         };
```
nbsp:
```suggestion
        let buffer = capture.buffer;
```

livekit-ffi/src/server/requests.rs
```diff
@@ -391,7 +394,7 @@ unsafe fn on_video_convert(
     server: &'static FfiServer,
     video_convert: proto::VideoConvertRequest,
 ) -> FfiResult<proto::VideoConvertResponse> {
-    let Some(ref buffer) = video_convert.buffer else {
+    let ref buffer = video_convert.buffer else {
         return Err(FfiError::InvalidRequest("buffer is empty".into()));
     };
```
nbsp:
```suggestion
    let ref buffer = video_convert.buffer;
```

nbsp (approved):

theomonnom (commented):
Is there any change required on the python side?

livekit-ffi/protocol/room.proto
```diff
@@ -26,215 +26,240 @@ import "stats.proto";
...
-  uint64 async_id = 1;
-  optional string error = 2;
-  OwnedRoom room = 3;
-  OwnedParticipant local_participant = 4;
-  repeated ParticipantWithTracks participants = 5;
+  message Result {
+    required OwnedRoom room = 1;
+    required OwnedParticipant local_participant = 2;
+    repeated ParticipantWithTracks participants = 3;
+  }
```
theomonnom, line 49: This doesn't seems right.
You do receive everything

livekit-ffi/protocol/room.proto
```diff
@@ -26,215 +26,240 @@ import "stats.proto";
...
+
+  required uint64 async_id = 1;
+  oneof message {
+    string error = 2;
```
theomonnom, line 53: There is no optional value?

lukasIO:
yeah, there's a couple of messages (callbacks with results) that I changed to be a oneof message of either error or result, so that's a change that has to be accounted for

lukasIO:
I think this is what's needed: https://github.com/livekit/python-sdks/pull/290 (some type additions in there to make catching errors easier)

theomonnom:
nice, any drawbacks we should be aware of when using the proto2 syntax?
e.g performance decrease, ..
otherwise lgtm, this is much cleaner on the rust side!

lukasIO:
not that I'm aware of, though performance will always depend on the language specific encoders and decoders

bcherry:
next time can we wait to merge something substantial like this until we have updates ready to go for all of the dependent SDKs? this is now blocking development of new FFI-based change in node and python

PR #471, opened 2024-10-17
Add SDK version, allow override of SDK and version from FFI

bcherry:
Two parts here:

1. Send the rust SDK version for the RTC client
2. Allow FFI-based clients to override SDK and version

### DIFF
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -281,7 +281,8 @@ impl SignalInner {
         let sid = &self.join_response.participant.as_ref().unwrap().sid;
         let token = self.token.lock().clone();
 
-        let mut lk_url = get_livekit_url(&self.url, &token, &self.options, &self.analytics_options).unwrap();
+        let mut lk_url =
+            get_livekit_url(&self.url, &token, &self.options, &self.analytics_options).unwrap();
         lk_url.query_pairs_mut().append_pair("reconnect", "1").append_pair("sid", sid);
 
         let (new_stream, mut events) = SignalStream::connect(lk_url).await?;
@@ -425,7 +426,12 @@ fn is_queuable(signal: &proto::signal_request::Message) -> bool {
     )
 }
 
-fn get_livekit_url(url: &str, token: &str, options: &SignalOptions, analytics_options: &SignalAnalyticsOptions) -> SignalResult<url::Url> {
+fn get_livekit_url(
+    url: &str,
+    token: &str,
+    options: &SignalOptions,
+    analytics_options: &SignalAnalyticsOptions,
+) -> SignalResult<url::Url> {
     let mut lk_url = url::Url::parse(url).map_err(|err| SignalError::UrlParse(err.to_string()))?;
 
     if !lk_url.has_host() {
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -65,6 +65,7 @@ pub enum SignalError {
 }
 
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct SignalSdkOptions {
     pub sdk: String,
     pub sdk_version: Option<String>,
@@ -77,6 +78,7 @@ impl Default for SignalSdkOptions {
 }
 
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct SignalOptions {
     pub auto_subscribe: bool,
     pub adaptive_stream: bool,
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -281,8 +281,7 @@ impl SignalInner {
         let sid = &self.join_response.participant.as_ref().unwrap().sid;
         let token = self.token.lock().clone();
 
-        let mut lk_url =
-            get_livekit_url(&self.url, &token, &self.options, &self.analytics_options).unwrap();
+        let mut lk_url = get_livekit_url(&self.url, &token, &self.options, &self.analytics_options).unwrap();
         lk_url.query_pairs_mut().append_pair("reconnect", "1").append_pair("sid", sid);
 
         let (new_stream, mut events) = SignalStream::connect(lk_url).await?;
@@ -426,12 +425,7 @@ fn is_queuable(signal: &proto::signal_request::Message) -> bool {
     )
 }
 
-fn get_livekit_url(
-    url: &str,
-    token: &str,
-    options: &SignalOptions,
-    analytics_options: &SignalAnalyticsOptions,
-) -> SignalResult<url::Url> {
+fn get_livekit_url(url: &str, token: &str, options: &SignalOptions, analytics_options: &SignalAnalyticsOptions) -> SignalResult<url::Url> {
     let mut lk_url = url::Url::parse(url).map_err(|err| SignalError::UrlParse(err.to_string()))?;
 
     if !lk_url.has_host() {
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -74,12 +74,7 @@ pub struct SignalOptions {
 
 impl Default for SignalOptions {
     fn default() -> Self {
-        Self {
-            auto_subscribe: true,
-            adaptive_stream: false,
-            sdk: "rust",
-            sdk_version: None,
-        }
+        Self { auto_subscribe: true, adaptive_stream: false, sdk: "rust", sdk_version: None }
     }
 }
 
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -64,12 +64,27 @@ pub enum SignalError {
     SendError,
 }
 
+
+#[derive(Debug, Clone)]
+pub struct SignalSdkOptions {
+    pub sdk: String,
+    pub sdk_version: Option<String>,
+}
+
+impl Default for SignalSdkOptions {
+    fn default() -> Self {
+        Self {
+            sdk: "rust".to_string(),
+            sdk_version: None,
+        }
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct SignalOptions {
     pub auto_subscribe: bool,
     pub adaptive_stream: bool,
-    pub sdk: String,
-    pub sdk_version: Option<String>,
+    pub sdk_options: SignalSdkOptions,
 }
 
 impl Default for SignalOptions {
@@ -77,8 +92,7 @@ impl Default for SignalOptions {
         Self {
             auto_subscribe: true,
             adaptive_stream: false,
-            sdk: "rust".to_string(),
-            sdk_version: None,
+            sdk_options: SignalSdkOptions::default(),
         }
     }
 }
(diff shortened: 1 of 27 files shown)


livekit-ffi/src/cabi.rs
```diff
@@ -15,13 +15,23 @@ pub type FfiCallbackFn = unsafe extern "C" fn(*const u8, usize);
 ///
 /// The foreign language must only provide valid pointers
 #[no_mangle]
-pub unsafe extern "C" fn livekit_ffi_initialize(cb: FfiCallbackFn, capture_logs: bool) {
+pub unsafe extern "C" fn livekit_ffi_initialize(
+    cb: FfiCallbackFn,
+    capture_logs: bool,
+    sdk: &'static [u8; 16],
+    sdk_version: &'static [u8; 16],
```
bcherry: i got warnings when using str or String here that they aren't FFI-safe so I used `[u8; 16]` which meant i also needed to handle nulls below so I made sdk and sdk version optional. I'm going to go ahead and do the python and maybe the unity implementation before I merge this, since I believe this bit here is used in those but not node, so its not tested yet

livekit-ffi/src/cabi.rs
```diff
@@ -15,13 +15,23 @@ pub type FfiCallbackFn = unsafe extern "C" fn(*const u8, usize);
...
 #[no_mangle]
-pub unsafe extern "C" fn livekit_ffi_initialize(cb: FfiCallbackFn, capture_logs: bool) {
+pub unsafe extern "C" fn livekit_ffi_initialize(
+    cb: FfiCallbackFn,
+    capture_logs: bool,
+    sdk: Option<&'static [u8; 16]>,
+    sdk_version: Option<&'static [u8; 16]>,
+) {
+    let sdk = std::str::from_utf8(sdk.unwrap()).unwrap().trim_end_matches('\0');
+    let sdk_version = std::str::from_utf8(sdk_version.unwrap()).unwrap().trim_end_matches('\0');
```
bcherry: i got warnings when using str or String here that they aren't FFI-safe so I used [u8; 16] which meant i also needed to handle nulls below so I made sdk and sdk version optional. I'm going to go ahead and do the python and maybe the unity implementation before I merge this, since I believe this bit here is used in those but not node, so its not tested yet

theomonnom: You should use a standard "C" null-terminated string

theomonnom:
```
use std::os::raw::c_char;

sdk: *const c_char
```

livekit/src/room/mod.rs
```diff
@@ -328,6 +330,16 @@ impl Room {
         url: &str,
         token: &str,
         options: RoomOptions,
+    ) -> RoomResult<(Self, mpsc::UnboundedReceiver<RoomEvent>)> {
+        Self::connect_with_sdk(url, token, options, Some("rust"), Some(SDK_VERSION)).await
+    }
+
+    pub async fn connect_with_sdk(
```
theomonnom: I think you can add those options inside `RoomOptions`. No need to create a new method. Let's keep everything inside the `connect` one?
Eitherway the fields will be public, there is no straightforward way to hide sdk & sdk_version to the users

bcherry: it felt extra weird inside `RoomOptions` to me... but yeah I can change it back

livekit-ffi/src/cabi.rs
bcherry: thanks that worked great paired with `lib::ffi::CStr`

livekit-api/src/signal_client/mod.rs
```diff
@@ -68,11 +68,13 @@ pub enum SignalError {
 pub struct SignalOptions {
     pub auto_subscribe: bool,
     pub adaptive_stream: bool,
+    pub sdk: &'static str,
+    pub sdk_version: Option<&'static str>,
```
theomonnom:
```suggestion
    pub sdk: String,
    pub sdk_version: Option<String>,
```

livekit/src/room/mod.rs
```diff
@@ -328,6 +330,16 @@ impl Room {
...
+    ) -> RoomResult<(Self, mpsc::UnboundedReceiver<RoomEvent>)> {
+        Self::connect_with_sdk(url, token, options, "rust", SDK_VERSION).await
+    }
+
+    pub async fn connect_with_sdk(
+        url: &str,
+        token: &str,
+        options: RoomOptions,
+        sdk: &'static str,
+        sdk_version: &'static str,
```
theomonnom:
```suggestion
        sdk: &str,
        sdk_version: &str,
```

livekit-ffi/src/cabi.rs
```diff
@@ -18,20 +20,17 @@ pub type FfiCallbackFn = unsafe extern "C" fn(*const u8, usize);
...
-
     FFI_SERVER.setup(FfiConfig {
         callback_fn: Arc::new(move |event| {
             let data = event.encode_to_vec();
             cb(data.as_ptr(), data.len());
         }),
         capture_logs,
-        sdk: Some(sdk),
-        sdk_version: Some(sdk_version),
+        sdk: CStr::from_ptr(sdk).to_str().ok(),
```
theomonnom: this isn't safe, we have no guarantee the pointers will still be alive.
they will be used after free

livekit/src/room/mod.rs
bcherry: @theomonnom I started on changing it back and found it kinda spiraled out pretty quick, RoomOptions is used a lot more places than need sdk and sdk_version, so there was a lot of unrelated surface area doing it that way. I decided to make a new options type called RoomAnalyticsOptions to contain them, to be passed alongside RoomOptions. And I decided to do the same thing for SignalClient.  It makes its purpose clearer. Maybe we could just change Room::connect  to take a new ConnectOptions instead of a RoomOptions, and ConnectOptions would include `roomOptions`, `sdk`, and `sdk_version`.

What do you think of this approach?

livekit-ffi/src/cabi.rs
bcherry: Makes sense - I'm not too familiar with this but cursor suggested  using `.to_string_lossy().into_owned()`:

```
sdk: CStr::from_ptr(sdk).to_string_lossy().into_owned(),
sdk_version: Some(CStr::from_ptr(sdk_version).to_string_lossy().into_owned()),
```

does that make sense?

livekit/src/room/mod.rs
bcherry: the other problem with this is it becomes a breaking change in the rust RTC library, just so that we can add some internal stuff for FFI. IMO better to have a separate method that includes the sdk overrides and is marked for FFI use only?

livekit-api/src/signal_client/mod.rs
```diff
@@ -111,14 +112,27 @@ impl Debug for SignalClient {
...
+        Self { sdk: "rust".to_string(), sdk_version: None }
+    }
+}
+
 impl SignalClient {
     pub async fn connect(
         url: &str,
         token: &str,
         options: SignalOptions,
+        analytics_options: SignalAnalyticsOptions,
```
theomonnom: It'll break the API if we add this new arg

davidzhao: let's put this inside of SignalOptions as a new optional field

bcherry: I had it that way originally - and happy to change it back - but FWIW it would break the API that way, too, depending on how you initialized your SignalOptions.

I think we should consider introducing this as a completely non breaking change with secondary versions of Room.connect and SignalClient.connect that take an extra argument to override these, and mark them as for internal use even though they're in the public SDK. it would minimize surface area of the change, while more cleanly delineating what these parameters are for.

Same basic thoughts in https://github.com/livekit/rust-sdks/pull/471#discussion_r1806794174, looking for feedback then I can finalize this PR @theomonnom @davidzhao

bcherry: For now I reverted the change made in https://github.com/livekit/rust-sdks/pull/471/commits/ce98197020136a37d6f44813b7cbe24af45392f6 so this is now back to the original proposal from an interface perspective:

FFI: The new parameters are required in `livekit_ffi_initialize`

Room: `Room::connect` is unchanged and a new `Room::connect_with_sdk` method is added to allow overriding the SDK. This keeps breaking changes out of the RTC package and avoids changes RoomOptions which are used in many places.

SignalClient: `SignalOptions` is extended to take sdk and sdk_version. This will be a breaking change for some users, depending on how they initialize the options, but within the API package is unlikely to cause issues as this project likely has very limited usage besides as a dependency of the RTC package.

livekit-api/src/signal_client/mod.rs
```diff
@@ -68,11 +68,18 @@ pub enum SignalError {
 pub struct SignalOptions {
     pub auto_subscribe: bool,
     pub adaptive_stream: bool,
+    pub sdk: String,
+    pub sdk_version: Option<String>,
```
theomonnom: let's put both of them to Option by default, just for consistency
```suggestion
    pub sdk: Option<String>,
    pub sdk_version: Option<String>,
```

theomonnom (approved):

livekit/src/room/mod.rs
```diff
@@ -328,6 +330,16 @@ impl Room {
         url: &str,
         token: &str,
         options: RoomOptions,
+    ) -> RoomResult<(Self, mpsc::UnboundedReceiver<RoomEvent>)> {
+        Self::connect_with_sdk(url, token, options, "rust", SDK_VERSION).await
+    }
+
+    pub async fn connect_with_sdk(
```
theomonnom: I would put sdk and sdk_version inside RoomOptions

livekit-ffi/src/server/room.rs
```diff
@@ -116,8 +116,24 @@ impl FfiRoom {
...
         let async_id = server.next_id();
 
+        let sdk;
+        let sdk_version;
+        {
+            let config = server.config.lock();
+            sdk = config.as_ref().map(|c| c.sdk.clone()).unwrap();
+            sdk_version = config.as_ref().map(|c| c.sdk_version.clone()).unwrap();
+        }
+
```
theomonnom:
```suggestion
        let (sdk, sdk_version) = {
            let config = server.config.lock();
            (config.as_ref().map(|c| c.sdk.clone()).unwrap(), config.as_ref().map(|c| c.sdk_version.clone()).unwrap())
        }

```

theomonnom: let's verify they're not empty instead of unrwap?

livekit-api/src/signal_client/mod.rs
theomonnom: it's more unclear to ppl to know what connect_with_sdk is doing. I would vote to put that inside RoomOptions

theomonnom: I think the right solution would have been to add `#[non_exhaustive]` to the Options struct in the earlier versions

livekit/src/room/mod.rs
bcherry: after further discussion with @theomonnom I've updated this to just include it in RoomOptions and SignalOptions. I chose to include it as a nested type rather than individual fields, as I anticipate adding at least one more field to it soon (for the `other_sdks` field to include the versions of associated sdks in use)

livekit-api/src/signal_client/mod.rs
bcherry: after further discussion with @theomonnom I've updated this to just include it in RoomOptions and SignalOptions. I chose to include it as a nested type rather than individual fields, as I anticipate adding at least one more field to it soon (for the other_sdks field to include the versions of associated sdks in use)

livekit-api/src/signal_client/mod.rs
bcherry: hmm I don't really want to... there wasn't an existing case where the sdk wasn't set already

livekit-api/src/signal_client/mod.rs
theomonnom: Can you make the SignalOptions & RoomOptions #[non_exhaustive] instead? (You can keep the inner type, but it's going to be more future proof this way)

livekit-ffi/src/server/room.rs
bcherry: yeah i reworked this and it turned out they would be non-empty, the problem was simply whether the config existed or not. i redid it to be more clear and it won't use them unless the config is non-empty

livekit-api/src/signal_client/mod.rs
bcherry: yeah sounds good - added that. i made the inner type non-exhaustive too so it won't be breaking later

PR #476, opened 2024-10-28
Check RPC version

bcherry:
This rejects requests with the wrong version, which we hopefully don't need to use but does allow us optionality in the future to change the protocol in a breaking way.

### DIFF
--- a/livekit/src/room/participant/local_participant.rs
+++ b/livekit/src/room/participant/local_participant.rs
@@ -786,6 +786,7 @@ impl LocalParticipant {
         method: String,
         payload: String,
         response_timeout_ms: u32,
+        version: u32,
     ) {
         if let Err(e) = self
             .publish_rpc_ack(RpcAck {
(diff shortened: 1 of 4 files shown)


theomonnom (approved):

livekit/src/room/participant/local_participant.rs
```diff
@@ -643,17 +637,10 @@ impl LocalParticipant {
...
-    pub async fn perform_rpc(
-        &self,
-        destination_identity: String,
-        method: String,
-        payload: String,
-        response_timeout_ms: Option<u32>,
-    ) -> Result<String, RpcError> {
-        let response_timeout = Duration::from_millis(response_timeout_ms.unwrap_or(10000) as u64);
+    pub async fn perform_rpc(&self, params: PerformRpcParams) -> Result<String, RpcError> {
         let max_round_trip_latency = Duration::from_millis(2000);
```
theomonnom, line 641: PerformRpcParams -> RpcData?

bcherry: I don't like it because I think we need to be clear this is used on the perform side, in contrast to RpcInvocationData which is used on the receive side...

bcherry: (still could be PerformRpcData if that feels more correct)

PR #483, opened 2024-11-09
simplify version cascading by using workspaces

nbsp:
this removes the need to go through every Cargo.toml file to fix dependency versions.
CI will complain if it's not set in the one place, and once that's done everything else should work.

when i'm free i'll refactor nanpa to utilize this, so we can actually start using it properly. in the meantime, nanpa bumps would require:

- adding a nanpa changeset file
- bumping Cargo.toml ← this constraint will be removed
- bumping root Cargo.toml ← this constraint will be removed


nbsp:
nanpa 0.2.0 has been released and it handles all of this automatically now. from now on, on every changing PR, add `.nanpa/something.kdl` that looks like:

```
minor type="added" package="libwebrtc" "add thing to stuff"
```

can be major/minor/patch, and the type can be any from Keep a Changelog https://keepachangelog.com

bcherry (approved):
nice - this is a lot simpler

theomonnom (approved):

PR #485, opened 2024-11-13
Use rust::String::lossy for stringification to avoid crashing due to non-utf8 content

mgsloan:
This was causing crashes [here](https://github.com/zed-industries/zed/blob/76936fa5b44103136375a823e5b76476ae82683e/crates/live_kit_client/examples/test_app.rs#L106) in Zed's live kit test app when run on Linux.


theomonnom (approved):

theomonnom:
Thanks!

PR #487, opened 2024-11-15
Add support for swapping out the HTTP Client 

mikayla-maki:
This PR updates the http_client to follow the `dispatcher` pattern.


mikayla-maki:
Note that the `dispatcher` feature we use here was broken by the inclusion of "tokio" as a default runtime feature in: https://github.com/livekit/rust-sdks/pull/495 combined with cargo always including default features from: https://github.com/livekit/rust-sdks/pull/483

However, after fixing those bugs the `livekit-runtime` crate's default features are still being included. Haven't quite figured out why yet...

mikayla-maki:
We no longer need this feature.

PR #493, opened 2024-11-25
Fix build of examples

mgsloan:
`wgpu_room` was broken when `#[non_exhaustive]` was added to some structs in #471.

`save_to_disk` was broken when resampling was added in #412.


nbsp (approved):

mgsloan:
Looks like compilation was fixed instead in #530 and some other commits

PR #501, opened 2024-12-02
ci: Fix wonky FFI builds

otoayana:
The FFI builds have been failing due to the WebRTC license file not existing. Since that file is only generated when compiling to Android as a target, this PR ensures the file exists in that location, and swaps double quotation marks for single ones.

### DIFF
--- a/.github/workflows/ffi-builds.yml
+++ b/.github/workflows/ffi-builds.yml
@@ -165,10 +165,11 @@ jobs:
 
       - name: Copy/Build licenses
         run: |
-          echo "# livekit" > TEMP_LICENSE.md
-          echo "```" >> TEMP_LICENSE.md
+          echo '# livekit' > TEMP_LICENSE.md
+          echo '```' >> TEMP_LICENSE.md
           cat LICENSE >> TEMP_LICENSE.md
-          echo "```" >> TEMP_LICENSE.md
+          echo '```' >> TEMP_LICENSE.md
+          touch livekit-ffi/WEBRTC_LICENSE.md
           cat livekit-ffi/WEBRTC_LICENSE.md >> TEMP_LICENSE.md
           mv TEMP_LICENSE.md target/${{ matrix.target }}/release/LICENSE.md
         shell: bash


tommady:
Could you take a look at [this PR](https://github.com/livekit/rust-sdks/pull/199/files#diff-3b8b1979be7be694d1abd5c00c8e860c1c9156e3c5adf16110ce473c93b51f62)? I believe it might be the root cause, but I’m curious why this .gitignore was added to exclude the LICENSE file.

.github/workflows/ffi-builds.yml
```diff
@@ -165,10 +165,11 @@ jobs:
...
       - name: Copy/Build licenses
         run: |
-          echo "# livekit" > TEMP_LICENSE.md
-          echo "```" >> TEMP_LICENSE.md
+          echo '# livekit' > TEMP_LICENSE.md
+          echo '```' >> TEMP_LICENSE.md
           cat LICENSE >> TEMP_LICENSE.md
-          echo "```" >> TEMP_LICENSE.md
+          echo '```' >> TEMP_LICENSE.md
+          touch livekit-ffi/WEBRTC_LICENSE.md
```
tommady, line 172: Regarding my comment about the .gitignore excluding the LICENSE file, simply creating an empty LICENSE file doesn’t seem like a good solution.

theomonnom:
This file should be generated for each platform: https://github.com/livekit/rust-sdks/blob/fb1efb03ff5109120ac639946c2ea3364c969aa9/webrtc-sys/libwebrtc/build_linux.sh#L126

tommady:
Ah, I understand now—thank you for taking the time to explain.

.github/workflows/ffi-builds.yml
wiredsis: Fair enough. I chose this route since, as seen [here](https://github.com/livekit/rust-sdks/blob/65a1b02a2806547fa1f77a55bab1e24eef194880/livekit-ffi/build.rs#L41), the file is only generated when the target is an Android device. Maybe a better alternative would be running this job when an Android target matches.

PR #507, opened 2024-12-07
Adding disconnect reason

davidzhao:
When a participant is disconnected, we have the reason for disconnection.

Pass through this event to FFI clients so they could determine why a RemoteParticipant had disconnected


livekit-api/src/services/sip.rs
```diff
@@ -58,18 +60,24 @@ pub struct CreateSIPTrunkOptions {
 #[derive(Default, Clone, Debug)]
 pub struct CreateSIPInboundTrunkOptions {
     /// Optional free-form metadata.
-    pub metadata: String,
+    pub metadata: Option<String>,
```
davidzhao, line 63: these aren't required, updating them to optional

livekit/src/room/mod.rs
```diff
@@ -783,6 +783,7 @@ impl RoomSession {
             let remote_participant = self.get_participant_by_sid(&participant_sid);
             if pi.state == proto::participant_info::State::Disconnected as i32 {
                 if let Some(remote_participant) = remote_participant {
+                    remote_participant.update_info(pi.clone());
```
davidzhao, line 786: even when the remote participant is disconnected, we have to update it to keep its reason

livekit-api/src/services/sip.rs
theomonnom: Generally using default values is fine in this case, but an Option is ok too

davidzhao: yeah, it's easier for users if they didn't have to fill out every field as default

PR #508, opened 2024-12-09
move libyuv/imgproc from theomonnom/mikado

theomonnom:
was initially doing a side project with it, but interested in other stuff now. So kind of abandoned


nbsp (commented):
maybe also set livekit-ffi dependencies to use relative imgproc? and imgproc to use relative yuv-sys? just for consistency with the rest of the repo

PR #509, opened 2024-12-10
fix(livekit): correctly calculate screenshare res

nbsp:
removes 1080p15 from presets


theomonnom (approved):

theomonnom (commented):
mmh tho we will not be able to get the full resolution?

nbsp:
resolution is computed separately; this just calculates the bitrate/framerate

theomonnom:
JS is doing the same: https://github.com/livekit/client-sdk-js/blob/0fe824b438cae99aa01a745c7165777a2ab4c2bb/src/room/participant/publishUtils.ts#L279

PR #522, opened 2024-12-14
ci: fix FFI and WebRTC builds with artifacts v4

nbsp:
https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact


lukasIO (approved):
thanks!

PR #532, opened 2024-12-19
Fix deadlocked engine in RPC

bcherry:
There's an issue where if you're busy handling an incoming rpc request, additional engine events will not be processed. this causes a deadlock essentially if you have nested RPC, since the incoming acks won't be processed, but also probably gums up plenty of other features of livekit...

this solution is probably not correct, but it works and i included an example. need help @typester !


typester:
Interesting. Will take a look!

typester:
The change you made (d7c8cdacf84b77e3711afe89254cf8cb7816f39b) seems like a good fix to me :+1:

typester:
What caught my attention in a different context is the `await` here: https://github.com/livekit/rust-sdks/blob/fea75571cfd021296525598a37752cc2f1fe55eb/livekit/src/room/mod.rs#L649

 If we `await` at this point, `engine_events` will be processed one by one. To me, it seems more natural not to `await` here, but what was the intention behind this design?

theomonnom:
yeah, it is expected for engine events to be processed one by one (keeping the order)

theomonnom (approved):

PR #535, opened 2024-12-20
add .nanpa folders

bcherry:
the instructions require you to add a file to the `.nanpa/` folder but these folders don't exist so requires you to first make it and then add the file.

should we just check in the folders with a `.gitkeep` to make this easier? i don't feel strongly but thought maybe its better @nbsp

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1321,7 +1321,7 @@ dependencies = [
 
 [[package]]
 name = "imgproc"
-version = "0.3.11"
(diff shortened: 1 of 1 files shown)


nbsp:
.nanpa directory at root is enough. you can have changesets for all in the root package using `package="livekit-ffi"`.

bcherry:
oh that makes sense but wasn't clear from the comment it leaves

PR #538, opened 2024-12-24
Make `deadlock_detection` an optional feature and disable it by default.

typester:
It seems that `parking_lot`'s `deadlock_detection` feature causes CPU usage to continuously increase when running for extended periods. I haven't investigated the details yet, but since this feature is likely intended for debugging, I have modified it to be disabled by default. It can now only be enabled by explicitly specifying `--features=deadlock_detection`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1321,7 +1321,7 @@ dependencies = [
 
 [[package]]
 name = "imgproc"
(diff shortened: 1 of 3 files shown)


typester:
This seems not meaningful fix. Will close at this moment.

PR #539, opened 2025-01-03
Close track FFI handles automatically when a FfiRoom is closed.

typester:
As shown in [this test script](https://gist.github.com/typester/4907bebe022ffd429eeb27eeb392b1ec), it seems that if the FFI client forgets to close Audio/Video Stream on their side, a memory leak occurs.
Looking at the Python-side code, the `__del__` method is supposed to dispose of FFI handles when objects like VideoStream are released. However, for some reason, in this test script, that part is not invoked, leaving the handles unclosed.

In this pull request, the `FfiRoom` is modified to automatically dispose of associated track handles when it is closed.

With the script mentioned above, after running it, the memory consumption of the test process decreased from around 110MB to about 80MB, indicating the change is effective. 
(The test was conducted in a local environment using a livekit-server with one video and one audio stream are running in a joined room)

### DIFF
--- a/livekit-ffi/src/server/room.rs
+++ b/livekit-ffi/src/server/room.rs
@@ -271,7 +271,7 @@ impl FfiRoom {
         for (_, &handle) in self.inner.track_handle_lookup.lock().iter() {
             if server.drop_handle(handle) {
                 // Store an empty handle for the FFI client that assumes a handle exists for this id.
-                server.store_handle(handle, vec![].into_boxed_slice());
+                server.store_handle(handle, ());
             }
         }
 
--- a/livekit-ffi/src/server/room.rs
+++ b/livekit-ffi/src/server/room.rs
@@ -267,9 +267,7 @@ impl FfiRoom {
 
     /// Close the room and stop the tasks
     pub async fn close(&self, server: &'static FfiServer) {
-        log_file(&format!("ffi_room will close: {}", self.inner.handle_id));
-
-        // close associated track handles
+        // drop associated track handles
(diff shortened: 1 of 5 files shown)


livekit-ffi/src/server/room.rs
```diff
@@ -266,7 +266,15 @@ impl FfiRoom {
     }
 
     /// Close the room and stop the tasks
-    pub async fn close(&self) {
+    pub async fn close(&self, server: &'static FfiServer) {
+        // drop associated track handles
+        for (_, &handle) in self.inner.track_handle_lookup.lock().iter() {
+            if server.drop_handle(handle) {
+                // Store an empty handle for the FFI client that assumes a handle exists for this id.
+                server.store_handle(handle, vec![].into_boxed_slice());
```
typester: The Python-side code assumes this handle ID exists, and if the ID does not exist when Python attempts to dispose of it, an assertion failure occurs. To avoid this, an empty entry is stored here.

bcherry (approved):

livekit-ffi/src/server/room.rs
theomonnom: Can we implement FfiHandle for ()?

`impl FfiHandle for () {`

typester: Let me try it

typester: @theomonnom Fixed in [d867bee](https://github.com/livekit/rust-sdks/pull/539/commits/d867bee2744d26307b33d8fa98f5103e0061e4ce)

theomonnom (approved):

PR #541, opened 2025-01-03
bump livekit-ffi to 0.12.5

typester:
Including fix for deadlock in nested RPC calls: #532


theomonnom (approved):
feel free to bump livekit too since we figured our nanpa

typester:
@theomonnom Ok, then I will bump livekit crate too. I will merge this after CI is finished

PR #544, opened 2025-01-09
fix: emit track-muted event for local tracks

kbalt:
Currently there's no way to know when a track was muted by a moderator since track changes from the server were ignored.

This updates the TrackInfo of the local participant in a similar way, to how its done for remote participants.

While this seems to work, I'm not sure if this is the correct way to do this, since the comment in https://github.com/livekit/rust-sdks/blob/main/livekit/src/room/publication/remote.rs#L157 seems to imply that this should already work for local publications?

### DIFF
--- a/.nanpa/track-muted-event-for-local-publications.kdl
+++ b/.nanpa/track-muted-event-for-local-publications.kdl
@@ -1,0 +1,1 @@
+patch type="fixed" package="livekit" "Emit TrackMuted/TrackUnmuted events for local publications"
--- a/livekit/src/room/participant/local_participant.rs
+++ b/livekit/src/room/participant/local_participant.rs
@@ -139,7 +139,14 @@ impl LocalParticipant {
     }
 
     pub(crate) fn update_info(&self, info: proto::ParticipantInfo) {
-        super::update_info(&self.inner, &Participant::Local(self.clone()), info);
+        super::update_info(&self.inner, &Participant::Local(self.clone()), info.clone());
+
+        for track in info.tracks {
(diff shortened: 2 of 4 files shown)


sandr01d:
Running into the same issue. Any chance this can be moved forward?

xianshijing-lk:
Thanks @davidzhao for looping me in. Looking at the code, I think it is doing the right thing, the existing Rust code assumes the the local participant does not need the event as the mute happen starting from the UIs, thus they don't need an event to notify them back. 
With a moderator mute behavior, this assumption is not correct any more, instead, they will need an event to get notified so that they can update the UI correspondingly.

Think about it for two use cases, if users update the UI to mute / unmute, then will get the redundant events, but that will be fine, since apps should not nothing if they check the cached value is the same as the mute value in the event, then do nothing. 
If moderator mute / unmute the local participant, apps will find a mismatch values, then they will update the UI to notify the users. 

the PR looks good to me in general. That says, I am pretty new to the our Rust SDK, please let me know if I miss anything.  @ladvoc, do you have any concern of landing it ? 


BTW, thanks for the PR and sorry for the length review process.

xianshijing-lk (approved):

.nanpa/track-muted-event-for-local-publications.kdl
ladvoc, line 1: This repository no longer uses the _nanpa_ release manager, so this file should be removed before merging.

ladvoc (approved):
Looks good to me, see one comment.

PR #545, opened 2025-01-11
Make `publish_data` wait until the DataChannel's bufferedAmount becomes low.

typester:
This pull request modifies `publish_data` to wait until the `bufferedAmount` of a DataChannel falls below a specified threshold before sending data, to support DataStream functionality.

The main changes are as follows:

1. Expose the `bufferedAmount` property from libwebrtc's DataChannel.
2. Forward the `bufferedAmountChange` callback of the DataChannel to an `RtcEvent`, allowing the rtc session to retrieve the current bufferedAmount.
3. Expose the ability for users to set and get the `buffered_amount_low_threshold` via FFI. (Default: 0)
4. Update `publish_data` to monitor the values in 2 and 3 and wait to send data until the bufferedAmount falls below the specified threshold.

Currently, the logic to wait for the buffer to become low has only been implemented for reliable DataChannels. In fact, all the features added in this update are exclusive to reliable DataChannels. Implementing this for lossy DataChannels didn’t make sense to me, but it might be worth discussing further.


livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -247,8 +250,8 @@ impl RtcSession {
         // Forward events received inside the signaling thread to our rtc channel
         rtc_events::forward_pc_events(&mut publisher_pc, rtc_emitter.clone());
         rtc_events::forward_pc_events(&mut subscriber_pc, rtc_emitter.clone());
-        rtc_events::forward_dc_events(&mut lossy_dc, rtc_emitter.clone());
-        rtc_events::forward_dc_events(&mut reliable_dc, rtc_emitter);
+        rtc_events::forward_dc_events(&mut lossy_dc, DataPacketKind::Lossy, rtc_emitter.clone());
+        rtc_events::forward_dc_events(&mut reliable_dc, DataPacketKind::Reliable, rtc_emitter);
```
bcherry, line 272: was this a separate bug?

typester: No, I just added it as a new argument because the newly added Event handler needed DataPacketKind.

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -747,6 +765,21 @@ impl SessionInner {
                     }
                 }
             }
+            RtcEvent::DataChannelBufferedAmountChange { sent: _, amount, kind } => {
+                match kind {
+                    DataPacketKind::Lossy => {
+                        // Do nothing at this moment
```
lukasIO: I agree with your assessment that this issue isn't as pressing on the lossy channel, but one thing we'd want to avoid is that the data channel just errors out if the buffered_amount surpassing an internal maximum value. I think @theomonnom mentioned that right now it will raise an error in this case? 
For that reason I think it could also make sense to include the same logic for the lossy data channel

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -958,6 +991,9 @@ impl SessionInner {
         kind: DataPacketKind,
     ) -> Result<(), EngineError> {
         self.ensure_publisher_connected(kind).await?;
+        if kind == DataPacketKind::Reliable {
+            self.wait_buffer_low().await?;
+        }
```
lukasIO: Is it guaranteed that successive calls to `publish_data` are handled in order here? 
I'm imagining a scenario where a user doesn't await `publish_data` and just fires a bunch of them in very quick succession, and we'd want to make sure that the next `self.data_channel().send(...)` is still processed in the correct order.

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -966,6 +1002,16 @@ impl SessionInner {
             })
     }
 
+    async fn wait_buffer_low(&self) -> Result<(), EngineError> {
+        let amount = self.reliable_dc_buffered_amount.load(Ordering::Relaxed);
+        let threshold = self.reliable_dc_buffered_amount_low_threshold.load(Ordering::Relaxed);
+        if amount <= threshold {
+            return Ok(());
+        }
+        self.reliable_dc_buffered_amount_low_notify.notified().await;
```
lukasIO: I'm not very familiar with tokio notify, so I'll ask a couple of questions that might be a bit naive: 

- what if `wait_for_buffer_low` is called exactly when `amount` drops under `threshold`? Is there a potential for a race here, where the new call will succeed before previous calls to `wait_buffer_low` will be notified ?
- if there are multiple calls to `self.reliable_dc_buffered_amount_low_notify.notified().await` because the threshold is exceeded, how would subsequent callers get notified? the `buffered_amount_changed` has a `notify_once` so it will only notify one caller when the buffered amount changes. I'm thinking of a scenario where the buffer drops below the threshold in `buffere_amount_changed` and there would be enough headroom to forward all pending packets. We would then still wait for each packet to be sent first, because we don't receive a `buffered_amount_changed` event until another packet has been sent, right? I'm not sure if that is actually a problem but  I think at least we'd have to be sure that `buffered_amount_changed` is reliably fired for every packet

lukasIO:
> Expose the ability for users to set and get the buffered_amount_low_threshold via FFI. (Default: 0)

It's great that it's configurable, but we might want to use a different threshold default value. This discussion [here](https://github.com/w3c/webrtc-pc/issues/1979#issuecomment-626190981) might be helpful for choosing some sane default values?

livekit/src/rtc_engine/rtc_session.rs
typester: I think there are no issues with the usage of `notify`, but I found one part that isn't ideal. I'll explain it in the comment below.

livekit/src/rtc_engine/rtc_session.rs
typester: I'm not sure if this is the same with your concern, but I found one part of my code that isn't ideal. Following is the content of `wait_for_buffer_low`. The code checks the threshold and immediately returns if there's no issue, but there's a possibility that a notify waiter already exists when this code is executed. This isn't ideal, so let me fix it.

https://github.com/livekit/rust-sdks/blob/88873783962558f7b0cf98f103eb71c517ad1241/livekit/src/rtc_engine/rtc_session.rs#L1008-L1012

----

> I'm thinking of a scenario where the buffer drops below the threshold in buffere_amount_changed and there would be enough headroom to forward all pending packets. We would then still wait for each packet to be sent first, because we don't receive a buffered_amount_changed event until another packet has been sent, right? I'm not sure if that is actually but I think at least we'd have to be sure that buffered_amount_changed is reliably fired for every packet

Yes, with my current code, calling `publish_data` multiple times could result in each call waiting for a notify. However, since `DataChannel.send` sends at least one buffer amount change event every time, I don't think it will end up waiting indefinitely. That said, as you pointed out, I agree that more testing is needed in this point.

typester:
I have made the following changes:

1. Updated the FFI functions to set and get the buffered amount low threshold, extending support to lossy data channels in addition to the already supported reliable data channels.
2. Created a dedicated worker task for data channels to enable a more robust message-sending mechanism.

Regarding 2, I believe that lowering latency is not an option to support lossy data channels. However, in my previous implementation, when multiple data packets were sent in a short period, a slight delay occurred even if the buffer had sufficient capacity.

To eliminate this delay, I created a new worker dedicated to data channels, which now manages the buffer by itself.

typester:
For the default value for buffered amount low threshold, I am still not sure what value should be ideal. :thinking:

lukasIO:
For a default value, maybe just half of the max value? So `8192` ?

livekit/src/rtc_engine/rtc_session.rs
theomonnom: Yes, I think we should use the same logic to avoid internal RtcError

theomonnom:
> Reference

Isn't the max value 65536?

livekit-ffi/protocol/ffi.proto
```diff
@@ -116,6 +116,9 @@ message FfiRequest {
     SendStreamHeaderRequest send_stream_header = 44;
     SendStreamChunkRequest send_stream_chunk = 45;
 
+    // Data Channel
+    GetDataChannelBufferedAmountLowThresholdRequest get_data_channel_buffered_amount_low_threshold = 46;
+    SetDataChannelBufferedAmountLowThresholdRequest set_data_channel_buffered_amount_low_threshold = 47;
```
theomonnom: I don't think we should "request" the buffered amount low threshold. Since datachannels are part of the roomm, let's add it to our RoomInfo?

The issue if we do a request, then on the python side, getting this threshold is going to be an async operation but this isn't ideal.

Ideally it is just:
```python
def buffered_amount_low_threshold(self) -> int:
	return self._info.buffered_amount_low_threshold
```

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -465,6 +511,103 @@ impl SessionInner {
...
+                .data_channel(SignalTarget::Publisher, kind)
+                .unwrap()
+                .send(&data, true)
+                .map_err(|err| {
+                    EngineError::Internal(format!("failed to send data packet: {:?}", err).into())
+                });
+
+            if let Err(_) = tx.send(result) {
+                log::error!("failed to send publish_data result");
+            }
```
theomonnom: I think we can ignore this failure? Not a big deal if the user cancel their call to `publish_data`

theomonnom (approved):
Otherwise lgtm! N

lukasIO:
> > For a default value, maybe just half of the max value? So 8192 ?
>
> Isn't the max value 65536?

that's the theoretic max value for a data packet, I think? 
But I was also missing a couple of zeros in my value, meant kB. 

Apparently the maximum bufferedAmount is 16MB in Chrome, so my suggestion was to use half of that as the threshold value, but we might also just do something a lot smaller like 1 or 2MB to be safe.

livekit-ffi/protocol/ffi.proto
typester: That super makes sense to me. Let me fix it

livekit/src/rtc_engine/rtc_session.rs
typester: Fixed in 0aad6d15ab8fd72e13f97d3751be36f7c6361b5b

lukasIO (approved):

typester:
Updated:

- Removed Set function for buffered_amount_low_threshold from FFI, instead, added it to RoomInfo.
- Added BufferedAmountLowThreasholdUpdated event so that FFI client can update room info

livekit-ffi/protocol/room.proto
```diff
@@ -369,13 +369,20 @@ message RoomEvent {
...
+    DataChannelBufferedAmountLowThresholdChanged data_channel_low_threshold_changed = 32;
   }
 }
 
 message RoomInfo {
   optional string sid = 1;
   required string name = 2;
   required string metadata = 3;
+  required DataChannelOptions lossy_dc_options = 4;
+  required DataChannelOptions reliable_dc_options = 5;
```
theomonnom: nit since we have one field I would have flatten the dc options

theomonnom (approved):
lgtm!

livekit-ffi/protocol/room.proto
typester: Will fix. Thank you!

typester: Fixed in 6c73254289f9131cc8b931f2f0cdb87f4e2f9bec.

PR #549, opened 2025-01-17
Add data_stream trailer event

lukasIO:


typester (approved):
Oh, I've missed this! Thanks!

PR #552, opened 2025-01-19
Set is_screencast on NativeVideoSource creation

iparaskev:
Makes InternalSource's is_screencast configurable when creating a NativeVideoSource. This improves the stream's latency when vp9 is used for sharing screen content.


typester (commented):
It looks like we already have a screenshare flag:

https://github.com/iparaskev/rust-sdks/blob/d276130dbd0205520f272861da718ce84b4bef80/livekit/src/room/options.rs#L76-L88

https://github.com/iparaskev/rust-sdks/blob/d276130dbd0205520f272861da718ce84b4bef80/livekit/src/room/track/mod.rs#L61-L68

Would it be sufficient to reuse this flag?

iparaskev:
> Would it be sufficient to reuse this flag?

I didn't think of this TBH. It could be done by exposing a setter in `VideoTrackSource` and then set the `is_screencast_` member in `publish_track`. If you are OK with this, I am happy to go with it.

Side note: The only "problem" I see with this, is that `NativeVideoSource` will require a kind of two stage initialization, which could be error prone. A way to mitigate it would be to create the `NativeVideoSource` object in `publish_track`, but I am not sure if this is acceptable.

iparaskev:
Something that I had forgotten to mention when I created this, is that when `is_screencast` is set, libwebrtc limits the FPS to 5. I had to add `spatial_layers.back().maxFramerate = video_codec.maxFramerate;` [here](https://github.com/webrtc-sdk/webrtc/blob/m125_release/modules/video_coding/video_codec_initializer.cc#L251) as a workaround.

iparaskev:
I tested it with m137 and the `libwebrtc` WA is not needed any more, [f126e8af824c45ba2343225229f051d3b3418923](https://source.chromium.org/chromium/_/webrtc/src/+/f126e8af824c45ba2343225229f051d3b3418923) fixed it. Setting the screensharing flag doesn't affect the FPS now.

PR #557, opened 2025-01-23
Add nanpa changeset for #555

lukasIO:


nbsp (approved):
for further reference, it's possible to have this file at the root `.nanpa`, and simply specify `package="livekit-ffi"` in the line. this should work though (if it doesn't i have a bug to fix :3)

lukasIO:
ah, nice, I just went with the man page said which called for "package root"

PR #559, opened 2025-01-27
feat: Audio Filter plugin

typester:
This pull request introduces a feature that allows users to load their own audio filter library and use it as an inbound audio filter. It can be used for purposes such as noise cancellation, VAD, and more.

The plugins must be dynamic libraries (`cdylib`) and are required to export the following functions:

```rust
type CreateFn = unsafe extern "C" fn(sampling_rate: u32, options: *const c_char) -> *mut c_void;
type DestroyFn = unsafe extern "C" fn(*const c_void);
type ProcessI16Fn = unsafe extern "C" fn(*const c_void, usize, *const i16, *mut i16);
type ProcessF32Fn = unsafe extern "C" fn(*const c_void, usize, *const f32, *mut f32);
```


bcherry (approved):
lgtm but i wouldn't give my review too much weight, needs @theomonnom

theomonnom (approved):

livekit/src/plugin.rs
```diff
@@ -0,0 +1,200 @@
...
+        Ok(Arc::new(this))
+    }
+
+    fn _new<P: AsRef<str>>(path: P, options: P) -> Result<Self, PluginError> {
+        let lib = unsafe { Library::new(path.as_ref()) }?;
+
+        let options = CString::new(options.as_ref()).unwrap_or(CString::new("").unwrap());
+        unsafe {
+            let on_load = lib.get::<Symbol<OnLoadFn>>(b"on_load")?;
+            on_load(options.as_ptr());
```
theomonnom: I see the on_load function do a blocking request to our livekit server (on the krisp-plugin repo). 
I think we should add 3 things:
- making sure we don't block the tokio loop when loaded.
- add retries to this request (or expose failures?)
- check how slow it is to load the weights into memory

livekit/src/plugin.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    pub fn new_session<S: AsRef<str>>(
+        self: Arc<Self>,
+        sampling_rate: u32,
+        options: S,
+    ) -> AudioFilterSession {
+        let create_fn: CreateFn = unsafe { std::mem::transmute(self.create_fn_ptr) };
+
+        let options = CString::new(options.as_ref()).unwrap_or(CString::new("").unwrap());
+        let ptr = unsafe { create_fn(sampling_rate, options.as_ptr()) };
```
theomonnom: should we check if the ptr is valid?

livekit/src/plugin.rs
typester: - Moved this initialization to `Room::connect`, now connect returns error if on_load fails.
- Put plugin loading codes in spawn_blocking so that we can avoid tokio runtime are blocked.

livekit/src/plugin.rs
typester: Just added for null check for this in https://github.com/livekit/rust-sdks/pull/559/commits/1bdb93281d435dd9bc98535d57f1de2eb8b93a6b

PR #560, opened 2025-01-28
Update protocol and add SendDataRequest nonce

lukasIO:


theomonnom (approved):

livekit-api/src/services/room.rs
```diff
@@ -288,6 +289,8 @@ impl RoomClient {
         data: Vec<u8>,
         options: SendDataOptions,
     ) -> ServiceResult<()> {
+        let mut rng = rand::rng();
+        let nonce: Vec<u8> = (0..16).map(|_| rng.random::<u8>()).collect();
```
theomonnom, line 293: nit, you could easily avoid heap allocation
`let mut nonce = [0u8; 16]; `

lukasIO: had this initially with `rnd:fill(arr)` but this required `.to_vec()` which copies everything into the vec. I guess it would still do heap allocation at the time of copying it into the vec?

theomonnom: ah yeah, seems unnecessary then

PR #578, opened 2025-02-19
update builder runner images

matkam:
updating images ahead of deprecation. updating all images, not just `ubuntu-20.04`.
https://github.com/actions/runner-images/issues/11101


theomonnom (approved):
lg

PR #580, opened 2025-02-20
add AudioProcessingModule

theomonnom:
https://github.com/webrtc-sdk/webrtc/blob/844bafa06d0b9088fd7fa4244832abf8e70a1d3d/modules/audio_processing/aec3/echo_canceller3.h#L120 isn't exposed, as it isn't needed for our local playout (no delay). This may need to be exposed for SDK like Unity


typester (approved):
LGTM.

This is exactly what I investigated earlier this month. Thank you for doing this.

PR #581, opened 2025-02-24
fix: pass disconnect reason explicitly

davidzhao:
I had thought the entire participant info was updated when changes were made.

confirmed this is the correct way to pass through disconnect_reason


lukasIO (approved):
lgtm

PR #583, opened 2025-02-24
fix yuv path 

theomonnom:


davidzhao:
doesn't seem necessary.. closing

PR #585, opened 2025-03-03
 Fix panic in Promise::try_result

typester:
Panic occurring in `Promise::try_result` is undesirable, so it has been replaced with RwLock to prevent panics.

### DIFF
--- a/livekit-protocol/src/promise.rs
+++ b/livekit-protocol/src/promise.rs
@@ -55,6 +55,6 @@ impl<T: Clone> Promise<T> {
     }
 
     pub fn try_result(&self) -> Option<T> {
-        self.result.try_read().unwrap().clone()
+        self.result.try_read().ok().and_then(|result| result.clone())
     }
 }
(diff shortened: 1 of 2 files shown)


livekit-protocol/src/promise.rs
```diff
@@ -37,14 +37,24 @@ impl<T: Clone> Promise<T> {
...
+            result
+        } else {
+            self.result.read().await.clone().unwrap()
         }
-        self.result.lock().await.clone().unwrap()
     }
 
     pub fn try_result(&self) -> Option<T> {
-        self.result.try_lock().unwrap().clone()
+        self.result.try_read().unwrap().clone()
```
theomonnom: isn't try_read also panicking? when unwrapping?

theomonnom (approved):

livekit-protocol/src/promise.rs
typester: Oh, good catch! Thank you. I believe a panic would occur if a write lock is held.

So in this case, I'll modify it to return None instead of unwrapping. Since Promise is designed to acquire the write lock only once, this should not be an issue.

bcherry (approved):

PR #589, opened 2025-03-04
Fix packaging issue in livekit-ffi

typester:
Definitions of dependencies without specifying versions cause [a packaging issue](https://github.com/livekit/rust-sdks/actions/runs/13659444398/job/38186921358) in livekit-ffi. Using a workspace definition resolves this.

We don't use a published crate for FFI, so this doesn't block any releases of foreign language SDKs, but I will fix this just in case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1655,7 +1655,7 @@ dependencies = [
 
 [[package]]
 name = "livekit-ffi"
-version = "0.12.12"
+version = "0.12.13"
 dependencies = [
  "console-subscriber",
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
lg.. though I haven't seen this before

typester:
> lg.. though I haven't seen this before

I feel the same way. I'm wondering why there was no issue before.
Anyway I've found a related issue: https://github.com/rust-lang/cargo/issues/11133
It looks like it might be fixed in the future.

PR #598, opened 2025-03-10
bump manylinux to manylinux_2_34

theomonnom:


theomonnom:
For reference, it seems like Debian 12 glibc is 2.36 by default.
https://packages.debian.org/stable/source/glibc

Latest Raspberry PI OS (Bookworm) runs Debian 12

typester (approved):
LGTM on the change itself — thanks :+1: 

Just wondering: are we okay with dropping compatibility with older systems that don't support glibc 2.34?

Personally, I’m not sure we really need to bump the version yet — is there any specific benefit you have in mind?

davidzhao:
does this bring about any benefits? as a SDK, it'd be ideal if we are not raising the requirement for OS versions too aggressively.

folks in server environments can be slow to update.

PR #599, opened 2025-03-11
Move RPC handlers to room

typester:


livekit/src/room/participant/local_participant.rs
```diff
@@ -690,198 +681,7 @@ impl LocalParticipant {
...
-                Err(e)
-            }
-            Ok(Ok(payload)) => {
-                // Successful response
-                Ok(payload)
-            }
-        }
-    }
-
-    pub fn register_rpc_method(
```
bcherry, line 775: can we keep deprecated stubs of these around that just forward to the new room method? that way its not a breaking change

bcherry (approved):
lgtm please add a changeset and add deprecated methods if not too much work

livekit/src/room/participant/local_participant.rs
typester: @bcherry 
Oh, I was thinking of covering this on the foreign language side.
I didn’t want to support the old interface from rust side, since that would mean we’d have to keep the old proto definitions around and start adding things like RegisterRpcMethodRequest2, which I’d rather avoid. What do you think?

bcherry: i don't think you need to expose the old interface over FFI (agree you can just handle it in python/node), but the underlying rust rtc sdk would ideally still offer it. i don't think there'd be extra proto definitions around in that case

typester: I see — you mean keeping the old interface on the livekit crate, not in livekit-ffi. That makes sense to me. Will do!

typester: Added in [cf40cb9](https://github.com/livekit/rust-sdks/pull/599/commits/cf40cb9228d0895d17ec82dd08b72ee9aee7535e)

PR #603, opened 2025-03-13
Data streams

ladvoc:
This PR adds native data streams support in Rust and a high-level interface for FFI clients.


bcherry (commented):
lgtm from a protocol level.

theomonnom:
I'm wondering how we could still allow to register RPC handlers before connecting to the room?

livekit/Cargo.toml
```diff
@@ -43,3 +43,5 @@ log = "0.4"
 chrono = "0.4.38"
 semver = "1.0"
 libloading = { version = "0.8.6" }
+bytes = "1.10.1"
+bmrng = "0.5.2"
```
typester, line 47: I wasn't familiar with the `bmrng` crate — it looks quite handy!
That said, it seems like the usage here is fairly straightforward, and I wonder if it's worth introducing a new dependency just for this.
Wouldn't it be possible to achieve the same behavior with a simple combination of `mpsc::unbounded_channel` and `oneshot::channel` instead?

livekit/src/room/mod.rs
```diff
@@ -469,6 +476,19 @@ impl Room {
...
+        let identity = local_participant.identity().clone();
+        tokio::task::spawn(async move {
+            // Receive packets from the outgoing stream manager and send them.
+            while let Ok((mut packet, responder)) = packet_rx.recv().await {
+                // Set packet's participant identity field
+                packet.participant_identity = identity.0.clone();
+                let result = engine.publish_data(packet, DataPacketKind::Reliable).await;
+                let _ = responder.respond(result);
+            }
+        });
```
typester: This spawn block could potentially be extracted into a dedicated function like `fn data_stream_task` for better readability and consistency.

Additionally, it might be beneficial to integrate a close_tx/close_rx mechanism to ensure the task can be cleanly shut down when the room is closed.

Currently, RoomSession only tracks a single task using close_tx, but if you look at the room implementation in the livekit-ffi side, you’ll see that it provides a more complete example of how to manage multiple room-related tasks properly.

While this task would eventually terminate on its own, we generally prefer to ensure all tasks are properly joined or terminated by the time Room::close() returns.
This helps prevent unexpected behavior or dangling background tasks.

typester (commented):
I just realized this PR depends on the changes I reverted yesterday.
@bcherry, how should we proceed?

@ladvoc 
Overall, I think this is a great PR. I left a few minor comments to align with the existing code style.

livekit/Cargo.toml
ladvoc: Yes, the same behavior can definitely be achieved by combining `mpsc::unbounded_channel` and `oneshot::channel`—that was my original approach. I introduced `bmrng` for readability and boilerplate reduction. However, I also agree that dependencies should be introduced sparingly. If you don’t see `bmrng` as being useful in other scenarios, I think I’ll go with the channels approach.

livekit/src/room/mod.rs
ladvoc: This is all good feedback; I will make the necessary changes.

ladvoc:
@theomonnom, I missed your previous comment. You mentioned RPC handlers, but I see how the same issue applies to data stream handlers as well. Using data streams as an example, from my current understanding, a race condition can occur when a participant opens a data stream in response to another participant joining, but the newly joined participant hasn't registered a handler for that stream's topic yet:

```mermaid
sequenceDiagram
    participant p1 as Participant 1
    participant r as Room
    participant p2 as Participant 2

    p1 ->> r: Join room
    p2 ->> r: Join room
    r ->> p1: Participant 2 Joined
    p1 ->> r: Open stream for "some-topic"
    r ->> r: No handler registered for "some-topic", stream dropped
    p2 ->> r: Register text stream handler for "some-topic"
```
Taking into account support for FFI and other languages, I propose allowing preregistration of stream topics (not handlers) in `RoomOptions` before connect. In Rust, this would look something like this:
```rust
let options = RoomOptions {
    stream_preregistration: StreamPreRegistration {
        text_topics: vec!["some-topic".to_string()],
        byte_topics: vec!["another-topic".to_string()]
    },
    // Other room options…
};
```
When a topic is preregistered, any incoming stream matching that topic will be queued until the corresponding handler is registered, at which point each queued stream is dequeued and passed to the handler:

```mermaid
sequenceDiagram
    participant p1 as Participant 1
    participant r as Room
    participant p2 as Participant 2

    p1 ->> r: Join room
    p2 ->> r: Join room, preregister "some-topic"
    r ->> p1: Participant 2 Joined
    p1 ->> r: Open stream for "some-topic"
    r ->> r: Queue stream reader
    p2 ->> r: Register text stream handler for "some-topic"
    r ->> p2: Dequeue stream reader, invoke handler for "some-topic"
```

A key consideration is that when a topic is preregistered, the user must register the corresponding handler within a short window after connecting. Otherwise, incoming streams could accumulate indefinitely without being handled. To prevent this, I propose a timeout interval (e.g., 30 seconds after connect), after which any queued stream readers for unregistered handlers will be dropped, and an error will be logged. In practice, as long as the handler is registered shortly after connect, the number of queued streams should remain minimal.

This approach could also be extended to RPC methods by allowing preregistration of method names in `RoomOptions` before connect and queuing RPC calls until the handler is registered afterward. I’d appreciate any feedback on this proposal, particularly on whether it fully addresses the issue.

theomonnom:
> When a topic is preregistered, any incoming stream matching that topic will be queued until the corresponding handler is registered, at which point each queued stream is dequeued and passed to the handler:
Does this mean that every client will have to return all registered handlers after login? It would be easier if we don't have to.


> A key consideration is that when a topic is preregistered, the user must register the corresponding handler within a short window after connecting. Otherwise, incoming streams could accumulate indefinitely without being handled. To prevent this, I propose a timeout interval (e.g., 30 seconds after connect), after which any queued stream readers that don't yet have handlers will be dropped, and an error will be logged. In practice, as long as the handler is registered shortly after connect, the number of queued streams should remain minimal.
If you send directly after connecting (assume the client language already has a registered handler), you could skip this part

lukasIO:
> Another solution would be to not register any callback to the room, because it seems like an anti-pattern with our channel based events? In this case everything will be buffered and you "support the register before connect" on all languages for free

@theomonnom do I understand your suggestion correctly that we could

- buffer all incoming RPC and StreamHeaders (with subsequent chunks) until connected
- when connected, have a short timeout (even 1s should be enough) that allows for patterns like await room.connect(); registerHandlers();
- if there are handlers registered within that timeout, pipe all buffered packets to the handlers
- after the timeout elapsed (and no handlers have been registered) discard all buffered data

theomonnom:
I think we don't even need to drop them? If we push the data inside the room_events channel. They're buffered but will be consumed as soon as the connection is established. (Like basically a new RoomEvent?)

The issue with the above pattern, is that every language will have to send a registerHandler each time after the connecting.

But if we "always" forward what we receive to the other languages, then no registration is needed.

theomonnom (approved):
lgtm! very clean impl!

PR #605, opened 2025-03-18
Add i420_to_nv12

ConradIrwin:
Currently our graphics framework at Zed requires NV12 buffers, but using the high-quality VP8 codec, we receive I420 buffers from livekit.

It'd be nice to figure out how to avoid that (and have libwebrtc do whatever conversions are needed internally), but in the meantime we need to call this on the received data.


theomonnom (approved):
Thanks

PR #607, opened 2025-03-20
Follow-up for #605

typester:
Looks like the changeset included in #605 was incorrect (missing package name), and it wasn’t detected by nanpabot. Here's the fix.


typester:
This pr isn't valid for now, as nanpa has been disabled. just close this

PR #619, opened 2025-04-04
fix rtc.AudioFrame.from_participant not closing

theomonnom:
This code really need cleanup..


davidzhao (approved):
nice! approving to unblock.

@typester should still review at your own pace

PR #629, opened 2025-04-30
Feat/metrics

anunaym14:


livekit-ffi/protocol/metrics.proto
theomonnom, line 1: Let's not copy this file, it is going to be very hard to keep it up to date since it's a copy (we will also have to release a new FFI versions each time we add a metric). 

Instead, let's build the livekit metrics protocol on the Python side and serialize it before sending it to the FFI. The FFI is just going to be a passthrough for sending it over the DC.


You can take this example on how you can send raw bytes from Python: https://github.com/livekit/rust-sdks/blob/0773bcec4e24812b382e71166b2ab359bf1a9384/livekit-ffi/protocol/room.proto#L100

livekit-ffi/protocol/ffi.proto
```diff
@@ -301,6 +306,18 @@ message FfiEvent {
   }
 }
 
+message PublishMetricsRequest {
+  required uint64 local_participant_handle = 1;
+  repeated string str_data = 2;
+  repeated TimeSeriesMetric time_series = 3;
+  repeated EventMetric events = 4;
+  optional uint64 async_id = 5;
```
theomonnom: This will then become
```suggestion
  required uint64 local_participant_handle = 1;
  required uint64 data_ptr = 2;
  required uint64 data_len = 3;
  optional uint64 async_id = 4;
```

livekit/src/room/metrics/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+                value: Some(data_packet::Value::Metrics(MetricsBatch {
+                    str_data: strings,
+                    time_series: metrics,
+                    timestamp_ms: timestamp,
+                    ..Default::default()
+                })),
+                ..Default::default()
+            };
+
+            let _ = rtc_engine.publish_data(data_packet, DataPacketKind::Reliable).await;
```
theomonnom: Don't ignore errors, this function should return a Result

livekit/src/room/metrics/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    pub(crate) async fn collect_metrics(
+        &self,
+        room: &Room,
+        rtc_engine: &RtcEngine,
+        cancellation_token: CancellationToken,
+    ) {
+        cancellation_token
+            .run_until_cancelled(async move {
+                std::thread::sleep(Duration::from_secs(1));
```
theomonnom: This is blocking the event loop, use a tokio sleep instead or an async interval instead

livekit/src/room/mod.rs
```diff
@@ -372,12 +375,15 @@ impl Default for RoomOptions {
...
             preregistration: None,
+            enable_metrics: true
         }
     }
 }
 
+#[derive(Clone)]
 pub struct Room {
     inner: Arc<RoomSession>,
+    cancellation_token: tokio_util::sync::CancellationToken
```
theomonnom: I think we can avoid the cancellation token. This seems unnecessary.

We can just keep track of the handle returned by tokio::spawn and await it on close.

To "close/cancel" the task, you can use a tokio::select pattern

livekit/src/room/metrics/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+        let sample = self.create_metric_sample(timestamp_ms, stat_value);
+
+        self.create_time_series(label, strings, vec![sample], identity, Some(&track_sid), rid)
+    }
+
+    pub(crate) async fn collect_metrics(
+        &self,
+        room: &Room,
+        rtc_engine: &RtcEngine,
+        cancellation_token: CancellationToken,
```
theomonnom: The cancellation can be done beforehand directly in the task inside Room using  tokio::select!

This function should be cancel safe: 
https://docs.rs/tokio/latest/tokio/macro.select.html#cancellation-safety

livekit/src/room/metrics/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+        stats: &Vec<RtcStats>,
+    ) {
+        let mut strings: Vec<String> = Vec::new();
+        let subscriber_ts_metrics = self.find_subscriber_stats(
+            &mut strings,
+            stats,
+            room.local_participant().identity().into(),
+        );
+
+        let _ = self.send_metrics(rtc_engine, strings, subscriber_ts_metrics).await;
```
theomonnom: Same here, let's return a Result

livekit/src/room/metrics/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    fn create_metric_sample(&self, timestamp_ms: i64, value: f32) -> MetricSample {
+        MetricSample { timestamp_ms, value, ..Default::default() }
+    }
+
+    fn create_time_series(
+        &self,
+        label: MetricLabel,
+        strings: &mut Vec<String>,
```
theomonnom, line 67: Any reason why this need to be a mutable reference? It seems like the "caller" doesn't expect it? So maybe it's fine to clone inside this method and keep the modifications in this scope

livekit/src/room/metrics/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+    ) {
+        let mut strings = Vec::new();
+        let publisher_ts_metrics = self.find_publisher_stats(
+            room,
+            &mut strings,
+            stats,
+            room.local_participant().identity().into(),
+        );
+
+        let _ = self.send_metrics(rtc_engine, strings, publisher_ts_metrics).await;
```
theomonnom: Same here, let's return a Result

livekit-ffi/generate_proto.sh
```diff
@@ -20,6 +20,8 @@ OUT_RUST=src
 protoc \
     -I=$PROTOCOL \
     --prost_out=$OUT_RUST \
+    --prost_opt=compile_well_known_types \
+    --prost_opt=extern_path=.google.protobuf=::pbjson_types \
```
theomonnom: Do we need json types? Seems like we never serialize to json?

.github/workflows/builds.yml
```diff
@@ -88,6 +88,8 @@ jobs:
       - name: Build (Android)
         if: ${{ contains(matrix.target, 'android') }}
         run: |
+          sudo apt install -y gcc-multilib
+
```
theomonnom, line 92:
```suggestion
```

The FFI builds are already passing, the issues is just the "default" CI builds, feel free to ignore

livekit-ffi/protocol/metrics.proto
theomonnom: On Python, it can simply be generated for https://github.com/livekit/python-sdks/tree/main/livekit-protocol
On Node, we can reuse: https://github.com/livekit/protocol/tree/main/packages/javascript

anunaym14: Updated, please check.

livekit/src/room/metrics/mod.rs
anunaym14: Handled, please check again

livekit/src/room/metrics/mod.rs
anunaym14: Fixed

livekit/src/room/mod.rs
anunaym14: Made it better, removed cancellation token, please have another look.

livekit-ffi/generate_proto.sh
anunaym14: Not anymore, livekit-protocol was converting .google.protobuf types to pbjson_types but livekit-ffi was not doing so, it was messing up timestamps in metrics.proto. Since we have removed metrics proto from ffi, we don't need this anymore.

livekit/src/room/metrics/mod.rs
anunaym14: Yes actually:

- Strings are modified here: https://github.com/livekit/rust-sdks/blob/fd33fe818a4e57fa69a6297e0bd96866721d6e88/livekit/src/room/metrics/mod.rs#L54
- but are created here: https://github.com/livekit/rust-sdks/blob/fd33fe818a4e57fa69a6297e0bd96866721d6e88/livekit/src/room/metrics/mod.rs#L230

livekit-ffi/src/server/room.rs
```diff
@@ -839,6 +839,35 @@ impl RoomInner {
...
+        let data = unsafe {
+            slice::from_raw_parts(publish.data_ptr as *const u8, publish.data_len as usize)
+        }
+        .to_vec();
+
+        let self_clone = self.clone();
+        let handle = server.async_runtime.spawn(async move {
+            if let Err(err) = self_clone.data_tx.send(FfiDataPacket {
+                payload: DataPacket {
+                    payload: data.to_vec(), // Avoid copy?
```
xianshijing-lk, line 859: assuming this PR is still active, can we address the comment to avoid the copy here ?

livekit-ffi/src/server/room.rs
```diff
@@ -839,6 +839,35 @@ impl RoomInner {
...
+        self: &Arc<Self>,
+        server: &'static FfiServer,
+        publish: proto::PublishMetricsRequest,
+    ) -> FfiResult<proto::PublishMetricsResponse> {
+        let async_id = publish.async_id.unwrap_or_else(|| server.next_id());
+
+        let data = unsafe {
+            slice::from_raw_parts(publish.data_ptr as *const u8, publish.data_len as usize)
+        }
+        .to_vec();
```
xianshijing-lk, line 853: does it copy the data here ?

PR #631, opened 2025-05-02
feat: Region pinning support

typester:


livekit/src/rtc_engine/mod.rs
```diff
@@ -337,13 +337,15 @@ impl EngineInner {
         let lk_runtime = LkRuntime::instance();
         let max_retries = options.join_retries;
 
+        let urls = RegionUrlProvider::fetch_region_urls(url, token).await?;
```
davidzhao: so I think we probably should connect directly first, instead of first fetching the URLs.. because this would slow down the initial connection (one more RTT)

in many cases, the default URL is the right choice. so we should only do this when it fails connection

bcherry: what is the precise behavior if region pinning is turned on for the project but you just connect to the default URL? does it route correctly under the hood anyways, or does it route to nearest but fail if that's a disallowed region?

davidzhao: it'll disconnect and return a `Forbidden` response ([source](https://github.com/livekit/cloud/blob/main/pkg/node/middleware.go#L100))

what clients need to do is:
* when the initial websocket connection fails
  * call validate to ensure the token is valid
    * abort if token is invalid
  * call /settings/regions
  * retry the same connection sequence with other nearby regions

typester: Ok, I will update the code.

theomonnom: Should we use the RegionUrlProvider directly inside the SignalClient instead? in the livekit-api crate

theomonnom: Some people (e.g gstreamer) use the SignalClient but doesn't use our `livekit` crate. This may be a better place?

livekit/src/rtc_engine/mod.rs
```diff
@@ -379,20 +381,22 @@ impl EngineInner {
...
-                            max_retries
-                        );
+        for url in urls.iter() {
+            for i in 0..(max_retries + 1) {
+                match try_connect(url.clone()).await {
+                    Ok(res) => return Ok(res),
+                    Err(e) => {
+                        let attempt_i = i + 1;
+                        if i < max_retries {
+                            log::warn!(
```
lukasIO: It would be nice if we had more insight here in the logs. 
Right now the attempts will just loop for every region, so the attempts and max retries will look odd. 
Maybe we can add the "url" parameter to the log to indicate which region the connection failed to

typester: Oh yeah, I thought the same thing but I forgot to update the code somehow :sweat_smile: 
Will do

livekit/src/rtc_engine/mod.rs
typester: That makes sense. Will do it

typester:
I made the following changes:

- Moved the connection fallback to regional URLs functionality to the SignalClient.
- Removed the initial fetching of regional URLs. If the default connection is refused, try connecting to the regional URLs.
- Log an error if the default URL returns an error other than 403.

bcherry (approved):

livekit-api/src/signal_client/region.rs
```diff
@@ -0,0 +1,70 @@
...
+                .send()
+                .await
+                .map_err(|e| SignalError::RegionError(e.to_string()))?;
+            let res = res
+                .json::<RegionUrlResponse>()
+                .await
+                .map_err(|e| SignalError::RegionError(e.to_string()))?;
+            Ok(res.regions.into_iter().map(|i| i.url).collect())
+        } else {
+            Ok(vec![url.into()])
```
davidzhao: IMO this should be returning empty.. no regional urls are fetched

lukasIO (approved):
lgtm, the only nit I have is that now for region pinned projects users that don't connect correctly via geo always get the penalty of a failed connection attempt first. 

Ideally we'd offer a `prepare_connection` method on top that sets the region url for the next connect attempt. see https://github.com/livekit/client-sdk-js/blob/main/src/room/Room.ts#L656-L665

livekit-api/src/signal_client/region.rs
typester: Oh, yes. Will fix

typester: Fixed in a4a5adc4c2f799fb96d1d55caa1c503a5cf00d86

typester:
> lgtm, the only nit I have is that now for region pinned projects users that don't connect correctly via geo always get the penalty of a failed connection attempt first.
> 
> Ideally we'd offer a `prepare_connection` method on top that sets the region url for the next connect attempt. see https://github.com/livekit/client-sdk-js/blob/main/src/room/Room.ts#L656-L665

Yeah, we'll need this at some point in the future, but I think it's fine for this PR in terms of minimal implementation.
I'll keep it in mind. Thank you!

PR #634, opened 2025-05-05
Bump protocol to v1.37.1 and add audio_features handling

lukasIO:
Main motivation is to get the new pre connect buffer AudioTrackFeature added for usage in https://github.com/livekit/agents/pull/2171

### DIFF
--- a/livekit-api/src/services/egress.rs
+++ b/livekit-api/src/services/egress.rs
@@ -153,8 +153,7 @@ impl EgressClient {
                     image_outputs,
                     output: None, // Deprecated
                     await_start_signal: options.await_start_signal,
-                    // TODO implement webhooks
-                    webhooks: Default::default(),
+                    ..Default::default()
                 },
                 self.base
                     .auth_header(VideoGrants { room_record: true, ..Default::default() }, None)?,
@@ -186,8 +185,7 @@ impl EgressClient {
                     stream_outputs,
                     segment_outputs,
                     image_outputs,
-                    // TODO implement webhooks
-                    webhooks: Default::default(),
+                    ..Default::default()
                 },
                 self.base
                     .auth_header(VideoGrants { room_record: true, ..Default::default() }, None)?,
@@ -219,8 +217,7 @@ impl EgressClient {
                     segment_outputs,
                     image_outputs,
                     output: None, // Deprecated
-                    // TODO implement webhooks
-                    webhooks: Default::default(),
+                    ..Default::default()
                 },
                 self.base
                     .auth_header(VideoGrants { room_record: true, ..Default::default() }, None)?,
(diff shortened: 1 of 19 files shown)


livekit-ffi/src/livekit.proto.rs
```diff
@@ -1694,6 +1694,7 @@ pub enum DisconnectReason {
     UserRejected = 12,
     /// SIP protocol failure or unexpected response
     SipTrunkFailure = 13,
+    ConnectionTimeout = 14,
```
lukasIO, line 1741: Can we simply add this without breaking users' code?

livekit/src/room/data_stream/outgoing.rs
```diff
@@ -74,7 +74,7 @@ impl<'a> StreamWriter<'a> for ByteStreamWriter {
         &self.info
     }
 
-    async fn write(&self, bytes: &[u8]) -> StreamResult<()> {
+    async fn write(&self, bytes: &'a [u8]) -> StreamResult<()> {
```
lukasIO, line 77: the lifetime annotation here was needed to make it compile locally on my setup (v1.83) - not sure why this wasn't failing on CI, any drawbacks around adding this?

livekit-api/src/services/egress.rs
```diff
@@ -215,6 +219,8 @@ impl EgressClient {
                     segment_outputs,
                     image_outputs,
                     output: None, // Deprecated
+                    // TODO implement webhooks
+                    webhooks: Default::default(),
```
theomonnom: The reason why we have to do this everytime is because we're using the wrong syntax
```suggestion
                    ..Default::default(),
```

theomonnom (approved):

lukasIO:
Ah, I assumed trackInfo just gets passes as is, but it doesn't. Will add handling for audio track features

lukasIO:
@theomonnom I added handling for audio_features both on the rust side and FFI

livekit-ffi/src/conversion/room.rs
```diff
@@ -98,6 +98,7 @@ impl From<DisconnectReason> for proto::DisconnectReason {
             DisconnectReason::UserUnavailable => Self::UserUnavailable,
             DisconnectReason::UserRejected => Self::UserRejected,
             DisconnectReason::SipTrunkFailure => Self::SipTrunkFailure,
+            DisconnectReason::ConnectionTimeout => Self::ConnectionTimeout,
```
theomonnom, line 101: Can you create a similar conversion for the TrackFeature?
`impl From< AudioTrackFeature > for proto:: AudioTrackFeature {`

We're currently only relying on the integer, which is unsafe and will panic as soon as the proto adds a new one

lukasIO: @theomonnom I added conversions in both directions and added AudioTrackFeature to the rtc prelude

theomonnom (approved):

PR #635, opened 2025-05-08
bump ffi, rtc, protocol

lukasIO:


theomonnom (approved):
lg!
you can then tag `rust-sdks/livekit-ffi@0.12.23` and you will automatically get a ffi release

PR #638, opened 2025-05-13
feat: VA-API support for linux.

cloudwebrtc:
vaapi works now, but there are some things need to confirm.

- [x] can call drm driver for h264 encoding for AMD/APU
- [x] check rate control is effective
- [x] confirm encoder adapts to resolution changes
- [x] check support for Intel CPU
- [x] check and confirm for H.264 encoding in baseline, main & high profiles
- [ ] simulcast support (maybe in next commit)
- [x] add capability detection support for vaapi encoder factory


cloudwebrtc:
Tried to install libva win32 today, and vaon12 driver based on d3d12 on windows, but unfortunately it didn't work, the encoder can be opened but the bitstream cannot be encoded. So at this moment we can only support linux

iliabaranov:
Exciting! :+1:

iliabaranov:
For sure!

webrtc-sys/libwebrtc/build_linux.sh
```diff
@@ -108,9 +108,9 @@ args="is_debug=$debug  \
   use_rtti=true \
   rtc_use_x11=false"
 
-if [ "$debug" = "true" ]; then
-  args="${args} is_asan=true is_lsan=true";
-fi
+#if [ "$debug" = "true" ]; then
```
chenosaurus: is this commented out section still useful?

chenosaurus (approved):

webrtc-sys/libwebrtc/build_linux.sh
cloudwebrtc: I remember that these args need to be removed, they will cause the compiled libwebrtc.a to be unable to link to the binary in rust ffi,  will double check

PR #646, opened 2025-05-16
try to fix webrtc build for iOS/macOS.

cloudwebrtc:


ladvoc (approved):
Looks good to me, I'll make sure the build artifact looks good too once the current run finishes

cloudwebrtc:
It looks like all binaries are compiled
https://github.com/livekit/rust-sdks/actions/runs/15127246993

PR #654, opened 2025-06-05
fix duration overflow

theomonnom:
There is no point about reducing the timeout if we don't know the latency.. It's hardcoded to 2s. 

Fixes substraction overflow


theomonnom:
fix https://github.com/livekit/agents/issues/2517

PR #662, opened 2025-06-16
Expose room updates, support MoveParticipant (protocol 15)

davidzhao:
giving clients other information about the room so they could receive all updates

also ensure that we pass through full participant updates, since sid is expected to change in certain situations (reconnect, move)

FYI @bcherry

### DIFF
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -47,7 +47,7 @@ pub type SignalEvents = mpsc::UnboundedReceiver<SignalEvent>;
 pub type SignalResult<T> = Result<T, SignalError>;
 
 pub const JOIN_RESPONSE_TIMEOUT: Duration = Duration::from_secs(5);
-pub const PROTOCOL_VERSION: u32 = 15;
+pub const PROTOCOL_VERSION: u32 = 16;
 
 #[derive(Error, Debug)]
 pub enum SignalError {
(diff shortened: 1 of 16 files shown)


livekit-ffi/src/server/room.rs
```diff
@@ -652,7 +652,7 @@ impl RoomInner {
                     ));
                 }
             }
-            sent_message;
+            drop(sent_message);
```
theomonnom: This line is not even needed fwiw

theomonnom (approved):

livekit-api/src/signal_client/mod.rs
```diff
@@ -47,7 +47,7 @@ pub type SignalEvents = mpsc::UnboundedReceiver<SignalEvent>;
 pub type SignalResult<T> = Result<T, SignalError>;
 
 pub const JOIN_RESPONSE_TIMEOUT: Duration = Duration::from_secs(5);
-pub const PROTOCOL_VERSION: u32 = 15;
+pub const PROTOCOL_VERSION: u32 = 16;
```
typester, line 50: Is this `PROTOCOL_VERSION` defined in the cloud side as well? I just wanted to know how we manage this

typester (approved):

livekit-ffi/src/server/room.rs
davidzhao: oh i was just trying to fix the warnings

livekit-api/src/signal_client/mod.rs
davidzhao: yeah it's defined [here](https://github.com/livekit/livekit/blob/master/pkg/rtc/types/protocol_version.go#L99)

PR #688, opened 2025-08-06
Data channel reliability

ladvoc:
**Summary of changes**:
- Implement data channel reliability improvements from other SDKs:
  - livekit/client-sdk-js/pull/1546
  - livekit/client-sdk-swift/pull/737
- Refactor handling of incoming data packets
  - Take value enum out of packet to avoid large clones 
  - Prefer packet level participant info fields over deprecated user packet fields
- Retain local participant info in session layer
  - Ensure all outgoing data packets have their participant SID and identity fields set
- End-to-end testing: introduce testing with a LiveKit server
  - Utilities for reading configuration from environment, creating multiple room connections
  - Test case to evaluate reliable data channel performance
  - Enable in CI (macOS and Linux only)

### DIFF
--- a/livekit/src/rtc_engine/rtc_session.rs
+++ b/livekit/src/rtc_engine/rtc_session.rs
@@ -975,144 +975,14 @@ impl SessionInner {
                 if !binary {
                     Err(EngineError::Internal("text messages aren't supported".into()))?;
                 }
-
-                let data = proto::DataPacket::decode(&*data).map_err(|err| {
+                let mut packet = proto::DataPacket::decode(&*data).map_err(|err| {
                     EngineError::Internal(format!("failed to decode data packet: {}", err).into())
                 })?;
                 if kind == DataPacketKind::Reliable {
-                    self.update_reliable_received_state(&data);
+                    self.update_reliable_received_state(&packet);
                 }
-                if let Some(packet) = data.value.as_ref() {
-                    match packet {
-                        proto::data_packet::Value::User(user) => {
-                            let participant_sid = user
-                                .participant_sid
-                                .is_empty()
-                                .not()
-                                .then_some(user.participant_sid.clone());
-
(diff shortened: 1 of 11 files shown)


bcherry (approved):
lgtm but would be good for @cnderrauber to confirm as well

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -620,10 +733,17 @@ impl SessionInner {
                                         reliable_buffered_amount -= sent;
                                     }
                                     let threshold = self.reliable_dc_buffered_amount_low_threshold.load(Ordering::Relaxed);
-                                    self._send_until_threshold(threshold, &mut reliable_buffered_amount, &mut reliable_queue);
+                                    self._send_until_threshold(DataPacketKind::Reliable, threshold, &mut reliable_buffered_amount, &mut reliable_queue);
+                                    // TODO: Ensure this is the proper quantity
+                                    let retry_min_amount = (threshold as usize * 5) / 4; // threshold * 1.25
```
cnderrauber: In js client it is the size of buffered messages (sent but not acked by sfu) to avoid lost message during reconnecting, is the `threshold` same as the js client?

ladvoc: The threshold I used here was taken from the Swift implementation ([_DataChannelPair.swift:375_](https://github.com/livekit/client-sdk-swift/pull/737/files#diff-e1083f9e36c8036ad9b3833dc639421b9add7114b7621f181d3f57bbb865920cR375));  I will take a look at how this is handled in JS.

ladvoc: @cnderrauber, I did some stress testing, and it looks like the packet loss [observed in Swift during testing](https://github.com/livekit/client-sdk-swift/pull/737#issuecomment-3073640165) cannot be reproduced in Rust, so I've gone ahead and removed the additional buffering.

PR #689, opened 2025-08-18
Expose participant active event

ladvoc:
This PR exposes the `ParticipantActive` event present in other SDKs, and documents when it is necessary to use it over `ParticipantConnected`. Example usage:
```rs
while let Some(ev) = room_rx.recv().await {
  let RoomEvent::ParticipantActive(participant) = ev else { continue };
  println!("Participant '{}' is active", participant.identity());
  // Send data streams or packets...
}  
```


ladvoc:
Closing in favor of #688

PR #691, opened 2025-08-19
Add send_bytes method

bcherry:


livekit/src/room/mod.rs
```diff
@@ -1721,6 +1727,44 @@ async fn incoming_data_stream_task(
     }
 }
 
+async fn event_for_incoming_stream(reader: AnyStreamReader, identity: String) -> Option<RoomEvent> {
+    let participant_identity = ParticipantIdentity(identity);
+    match reader {
+        AnyStreamReader::Byte(reader) => {
+            if reader.info().attributes.contains_key("__push") {
+                // TODO: avoid cloning info
+                let info = reader.info().clone();
```
1egoman: I'm not sure how big `TextStreamInfo`/`ByteStreamInfo` would be in practice (so I'm not sure how important / valid this suggestion really is) but you could probably get rid of the `.clone()` by adding something like a `fn info_owned(self) -> Self::Info;` method to the `StreamReader` implementation [here](https://github.com/livekit/rust-sdks/blob/9fb6b4d708c5e5d3c82926af10b63b79ef3e0b6f/livekit/src/room/data_stream/incoming.rs#L37-L53) and then moving this line below the `reader.read_all` so `info_owned` could fully take the value / not just operate on a reference?

livekit/src/room/mod.rs
```diff
@@ -1721,6 +1727,44 @@ async fn incoming_data_stream_task(
...
+            Some(RoomEvent::ByteStreamOpened {
+                topic: reader.info().topic.clone(),
+                reader: TakeCell::new(reader),
+                participant_identity,
+            })
+        }
+        AnyStreamReader::Text(reader) => {
+            if reader.info().attributes.contains_key("__push") {
+                // TODO: avoid cloning info
+                let info = reader.info().clone();
```
1egoman: (The same suggestion / caveats from previous comment also apply here!)

livekit/src/room/data_stream/outgoing.rs
```diff
@@ -357,6 +357,8 @@ impl OutgoingStreamManager {
             attached_stream_ids: options.attached_stream_ids,
             generated: options.generated.unwrap_or_default(),
         };
+
+        options.attributes.insert("__push".to_owned(), "1".into());
```
1egoman: I'm not super familiar with what is expected to be within `attributes` in this context by external systems / other clients, but is this just being used as an internal flag within the rust sdk to track whether push is enabled or disabled, or are other external system / other clients depending on this value being set as well?

If it is purely internal, it might be cleaner to have an explicit flag somewhere else? To be fair I can see why you did it this way (I don't see any other obvious places where it could go) but I think worst case, if you have to keep it there, it would be worth giving it a `lk:` prefix like other attribute keys in other places so it should never collide with a customer provided value.

Also no matter whether it is internal or external, IMO it would be good to extract it into a named constant somewhere (or maybe even an enum if there are more values like this) which can make this a bit more self descriptive / not just be a "magic string".

ladvoc: Agreed, this is a temporary solution to demonstrate proof-of-concept. I have proposed adding a dedicated "push" field to stream header to replace this:
- https://github.com/livekit/protocol/pull/1208

bcherry:
What is the benefit of tightly coupling the send and receive methods to only be compatible with the same variant on each side?  

It seems like you could have all of the benefits if you just have the receiving client either register a stream handler or a blob handler (but not both) while the sender can choose to send as a stream or as a blob. Then its fully compatible either way and also gives flexibility to developers to choose implementations that match their needs (for instance, maybe on your memory-constrained hardware device you'll choose to stream a large file from disk to avoid a memory usage spike, but on the agent side you'll just want to receive it in one blob to upload to your LLM).

In JS etc it could be something like this:

```
room.registerByteStreamHandler(for: "my-topic") { reader, participantIdentity in
        for try await chunk in reader {
            // do something
        }
}
```

vs 

```
room.registerBytesHandler(for: "my-topic") { event, participantIdentity in
    // event.bytes would hold all the data
}
```

I think only Rust is more of a challenge due to the events-based structure but to be honest, maybe it's a problem we don't really need to solve anyways. The rust SDK is not used independently by very many developers and the existing `reader.take()` syntax isn't that much more complicated than the proposal here.

btw I don't think clients v3 is a good solve here either. Making client v3 a breaking change vs your existing _code_ is fine (developers just need to update some stuff) but making them incompatible with other _participants_ running older sdk versions is going to make it all very complicated to roll out.

The thing I want us to optimize for in my head is ensuring that agents developers who just adopt a starter agent and starter frontend don't have any difficulties with chat and transcriptions. We would need to tightly coordinate the release of a new agents sdk with new frontend starters for all platforms at the same time to roll out a breaking change here, and we'd still be left with the risk that a developer unknowingly updates their agents SDK (of which the RTC SDK is just a hidden dependency) without updating their frontend SDK as well (not to mention the challenges in actually forcing a quick update when using a mobile app distributed through a store). There's really no good way to introduce a breaking change into the way text transcriptions and chat go back and forth here so we should do it very sparingly.

My advice is to add the new send_bytes API using the existing stream primitives, and separately work on simplifying the receiving side so there's an easier way to just collect whole streams.

ladvoc:
After further discussion, we decided to limit the scope of this PR to only add the `send_bytes` method.

livekit/src/room/data_stream/outgoing.rs
```diff
@@ -386,6 +386,52 @@ impl OutgoingStreamManager {
...
+    /// as a byte stream. It opens a stream using the provided options, writes the
+    /// entire buffer, and closes the stream before returning.
+    ///
+    /// The `total_length` in the header is set from the provided data and is not
+    /// overridable by `options.total_length`.
+    pub async fn send_bytes(
+        &self,
+        data: impl AsRef<[u8]>,
+        options: StreamByteOptions,
+    ) -> StreamResult<ByteStreamInfo> {
```
1egoman, line 401: I get why you wouldn't want to let a user override `total_length` but also I wonder if maybe this could be enforced at the type level instead? I bring this up because it seems like it could be a really easy thing to accidentally set (probably errantly) and not realize that it's not actually doing anything.

One idea - maybe you could add a `SendByteOptions` which doesn't have a `total_length` field in it but otherwise mirrors `StreamByteOptions`? I'd imagine if you went with this option, it might make sense at the same time to do the same thing with the already existing `send_text` too - ie, `options: StreamTextOptions` => `options: SendTextOptions` since it looks like that method also suffers from this same problem (I realize it could be more challenging for the existing method since you'd probably have to make it generic with `options` accepting a trait for it to be able to be done in a backwards compatible way... so maybe migrating the existing one is something that could be done as part of the client v3 initiative)

Alternatively if that's too burdensome maybe an assertion that checks to make sure it isn't set could be another approach, so it at least fails fast / panics versus just silently working.

lukasIO (approved):

livekit/src/room/data_stream/outgoing.rs
ladvoc: I like this idea, but I think it should wait until v3. While I could do this for `send_bytes` now without breaking, I think it is better to avoid introducing a new type that would be inconsistent with `send_text`. For now, I’ve added a warning if this option is set for `send_bytes`.

For v3, something to consider is whether structs for defining options, not just those related to data streams, should have private fields. This would make API changes easier, and, for this case, it would also allow using the type state pattern to only allow setting options that make sense in the context. For example, `with_total_length()` could only be defined on `ByteStreamOptions<Incremental>` for example.

PR #695, opened 2025-08-27
fix: hardware rendering

cs50victor:
Resolves CUDA_ERROR_INVALID_VALUE error caused by attempting to access the video buffer from GPU instead of CPU. Previous SDK hardware rendering attempts on GPU resulted in silent failures and blank image output. This PR corrects the buffer access issue.


cloudwebrtc (approved):
lg！

Issue #699, opened 2025-09-04
Unable to compile libwebrtc 0.3.13


vr7bd:
Hey guys, I've been waiting to try out hardware decoding that was introduced in #698 . I was previously on version 0.3.10 which was working fine. After upgrading to 0.3.13, my program is not able to compile. It fails while compiling `webrtc-sys`. I get a bunch of warnings followed by the error

<details>
<summary>Error logs</summary>

```
warning: webrtc-sys@0.3.10:                  from /usr/include/c++/15/x86_64-redhat-linux/bits/c++config.h:2641,
warning: webrtc-sys@0.3.10:                  from /usr/include/c++/15/bits/requires_hosted.h:31,
warning: webrtc-sys@0.3.10:                  from /usr/include/c++/15/vector:62,
warning: webrtc-sys@0.3.10:                  from src/nvidia/nvidia_encoder_factory.h:6,
warning: webrtc-sys@0.3.10:                  from src/nvidia/nvidia_encoder_factory.cpp:1:
warning: webrtc-sys@0.3.10: /usr/include/features.h:435:4: warning: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Wcpp]
warning: webrtc-sys@0.3.10:   435 | #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
warning: webrtc-sys@0.3.10:       |    ^~~~~~~
warning: webrtc-sys@0.3.10: In file included from src/nvidia/nvidia_encoder_factory.h:11:
warning: webrtc-sys@0.3.10: src/nvidia/cuda_context.h:4:10: fatal error: cuda.h: No such file or directory
warning: webrtc-sys@0.3.10:     4 | #include <cuda.h>
warning: webrtc-sys@0.3.10:       |          ^~~~~~~~
warning: webrtc-sys@0.3.10: compilation terminated.
[...]

PR #701, opened 2025-09-09
chore: Optional flags for video hw codec.

cloudwebrtc:
and fixed build issue for ios sim.

### DIFF
--- a/livekit-ffi/Cargo.toml
+++ b/livekit-ffi/Cargo.toml
@@ -19,6 +19,7 @@ tracing = ["tokio/tracing", "console-subscriber"]
 
 [dependencies]
 livekit = { workspace = true }
+webrtc-sys = { workspace = true , features = [ "use_vaapi", "use_nvidia" ]}
 soxr-sys = { workspace = true }
 imgproc = { workspace = true }
 livekit-protocol = { workspace = true }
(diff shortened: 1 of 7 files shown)


livekit-ffi/Cargo.toml
```diff
@@ -19,6 +19,7 @@ tracing = ["tokio/tracing", "console-subscriber"]
 
 [dependencies]
 livekit = { workspace = true }
+webrtc-sys = { workspace = true , features = [ "use_vaapi", "use_nvidia" ]}
```
ladvoc, line 22: Do you think these features should be enabled by default, or should they be gated behind a feature on _livekit-ffi_?

chenosaurus: I was in the processing of creating a new python-sdk release with `livekit-ffi@0.12.32` that contains the hw encoder changes.  This might be relevant to wait for.  I think default enabled?

cloudwebrtc: based on this issue https://github.com/livekit/rust-sdks/issues/699, some users will use the Rust SDK, but they do not have NVIDIA hardware and drivers, so users may want to manually enable it (because the Rust app will be compiled from the webrtc-sys source code when used). Although LiveKit FFI enables vaapi or NVIDIA HW support at compile time, because FFI supports lazy loading of vaapi/nvidia drivers, even if the user does not install these drivers, FFI can still run (just without loading vaapi/nvidia drivers and hw codec supports)

ladvoc (approved):
LGTM ✅

PR #708, opened 2025-09-18
feature(e2ee): add data channel encryption

lukasIO:
this PR adds data channel encryption capabilities. 
For backwards compatibility this is not enabled on existing implementations. 
Instead `RoomOptions.e2ee` is being deprecated (no dc encryption) and a new `RoomOptions.encryption` field is introduced which enables data channel encryption


livekit/src/room/mod.rs
```diff
@@ -459,7 +462,9 @@ impl Room {
         options: RoomOptions,
     ) -> RoomResult<(Self, mpsc::UnboundedReceiver<RoomEvent>)> {
         // TODO(theomonnom): move connection logic to the RoomSession
-        let e2ee_manager = E2eeManager::new(options.e2ee.clone());
+        let with_dc_encryption = options.encryption.is_some();
+        let encryption_options = options.encryption.clone().or(options.e2ee.clone());
```
ladvoc: Since connect takes ownership of room options, you could mark the argument as mut (`mut options: RoomOptions`) and use `take` instead of `clone` for `options.encryption` and `options.e2ee`.

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -1128,13 +1136,130 @@ impl SessionInner {
...
+                }
+            }
             _ => Ok(()),
         };
         if let Err(err) = send_result {
             log::error!("failed to emit incoming data packet: {:?}", err);
         }
     }
 
+    fn handle_decrypted_payload(
```
ladvoc: Instead of creating a new method for this, I think we could modify the signature for `emit_incoming_packet` to accept `participant_sid` and `participant_identity` (instead of the entire packet) and then call it recursively to handle the decrypted packet's value.

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -1346,6 +1471,82 @@ impl SessionInner {
...
+                                proto::EncryptedPacket {
+                                    encrypted_value: encrypted_data.data,
+                                    iv: encrypted_data.iv,
+                                    key_index: encrypted_data.key_index,
+                                    encryption_type: e2ee_manager.encryption_type().into(),
+                                },
+                            ));
+                        }
+                        Err(e) => {
+                            log::warn!("Failed to encrypt data packet: {}", e);
```
ladvoc: If E2EE is enabled and encryption fails, it seems to me `publish_data` should return the error and not continue?

examples/encrypted_text_stream/src/main.rs
```diff
@@ -0,0 +1,172 @@
...
+};
+use livekit_api::access_token;
+use std::{env, error::Error, io::Write};
+use tokio::{
+    io::{self, AsyncBufReadExt, BufReader},
+    sync::mpsc::UnboundedReceiver,
+};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
ladvoc, line 18: Nice example!

livekit/src/room/mod.rs
```diff
@@ -350,7 +350,9 @@ pub struct RoomOptions {
     pub auto_subscribe: bool,
     pub adaptive_stream: bool,
     pub dynacast: bool,
+    #[deprecated(note = "Use `encryption` field instead, see x for a detailed explanation")]
```
ladvoc: Do we have something ready to link for "x"?

livekit/tests/data_channel_encryption.rs
```diff
@@ -0,0 +1,69 @@
...
+mod common;
+
+#[cfg(feature = "__lk-e2e-test")]
+#[tokio::test]
+async fn test_data_channel_encryption() -> Result<()> {
+    const ITERATIONS: usize = 128;
+    const PAYLOAD_SIZE: usize = 4096;
+
+    let key_provider =
+        KeyProvider::with_shared_key(KeyProviderOptions::default(), "password".as_bytes().to_vec());
```
ladvoc, line 25: For the test, we should initialize a separate key provider for each room instead of cloning since they share the same atomic value inside—I think I messed this up when I modified the test.

livekit/src/rtc_engine/rtc_session.rs
lukasIO: makes sense

livekit/src/room/mod.rs
lukasIO: nice, I'll do that

livekit/src/room/mod.rs
lukasIO: currently wip, will wait for the actual link to update and merge

livekit/src/rtc_engine/rtc_session.rs
lukasIO: sounds good!

livekit/src/rtc_engine/rtc_session.rs
ladvoc: To do that, I think the best approach is to define a new error type (i.e. `EncryptionError`) and add a case for it in the `RoomError` enum. I'm not sure how much information can be extracted from the underlying exception from _libwebrtc_, so it probably makes sense to just dump the error string inside `EncryptionError`.

lukasIO: wouldn't introducing a new error type be a breaking change for match clauses? 🤔 
there's no `non_exhaustive` declared for it afaict

ladvoc: Ah good point, I had assumed it was declared as non exhaustive—it should be for v3. Given that it isn't, I think using the existing `Rtc` case is the way to go.

lukasIO: for `Rtc` I'm not able to supply a string argument and the error type is defined as `webrtc internal error`. 

I was going for `Internal` as an error in order to be able to supply a reason

livekit/src/room/data_stream/incoming.rs
```diff
@@ -246,15 +247,21 @@ impl IncomingStreamManager {
...
-    pub fn handle_header(&self, header: proto::Header, identity: String) {
-        let Ok(info) =
-            AnyStreamInfo::try_from(header).inspect_err(|e| log::error!("Invalid header: {}", e))
+    pub fn handle_header(
+        &self,
+        header: proto::Header,
+        identity: String,
+        encryption_type: livekit_protocol::encryption::Type,
+    ) {
+        let Ok(info) = AnyStreamInfo::try_from_with_encryption(header, encryption_type.into())
```
lukasIO, line 256: @ladvoc sorry, ended up messing with this anyways as I feel having the encryption type on the stream info does in fact make sense? 
Let me know if you think that's unnecessary and I'll remove this again

livekit/src/room/data_stream/mod.rs
```diff
@@ -59,6 +61,9 @@ pub enum StreamError {
 
     #[error("internal error")]
     Internal,
+
+    #[error("encryption type mismatch")]
+    EncryptionTypeMismatch,
```
lukasIO, line 66: @ladvoc is it ok to introduce this as a new error type? I think we also forgot non_exhaustive here?

livekit/src/room/data_stream/mod.rs
```diff
@@ -136,23 +145,40 @@ impl TryFrom<proto::Header> for AnyStreamInfo {
     type Error = StreamError;
 
     fn try_from(mut header: proto::Header) -> Result<Self, Self::Error> {
+        Self::try_from_with_encryption(header, EncryptionType::None)
```
lukasIO, line 148: if we keep the new try_from_with_encryption, we could probably remove this one here

livekit/src/room/data_stream/incoming.rs
ladvoc: Yes, I think it makes sense to enforce the encryption type specified in the header.

livekit/src/room/data_stream/mod.rs
ladvoc: The more I think about this, while technically this should be non_exhaustive, there is almost no situation where an existing application would be matching the error without a catchall branch (_ => {}). For that reason, I think it is safe to add.

PR #709, opened 2025-09-18
fix: apply original participant fields in data messages

s-hamdananwar:
Participant SID and Identity were being set as empty since `.then_some(user.participant_sid)` would return `None` when original fields has some value (ie. not none)


s-hamdananwar:
@chenosaurus are the build errors safe to be ignored?

ladvoc:
I also thought these build issues were fixed, but it looks like they are not related to this PR.

PR #714, opened 2025-09-23
Do not modify raw packets

ladvoc:
Normally, the participant identity and SID fields should be set for any outgoing packet based on the local participant's info; #688 introduced setting these fields at the session level to enforce this. However, this should not apply to raw packets. 

Currently, Python uses raw packets for data streams. In the case of user transcriptions, the sender identity field is explicitly set to the user’s identity to indicate that the transcription corresponds to the user’s voice rather than the agent’s—even though the agent is technically the sender.


cloudwebrtc (approved):
lgtm!

s-hamdananwar (approved):
lgtm!

PR #715, opened 2025-09-23
fix Builds/E2E Tests CI.

cloudwebrtc:


libwebrtc/src/video_source.rs
```diff
@@ -58,7 +58,7 @@ pub mod native {
 
     impl Default for NativeVideoSource {
         fn default() -> Self {
-            Self::new(VideoResolution::default())
+            Self::new(VideoResolution { width: 640, height: 480 })
```
davidzhao: is this the right default?

cloudwebrtc: VideoResolution::default() defaults to 0x0. I will add a default value for VideoResolution to use 1280x720 consistent with the Client SDKs.

PR #725, opened 2025-10-05
Expose desktop capturer

iparaskev:
Implements screen sharing by exposing libwebrtc's `DesktopCapturer`. 

A few notes:
* I have exposed only a subset of [DesktopCaptureOptions](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/desktop_capture/desktop_capture_options.h;l=36?q=desktopCaptureOptions&ss=chromium) which have been tested with [Hopp](https://github.com/gethopp/hopp).
* On macos the only way for the user to select a source when sharing a display is via the system picker as libwebrtc hasn't implemented getting the source list. In [our fork](https://github.com/gethopp/webrtc/commit/8948bdab0b2ceeac76c820cb99d50292559179f0#diff-f85e2fba4aa1cf9f0f9310861143acb8afb0f502a36ce2aea55951d64f0100ae) I have extended `GetSourceList` to actually get the available displays without the system picker. I could port my patch to your libwebrtc fork if you want. For selecting windows `GetSourceList` seems to work.
* AFAIK when using pipewire the only way to select a source is via the system picker.
* I have put the commit on top of another open PR I have, which enables buffer scaling. This makes sharing the buffer much easier, because we don't need to know in advance the source dims. If you have another way for scaling buffers I would be happy to use it instead of my patch.

Tested on:
* windows 11
* macOS Tahoe
* ubuntu 25.04 (wayland)

This is related to #92 and https://github.com/zed-industries/zed/issues/28754.

### DIFF
--- a/examples/screensharing/src/main.rs
+++ b/examples/screensharing/src/main.rs
@@ -1,0 +1,166 @@
+use clap::Parser;
+use livekit::options::{TrackPublishOptions, VideoCodec};
+use livekit::prelude::*;
+use livekit::track::{LocalTrack, LocalVideoTrack, TrackSource};
+use livekit::webrtc::desktop_capturer::{
+    CaptureResult, DesktopCapturer, DesktopCapturerOptions, DesktopFrame,
+};
+use livekit::webrtc::native::yuv_helper;
+use livekit::webrtc::prelude::{
+    I420Buffer, RtcVideoSource, VideoBuffer, VideoFrame, VideoResolution, VideoRotation,
+};
+use livekit::webrtc::video_source::native::NativeVideoSource;
+use livekit_api::access_token;
+use std::env;
+use std::sync::Arc;
+use std::sync::Mutex;
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// Capture the mouse cursor
+    #[arg(long)]
+    capture_cursor: bool,
+
+    /// Capture a specific window (requires window ID)
+    #[arg(long)]
+    capture_window: bool,
+
+    /// Use system screen picker (macOS only)
+    #[cfg(target_os = "macos")]
+    #[arg(long)]
+    use_system_picker: bool,
+}
+
+#[tokio::main]
+async fn main() {
+    env_logger::init();
+    let args = Args::parse();
+
+    #[cfg(target_os = "linux")]
+    {
+        /* This is needed for getting the system picker for screen sharing. */
+        use glib::MainLoop;
+        let main_loop = MainLoop::new(None, false);
+        let _handle = std::thread::spawn(move || {
+            main_loop.run();
+        });
+    }
+
+    let url = env::var("LIVEKIT_URL").expect("LIVEKIT_URL is not set");
+    let api_key = env::var("LIVEKIT_API_KEY").expect("LIVEKIT_API_KEY is not set");
+    let api_secret = env::var("LIVEKIT_API_SECRET").expect("LIVEKIT_API_SECRET is not set");
+
+    let token = access_token::AccessToken::with_api_key(&api_key, &api_secret)
+        .with_identity("rust-bot")
+        .with_name("Rust Bot")
+        .with_grants(access_token::VideoGrants {
+            room_join: true,
+            room: "dev_room".to_string(),
+            ..Default::default()
+        })
+        .to_jwt()
+        .unwrap();
+
+    let (room, _) = Room::connect(&url, &token, RoomOptions::default()).await.unwrap();
+    log::info!("Connected to room: {} - {}", room.name(), String::from(room.sid().await));
+
+    let stream_width = 1920;
+    let stream_height = 1080;
+    let buffer_source =
+        NativeVideoSource::new(VideoResolution { width: stream_width, height: stream_height });
+    let track = LocalVideoTrack::create_video_track(
+        "screen_share",
+        RtcVideoSource::Native(buffer_source.clone()),
+    );
+
+    room.local_participant()
+        .publish_track(
+            LocalTrack::Video(track),
+            TrackPublishOptions {
+                source: TrackSource::Screenshare,
+                video_codec: VideoCodec::VP9,
+                ..Default::default()
+            },
+        )
+        .await
+        .unwrap();
+
+    let buffer_source_clone = buffer_source.clone();
+    let video_frame = Arc::new(Mutex::new(VideoFrame {
+        rotation: VideoRotation::VideoRotation0,
+        buffer: I420Buffer::new(stream_width, stream_height),
+        timestamp_us: 0,
+    }));
+    let capture_buffer = Arc::new(Mutex::new(I420Buffer::new(stream_width, stream_height)));
+    let callback = move |result: CaptureResult, frame: DesktopFrame| {
+        match result {
+            CaptureResult::ErrorTemporary => {
+                log::info!("Error temporary");
+                return;
+            }
+            CaptureResult::ErrorPermanent => {
+                log::info!("Error permanent");
+                return;
+            }
+            _ => {}
+        }
+        let video_frame = video_frame.clone();
+        let height = frame.height();
+        let width = frame.width();
+
+        {
+            let mut capture_buffer = capture_buffer.lock().unwrap();
+            let capture_buffer_width = capture_buffer.width() as i32;
+            let capture_buffer_height = capture_buffer.height() as i32;
+            if height != capture_buffer_height || width != capture_buffer_width {
+                *capture_buffer = I420Buffer::new(width as u32, height as u32);
+            }
+        }
+
+        let stride = frame.stride();
+        let data = frame.data();
+
+        let mut capture_buffer = capture_buffer.lock().unwrap();
+        let (s_y, s_u, s_v) = capture_buffer.strides();
+        let (y, u, v) = capture_buffer.data_mut();
+        yuv_helper::argb_to_i420(data, stride, y, s_y, u, s_u, v, s_v, width, height);
+
+        let scaled_buffer = capture_buffer.scale(stream_width as i32, stream_height as i32);
+        let (scaled_y, scaled_u, scaled_v) = scaled_buffer.data();
+
+        let mut framebuffer = video_frame.lock().unwrap();
+        let buffer = &mut framebuffer.buffer;
+        let (y, u, v) = buffer.data_mut();
+        y.copy_from_slice(scaled_y);
+        u.copy_from_slice(scaled_u);
+        v.copy_from_slice(scaled_v);
+
+        buffer_source_clone.capture_frame(&*framebuffer);
+    };
+    let mut options = DesktopCapturerOptions::new();
+    #[cfg(target_os = "macos")]
+    {
+        options.set_sck_system_picker(args.use_system_picker);
+    }
+    options.set_window_capturer(args.capture_window);
+    options.set_include_cursor(args.capture_cursor);
+    #[cfg(target_os = "linux")]
+    {
+        options.set_pipewire_capturer(true);
+    }
+
+    let mut capturer =
+        DesktopCapturer::new(callback, options).expect("Failed to create desktop capturer");
+    let sources = capturer.get_source_list();
+    log::info!("Found {} sources", sources.len());
+
+    let selected_source = sources.first().cloned();
+    capturer.start_capture(selected_source);
+
+    let now = tokio::time::Instant::now();
+    while now.elapsed() < tokio::time::Duration::from_secs(30) {
+        capturer.capture_frame();
+        tokio::time::sleep(tokio::time::Duration::from_millis(16)).await;
+    }
+}
--- a/libwebrtc/src/desktop_capturer.rs
+++ b/libwebrtc/src/desktop_capturer.rs
@@ -1,0 +1,239 @@
+// Copyright 2025 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::imp::desktop_capturer as imp_dc;
+
+/// Configuration options for creating a desktop capturer.
+///
+/// It contains a subset of libwertc's DesktopCaptureOptions.
+///
+/// By default, it captures the entire screen and does not include the cursor.
+///
+/// # Example
+/// ```no_run
+/// use libwebrtc::desktop_capturer::DesktopCapturerOptions;
+///
+/// let mut options = DesktopCapturerOptions::new();
+/// options.set_include_cursor(true);
+/// ```
+pub struct DesktopCapturerOptions {
+    pub(crate) sys_handle: imp_dc::DesktopCapturerOptions,
+}
+
+impl DesktopCapturerOptions {
+    /// Creates a new `DesktopCapturerOptions` with default values.
+    pub fn new() -> Self {
+        Self { sys_handle: imp_dc::DesktopCapturerOptions::new() }
+    }
+
+    /// Sets whether to include the cursor in captured frames.
+    pub fn set_include_cursor(&mut self, include: bool) {
+        self.sys_handle = self.sys_handle.with_cursor(include);
+    }
+
+    /// Sets whether to capture windows instead of the entire screen.
+    pub fn set_window_capturer(&mut self, window_capturer: bool) {
+        self.sys_handle = self.sys_handle.with_window_capturer(window_capturer);
+    }
+
+    /// Sets whether to allow the ScreenCaptureKit (SCK) capturer on macOS.
+    ///
+    /// This is enabled by default.
+    #[cfg(target_os = "macos")]
+    pub fn set_sck_capturer(&mut self, allow_sck_capturer: bool) {
+        self.sys_handle = self.sys_handle.with_sck_capturer(allow_sck_capturer);
+    }
+
+    /// Sets whether to allow the ScreenCaptureKit system picker on macOS.
+    ///
+    /// This is enabled by default.
+    ///
+    /// When disabled, for capturing displays the client should get the source id
+    /// via a different way as [`DesktopCapturer::get_source_list`] returns an empty vector.
+    #[cfg(target_os = "macos")]
+    pub fn set_sck_system_picker(&mut self, allow_sck_system_picker: bool) {
+        self.sys_handle = self.sys_handle.with_sck_system_picker(allow_sck_system_picker);
+    }
+
+    /// Sets whether to allow the Windows Graphics Capture (WGC) capturer on Windows.
+    ///
+    /// This is enabled by default.
+    #[cfg(target_os = "windows")]
+    pub fn set_wgc_capturer(&mut self, allow_wgc_capturer: bool) {
+        self.sys_handle = self.sys_handle.with_wgc_capturer(allow_wgc_capturer);
+    }
+
+    /// Sets whether to allow the DirectX capturer on Windows.
+    ///
+    /// This is enabled by default.
+    #[cfg(target_os = "windows")]
+    pub fn set_directx_capturer(&mut self, allow_directx_capturer: bool) {
+        self.sys_handle = self.sys_handle.with_directx_capturer(allow_directx_capturer);
+    }
+
+    /// Sets whether to allow the PipeWire capturer on Linux.
+    ///
+    /// It should be enabled when using Wayland.
+    #[cfg(target_os = "linux")]
+    pub fn set_pipewire_capturer(&mut self, allow_pipewire_capturer: bool) {
+        self.sys_handle = self.sys_handle.with_pipewire_capturer(allow_pipewire_capturer);
+    }
+}
+
+/// A desktop capturer for capturing screens or windows.
+pub struct DesktopCapturer {
+    handle: imp_dc::DesktopCapturer,
+}
+
+impl DesktopCapturer {
+    /// Creates a new `DesktopCapturer` with the specified callback and options.
+    ///
+    /// # Arguments
+    ///
+    /// * `callback` - A function that will be called for each captured frame. The callback
+    ///   receives a [`CaptureResult`] indicating success or error, and a [`DesktopFrame`]
+    ///   containing the captured image data.
+    /// * `options` - Configuration options for the capturer
+    ///
+    /// # Returns
+    ///
+    /// Returns `Some(DesktopCapturer)` if the capturer was created successfully,
+    /// or `None` if creation failed (e.g., due to platform limitations or permissions).
+    pub fn new<T>(callback: T, options: DesktopCapturerOptions) -> Option<Self>
+    where
+        T: Fn(CaptureResult, DesktopFrame) + Send + 'static,
+    {
+        let inner_callback = move |result: imp_dc::CaptureResult, frame: imp_dc::DesktopFrame| {
+            callback(capture_result_from_sys(result), DesktopFrame::new(frame));
+        };
+        let desktop_capturer = imp_dc::DesktopCapturer::new(inner_callback, options.sys_handle);
+        if desktop_capturer.is_none() {
+            return None;
+        }
+        Some(Self { handle: desktop_capturer.unwrap() })
+    }
+
+    /// Starts capturing from the specified source.
+    ///
+    /// # Arguments
+    ///
+    /// * `source` - The capture source to use. It should be None when the capturer
+    /// is configured to use the system picker (on platforms that support it).
+    ///
+    /// # Note
+    ///
+    /// After calling this method, you must call [`capture_frame`](Self::capture_frame)
+    /// to actually capture frames. This method only initializes the capture session.
+    pub fn start_capture(&mut self, source: Option<CaptureSource>) {
+        if let Some(source) = source {
+            self.handle.select_source(source.sys_handle.id());
+        }
+        self.handle.start();
+    }
+
+    /// Captures a single frame.
+    ///
+    /// You must call [`start_capture`](Self::start_capture) before calling this method.
+    pub fn capture_frame(&mut self) {
+        self.handle.capture_frame();
+    }
+
+    /// Retrieves a list of available capture sources.
+    ///
+    /// Returns a list of screens or windows that can be captured, depending
+    /// on whether the capturer was configured for window or screen capture.
+    ///
+    /// # Returns
+    ///
+    /// A vector of [`CaptureSource`] objects representing available capture sources.
+    pub fn get_source_list(&self) -> Vec<CaptureSource> {
+        let source_list = self.handle.get_source_list();
+        source_list.into_iter().map(|source| CaptureSource { sys_handle: source }).collect()
+    }
+}
+
+pub struct DesktopFrame {
+    pub(crate) sys_handle: imp_dc::DesktopFrame,
+}
+
+impl DesktopFrame {
+    pub fn new(sys_handle: imp_dc::DesktopFrame) -> Self {
+        Self { sys_handle }
+    }
+
+    pub fn width(&self) -> i32 {
+        self.sys_handle.width() as i32
+    }
+
+    pub fn height(&self) -> i32 {
+        self.sys_handle.height() as i32
+    }
+
+    pub fn stride(&self) -> u32 {
+        self.sys_handle.stride() as u32
+    }
+
+    pub fn left(&self) -> i32 {
+        self.sys_handle.left()
+    }
+
+    pub fn top(&self) -> i32 {
+        self.sys_handle.top()
+    }
+
+    pub fn data(&self) -> &[u8] {
+        &self.sys_handle.data()
+    }
+}
+
+#[derive(Clone)]
+pub struct CaptureSource {
+    pub(crate) sys_handle: imp_dc::CaptureSource,
+}
+
+impl CaptureSource {
+    pub fn id(&self) -> u64 {
+        self.sys_handle.id()
+    }
+    pub fn title(&self) -> String {
+        self.sys_handle.title()
+    }
+    pub fn display_id(&self) -> i64 {
+        self.sys_handle.display_id()
+    }
+}
+
+impl std::fmt::Display for CaptureSource {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CaptureSource")
+            .field("id", &self.id())
+            .field("title", &self.title())
+            .field("display_id", &self.display_id())
+            .finish()
+    }
+}
+
+pub enum CaptureResult {
+    Success,
+    ErrorTemporary,
+    ErrorPermanent,
+}
+
+fn capture_result_from_sys(result: imp_dc::CaptureResult) -> CaptureResult {
+    match result {
+        imp_dc::CaptureResult::Success => CaptureResult::Success,
+        imp_dc::CaptureResult::ErrorTemporary => CaptureResult::ErrorTemporary,
+        imp_dc::CaptureResult::ErrorPermanent => CaptureResult::ErrorPermanent,
+    }
+}
--- a/libwebrtc/src/native/desktop_capturer.rs
+++ b/libwebrtc/src/native/desktop_capturer.rs
@@ -1,0 +1,261 @@
+// Copyright 2025 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use cxx::UniquePtr;
+use webrtc_sys::desktop_capturer::{self as sys_dc, ffi::new_desktop_capturer};
+
+#[derive(Copy, Clone, Debug)]
+pub struct DesktopCapturerOptions {
+    pub window_capturer: bool,
+    pub include_cursor: bool,
+    #[cfg(target_os = "macos")]
+    pub allow_sck_capturer: bool,
+    #[cfg(target_os = "macos")]
+    pub allow_sck_system_picker: bool,
+    #[cfg(target_os = "windows")]
+    pub allow_wgc_capturer: bool,
+    #[cfg(target_os = "windows")]
+    pub allow_directx_capturer: bool,
+    #[cfg(target_os = "linux")]
+    pub allow_pipewire_capturer: bool,
+}
+
+impl Default for DesktopCapturerOptions {
+    fn default() -> Self {
+        Self {
+            window_capturer: false,
+            include_cursor: false,
+            #[cfg(target_os = "macos")]
+            allow_sck_capturer: true,
+            #[cfg(target_os = "macos")]
+            allow_sck_system_picker: true,
+            #[cfg(target_os = "windows")]
+            allow_wgc_capturer: true,
+            #[cfg(target_os = "windows")]
+            allow_directx_capturer: true,
+            #[cfg(target_os = "linux")]
+            allow_pipewire_capturer: false,
+        }
+    }
+}
+
+impl DesktopCapturerOptions {
+    pub(crate) fn new() -> Self {
+        Self { window_capturer: false, include_cursor: false, ..Default::default() }
+    }
+
+    pub(crate) fn with_cursor(mut self, include: bool) -> Self {
+        self.include_cursor = include;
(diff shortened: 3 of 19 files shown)


examples/screensharing/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    #[cfg(target_os = "linux")]
+    {
+        /* This is needed for getting the system picker for screen sharing. */
+        use glib::MainLoop;
+        let main_loop = MainLoop::new(None, false);
+        let _handle = std::thread::spawn(move || {
+            main_loop.run();
+        });
+    }
```
Be-ing: What is the need for this GLib event loop? With the GStreamer Rust bindings for example, GStreamer events are usually handled by a GLib event loop, but GStreamer has an API to register an event callback that the library calls internally, which can be used from Rust to send events to a Rust thread or async stream over a channel. This way the Rust program calling the library doesn't need to setup its own GLib event loop. Here's an example: https://codeberg.org/Be.ing/dance-video-recorder/src/commit/4aab73f5514ff08000ef330f221dc6035831a4ee/src/camera_monitor.rs#L22

examples/screensharing/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+    #[cfg(target_os = "macos")]
+    {
+        options.set_sck_system_picker(args.use_system_picker);
+    }
+    options.set_window_capturer(args.capture_window);
+    options.set_include_cursor(args.capture_cursor);
+    #[cfg(target_os = "linux")]
+    {
+        options.set_pipewire_capturer(true);
+    }
```
Be-ing: What happens if this is false? I presume libwebrtc tries to capture via X11? If so, this should check at runtime whether it's being run under X11 or Wayland rather than assuming `target_os = "linux"` means Wayland.

Be-ing:
I will give this a closer look and testing this week. Big thanks for adding a new example program!

examples/screensharing/src/main.rs
iparaskev: Yep the proper way would be to check on runtime and not enable it (I have no idea if pipewire works with X11). I didn't do it here because it would have taken more time to set up the X11 dependencies and my vm. I will try to test on X11 during the week.

examples/screensharing/src/main.rs
iparaskev: Thanks I didn't know about this. I will update the example to do it this way.

Be-ing: I'm not sure if such an approach is feasible without a clearer idea of what the need for the GLib event loop is.

Be-ing: In the example I linked above, the GStreamer bindings are creating a futures_channel::mpsc under the hood to create an ergonomic Rust stream API: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/c533fe960af057c55916ba3fb48c42837da98565/gstreamer/src/bus.rs#L345

iparaskev: >  the need for the GLib event loop is

The event loop is needed for opening the system ui picker. It is related to the xdg_portal setup. Maybe there is something [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/desktop_capture/linux/wayland/screen_capture_portal_interface.cc). I need to have a better look.

Be-ing: Ah, so libwebrtc invokes GDBus under the hood. That's inconvenient for usage from Rust, but okay. At least this requirement should be documented for Livekit.

It would be nice to use zbus/ashpd instead, but at a quick glance that would require somewhat extensive changes to libwebrtc's C++ code to call into Rust (via cxx) instead of GDBus. If you're up for doing that, that'd be great, but no pressing need for it. I might give that a try eventually.

Be-ing: It would be nice if Livekit handled starting the GLib main loop so users of the library wouldn't have to, but that wouldn't be good for applications that are already using GLib for something such as the GStreamer or GTK Rust bindings. So, best to leave it to the user of the library with documentation.

iparaskev: Yes I agree that is better to leave it to the client to decide what to do with the main loop.

examples/screensharing/src/main.rs
iparaskev: I managed to set up the dependencies and tested the example on X11 too.

webrtc-sys/build/src/lib.rs
```diff
@@ -117,20 +118,28 @@ pub fn webrtc_dir() -> path::PathBuf {
 pub fn webrtc_defines() -> Vec<(String, Option<String>)> {
     // read preprocessor definitions from webrtc.ninja
     let defines_re = Regex::new(r"-D(\w+)(?:=([^\s]+))?").unwrap();
-    let webrtc_gni = fs::File::open(webrtc_dir().join("webrtc.ninja")).unwrap();
+    let files = [webrtc_dir().join("webrtc.ninja"), webrtc_dir().join("desktop_capture.ninja")];
```
Be-ing: What's the need for this change? I don't see desktop_capturer.cpp nor desktop_capturer.h referencing any compiler definitions from desktop_capture.ninja.

iparaskev: `DesktopCaptureOptions` does. In particular without `WEBRTC_USE_X11` [this](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/desktop_capture/desktop_capture_options.h;l=259?q=desktopCaptureOptions&ss=chromium%2Fchromium%2Fsrc) is compiled out in the header used when building livekit and then you have a mismatch with what `libwebrtc` is using. I learned it by getting a very nice stack smashing error when running the example.

examples/screensharing/src/main.rs
Be-ing: libwebrtc takes care of checking whether it's running under Wayland or X11 and picking the appropriate backend, so there's no need to make the downstream user of this library concerned with setting this option. I removed the option in https://github.com/gethopp/rust-sdks/pull/3

examples/screensharing/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+        u.copy_from_slice(scaled_u);
+        v.copy_from_slice(scaled_v);
+
+        buffer_source_clone.capture_frame(&*framebuffer);
+    };
+    let mut options = DesktopCapturerOptions::new();
+    #[cfg(target_os = "macos")]
+    {
+        options.set_sck_system_picker(args.use_system_picker);
+    }
```
Be-ing: Is there any reason for this to be false on macOS? Can we just hardcode it to true and remove this platform-specific code from the application?

I don't have a Mac available for testing.

examples/screensharing/src/main.rs
Be-ing: In https://github.com/gethopp/rust-sdks/pull/3 I moved this into the library behind a Cargo feature flag. Most Rust applications won't need to be concerned with this, but if they want to opt out, they can disable default features for the libwebrtc crate.

iparaskev: I would prefer the maintainers of LiveKit to decide how they want to handle this, in the end of the day is their project. I was thinking that is more appropriate to have desktop capturer as a feature in general.

examples/screensharing/src/main.rs
iparaskev: Good point.

examples/screensharing/src/main.rs
iparaskev: > Can we just hardcode it to true 

macOS allow you to select a source without the system picker, which is valuable for applications that want to implement their own picker (this is what we do at Hopp). IMO it is useful to give the option to the application to decide how to do its source selection..

Be-ing: okay

examples/screensharing/src/main.rs
```diff
@@ -138,19 +138,17 @@ async fn main() {
 
         buffer_source_clone.capture_frame(&*framebuffer);
     };
-    let mut options = DesktopCapturerOptions::new();
+    let source_type = if args.capture_window {
+        DesktopCaptureSourceType::WINDOW
+    } else {
+        DesktopCaptureSourceType::SCREEN
+    };
+    let mut options = DesktopCapturerOptions::new(source_type);
```
Be-ing: Oh I like this API better, nice idea. It's a bit odd that the WINDOW and SCREEN enum variants are all caps though.

webrtc-sys/build/src/lib.rs
```diff
@@ -118,7 +118,15 @@ pub fn webrtc_dir() -> path::PathBuf {
...
-    let files = [webrtc_dir().join("webrtc.ninja"), webrtc_dir().join("desktop_capture.ninja")];
+    let mut files = vec![webrtc_dir().join("webrtc.ninja")];
+
+    // Avoid ABI mismatch for DesktopCaptureOptions due to WEBRTC_USE_X11 missing
+    let desktop_capture_path = webrtc_dir().join("desktop_capture.ninja");
+    if cfg!(target_os = "linux") {
+        files.push(desktop_capture_path);
+    } else if desktop_capture_path.exists() {
+        files.push(desktop_capture_path);
+    }
```
Be-ing: The conditions of this if-else block don't make much sense to me. Shouldn't this just check if the path exists regardless of target_os?

examples/screensharing/src/main.rs
Be-ing: My thinking is that this library is likely to get used by cross-platform applications whose developers may or may not know anything about Linux. It would be easy for a downstream developer who doesn't know anything about Linux to overlook starting a GLib event loop and have their application not work on Wayland without realizing it. So I think it'd be good to have it just work by default and let developers opt out if they don't need it. I don't have a strong opinion on this either way though; whatever the Livekit developers decide is fine.

> I was thinking that is more appropriate to have desktop capturer as a feature in general.

Sure, though that's orthogonal to this question about the GLib event loop.

examples/screensharing/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+    log::info!("Connected to room: {} - {}", room.name(), String::from(room.sid().await));
+
+    let stream_width = 1920;
+    let stream_height = 1080;
+    let buffer_source =
+        NativeVideoSource::new(VideoResolution { width: stream_width, height: stream_height });
+    let track = LocalVideoTrack::create_video_track(
+        "screen_share",
+        RtcVideoSource::Native(buffer_source.clone()),
+    );
```
Be-ing: I'm confused what the right thing to do is here for the resolution. Unless I'm missing something, libwebrtc doesn't expose an API to get the resolution of a DesktopCapturer::Source; you can only get the resolution of the captured screen/window from the DesktopFrame passed to the callback after the stream has started. Nor is there a cross-platform way to change the resolution of the VideoTrackSource after it is created. There is a method on DesktopCapturer `void UpdateResolution(uint32_t width, uint32_t height)` which is guarded by `defined(WEBRTC_USE_PIPEWIRE) || defined(WEBRTC_USE_X11)` though.

Be-ing: Maybe someone from Livekit has some idea about this?

examples/screensharing/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        let stride = frame.stride();
+        let data = frame.data();
+
+        let mut capture_buffer = capture_buffer.lock().unwrap();
+        let (s_y, s_u, s_v) = capture_buffer.strides();
+        let (y, u, v) = capture_buffer.data_mut();
+        yuv_helper::argb_to_i420(data, stride, y, s_y, u, s_u, v, s_v, width, height);
+
+        let scaled_buffer = capture_buffer.scale(stream_width as i32, stream_height as i32);
```
Be-ing: This scaling is redundant. NativeVideoSource::capture_frame will scale it automatically.

iparaskev: Yes you are right, I had missed the `VideoTrackSource::InternalSource::on_captured_frame` implementation. Thanks!

iparaskev: Actually I had another look and I think the scaling is actually needed. When setting the `NativeVideoSource` to 1080p and the frame dims are 2880x1800, `adapted_width/height` are becoming 2880x1800 and not 1080p.

examples/screensharing/src/main.rs
Be-ing: Digging deeper into the NativeVideoSource code, it seems the requirement for the video resolution is for the scaling performed by VideoTrackSource::InternalSource::on_captured_frame. I was wondering if this was a requirement imposed by libwebrtc somehow, but in that case, it wouldn't make sense that libwebrtc doesn't provide an API to get the resolution before starting the stream. If I understand that correctly, then I think we can add a method to NativeVideoSource to change the VideoResolution after the NativeVideoSource is created and call that in the callback. Or better yet, we could add a method to VideoTrackSource that takes a DesktopFrame and takes care of this automatically.

webrtc-sys/build/src/lib.rs
iparaskev: I just got lazy here. I will update the scripts for every platform.

examples/screensharing/src/main.rs
iparaskev: I don't understand what is the problem here TBH.

You should know with which resolution you want to share your stream (of course you need to do change the dims to match the aspect ration of the source to avoid distortions), which of course most of the times will be different from the capture res. 

In platforms that allow you to choose your source without a system picker you should be able to get the source's dims before starting capturing. When you are using the system picker and you can't know which source the user selected you can simply start publishing your track after capturing has started. 

IMO the only improvement that could be done here is to change `VideoTrackSource::InternalSource::on_captured_frame` to scale the buffer to match `resolution_` (with the aspect ratio change).

Be-ing: > You should know with which resolution you want to share your stream

How can you know that without any idea what the resolution of the source is? There's no way to know in advance if you're capturing a 200 x 200 window or a 7680 × 4320 screen. Up/down scaling either of these to some arbitrary assumed sized is not going to look good.

Be-ing: For context, I started looking into this because the current screen capturing code in Zed assumes the resolution is known in advance and sets the stream dimensions based on that. That assumption is true for the current capture mechanisms (the scap crate and macOS APIs), but not for libwebrtc.

Be-ing: Digging more into the code, I'm getting more confused. I don't understand the point of calling AdaptedVideoTrackSource::AdaptFrame in VideoTrackSource::InternalSource::on_captured_frame because the AdaptedVideoTrackSource's VideoAdapter is never really configured with a VideoSinkWants... there's just this code with a default constructed VideoSinkWants:

```c++
void VideoTrack::add_sink(const std::shared_ptr<NativeVideoSink>& sink) const {
  webrtc::MutexLock lock(&mutex_);
  track()->AddOrUpdateSink(sink.get(),
                           webrtc::VideoSinkWants());  // TODO(theomonnom): Expose
                                                    // VideoSinkWants to Rust?
  sinks_.push_back(sink);
}
```

When the AdaptFrame call does change the resolution, it always outputs a square resolution (height and width equal) regardless of the input resolution.



Another confusing bit of code I found is this in libwebrtc::NativeVideoSource::new

```rust
        livekit_runtime::spawn({
            let source = source.clone();
            let i420 = I420Buffer::new(resolution.width, resolution.height);
            async move {
                let mut interval = interval(Duration::from_millis(100)); // 10 fps

                loop {
                    interval.tick().await;

                    let inner = source.inner.lock();
                    if inner.captured_frames > 0 {
                        break;
                    }

                    let mut builder = vf_sys::ffi::new_video_frame_builder();
                    builder.pin_mut().set_rotation(VideoRotation::VideoRotation0);
                    builder.pin_mut().set_video_frame_buffer(i420.as_ref().sys_handle());

                    let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
                    builder.pin_mut().set_timestamp_us(now.as_micros() as i64);

                    source.sys_handle.on_captured_frame(&builder.pin_mut().build());
                }
            }
        });
```

With desktop capturing, we're calling `on_captured_frame` when the DesktopCapturer's callback is invoked, so it doesn't make sense to me to have livekit_runtime running this loop. And I definitely want more than 10 FPS for screen captures. I also don't understand why livekit_runtime is a dependency of the libwebrtc crate.

I found that code because I tried setting the VideoResolution for NativeVideoSource::new to 0,0 to test this code in VideoTrackSource::InternalSource::on_captured_frame:

```c++
  if (resolution_.height == 0 || resolution_.width == 0) {
    resolution_ = VideoResolution{static_cast<uint32_t>(buffer->width()),
                                  static_cast<uint32_t>(buffer->height())};
  }
```

However, I have to comment out the call to `livekit_runtime::spawn` above because I420Buffer::new asserts that the width and height aren't 0, so this code I pasted from VideoTrackSource::InternalSource::on_captured_frame is effectively unusable.

@theomonnom as the author of much of this code, can you explain what's going on here? What is really the purpose of the VideoResolution passed to NativeVideoSource::new?

Be-ing: Ah I see, the NativeVideoSource's resolution is read in livekit::LocalParticipant::publish_track. So I think what is needed is a new API to change the resolution of an existing track after it has been published.

Be-ing: > When you are using the system picker and you can't know which source the user selected you can simply start publishing your track after capturing has started.

After stepping away and rereading this, yeah I think that could work. I'll give it a try.

An API for updating the resolution after capturing has started is still needed because windows can be resized during capturing, though I don't think that needs to be done immediately for this to be usable. Of course a screen isn't going to be resized during capturing.

Be-ing: I have refactored the example to publish the track after the resolution is known from the first callback: https://github.com/gethopp/rust-sdks/pull/4. This makes a clearly noticeable difference in text legibility when sharing my 3840 x 2160 screen without downscaling it to the arbitrary 1920 x 1080 resolution that was used before. Sharing a window and resizing it works well enough. A window that is resized to smaller than its initial size looks fine; a window that is resized to bigger than its initial size looks like it gets downscaled to the initial size.

Be-ing:
I am finally using this downstream in [a PR for Zed](https://github.com/zed-industries/zed/pull/42670). It would help move that along to get this merged soon. @iparaskev could you resolve the merge conflicts and review https://github.com/gethopp/rust-sdks/pull/4? If you don't have time to continue working on this anytime soon, I could open a new PR and you could close this one.

iparaskev:
@Be-ing I was busy the previous weeks but I have some time this weekend to update the PR. Also thanks for the PR for my fork, I will have a look and keep what I think it makes sense to be kept. Feel free to leave comments here so the maintainers can also chip in.

Be-ing:
Note that https://github.com/livekit/rust-sdks/pull/784 was just merged yesterday to get the next release to build libwebrtc with the required gn arguments regardless of when this PR gets merged.

iparaskev:
I have only tested it on macos, will check windows and ubuntu tomorrow. Also about the track dims for windows that have been resized, if I remember correctly in other sdks LiveKit allows this so we should be able to add in this sdk as well.

Be-ing:
> about the track dims for windows that have been resized, if I remember correctly in other sdks LiveKit allows this so we should be able to add in this sdk as well.

Yes, in the code generated from the protobuf, there is code for it, but that generated code is never called in the Rust APIs.

Be-ing:
Can you explain why you didn't merge some code from https://github.com/gethopp/rust-sdks/pull/4? https://github.com/gethopp/rust-sdks/pull/4/commits/b3ef1e07d55cb153583476081147638540786db4 is helpful so the callback can have its own mutable state without needing to share state with another thread and thus having to wrap everything in `Arc<Mutex>` and deal with cloning & locking. https://github.com/gethopp/rust-sdks/pull/4/commits/22c39f2dcf30fd61bf3242578e404ee1077ac59a is helpful because a real application (Zed, for example) is likely to instantiate a DesktopCapturer somewhere in UI code to enumerate the available CaptureSources before starting capturing. This is evident in the changes to example in https://github.com/gethopp/rust-sdks/pull/4/commits/22c39f2dcf30fd61bf3242578e404ee1077ac59a, but I don't know why you didn't merge the TUI for selecting the source in the example.

cloudwebrtc:
cool, works fine on macOS.

webrtc-sys/libwebrtc/build_linux.sh
```diff
@@ -137,6 +137,7 @@ python3 "./src/tools_webrtc/libs/generate_licenses.py" \
   --target :default "$OUTPUT_DIR" "$OUTPUT_DIR"
 
 cp "$OUTPUT_DIR/obj/webrtc.ninja" "$ARTIFACTS_DIR"
+cp "$OUTPUT_DIR/obj/modules/desktop_capture/desktop_capture.ninja" "$ARTIFACTS_DIR"
```
Be-ing, line 140: @cloudwebrtc I forgot in https://github.com/livekit/rust-sdks/pull/784 that this change to the libwebrtc build script is also needed in addition to the change for the gn arguments. I could make another PR for just this little change and you could tag a new webrtc release, or we could merge this PR now and then deal with fixing the CI by bumping the tag after. How do you want to proceed?

By the way, the tag that was just pushed the other day confusingly has the same short commit hash as the last tag but has `-3` at the end instead of an updated hash.

Be-ing: > I could make another PR for just this little change and you could tag a new webrtc release,

PR is at https://github.com/livekit/rust-sdks/pull/787

cloudwebrtc: Hey, thanks for reminding me. Yes, we can update to the new hash 0001d84.

libwebrtc/src/desktop_capturer.rs
```diff
@@ -0,0 +1,227 @@
...
+///
+/// It contains a subset of libwertc's DesktopCaptureOptions.
+///
+/// By default, it captures the entire screen and does not include the cursor.
+///
+/// # Example
+/// ```no_run
+/// use libwebrtc::desktop_capturer::{DesktopCapturerOptions, DesktopCaptureSourceType};
+///
+/// let mut options = DesktopCapturerOptions::new(DesktopCaptureSourceType::SCREEN);
```
xianshijing-lk: nit, it should be DesktopCaptureSourceType::Screen

libwebrtc/src/desktop_capturer.rs
```diff
@@ -0,0 +1,227 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::imp::desktop_capturer as imp_dc;
+
+/// Configuration options for creating a desktop capturer.
+///
+/// It contains a subset of libwertc's DesktopCaptureOptions.
```
xianshijing-lk: nit, libwertc -> libwebrtc

libwebrtc/src/desktop_capturer.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub fn left(&self) -> i32 {
+        self.sys_handle.left()
+    }
+
+    pub fn top(&self) -> i32 {
+        self.sys_handle.top()
+    }
+
+    pub fn data(&self) -> &[u8] {
+        &self.sys_handle.data()
```
xianshijing-lk: should it be self.sys_handle.data() instead ? rather than &self.sys_handle.data()

libwebrtc/src/native/desktop_capturer.rs
```diff
@@ -0,0 +1,214 @@
...
+            include_cursor: false,
+            #[cfg(target_os = "macos")]
+            allow_sck_system_picker: true,
+        }
+    }
+}
+
+impl DesktopCapturerOptions {
+    pub(crate) fn new(source_type: SourceType) -> Self {
+        Self { source_type, include_cursor: false, ..Default::default() }
```
xianshijing-lk: I think default() already set the include_cursor to false, simply do 
Self { source_type, ..Default::default() }

webrtc-sys/include/livekit/desktop_capturer.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "webrtc-sys/src/desktop_capturer.rs.h"
+
+namespace livekit {
+
+class DesktopCapturer : public webrtc::DesktopCapturer::Callback {
+ public:
+  explicit DesktopCapturer(std::unique_ptr<webrtc::DesktopCapturer> capturer)
+      : capturer(std::move(capturer)),
+        callback(rust::Box<DesktopCapturerCallbackWrapper>::from_raw(nullptr)) {};
```
xianshijing-lk: I think this is a concerning, is it Ok to use nullptr with rust::Box<T>::from_raw() ? which takes a pointer allocated from rust to C++ ? 

@ladvoc for suggestion here. 

Alternatively, we can change callback to optional in C++
std::optional<rust::Box<DesktopCapturerCallbackWrapper>> callback;

 then we can remove this code 
'callback(rust::Box<DesktopCapturerCallbackWrapper>::from_raw(nullptr))'

webrtc-sys/include/livekit/desktop_capturer.h
```diff
@@ -0,0 +1,74 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <memory>
+
+#include "modules/desktop_capture/desktop_capture_options.h"
```
xianshijing-lk: can you forward declaring the following type, then move these #include  desktop_capture_options.h / desktop_capturer.h to the .cpp implementation? 

namespace webrtc {
class DesktopFrame;
class DesktopCapturer;
}

webrtc-sys/include/livekit/desktop_capturer.h
```diff
@@ -0,0 +1,74 @@
...
+  void capture_frame() const { capturer->CaptureFrame(); };
+
+ private:
+  std::unique_ptr<webrtc::DesktopCapturer> capturer;
+  rust::Box<DesktopCapturerCallbackWrapper> callback;
+};
+
+class DesktopFrame {
+ public:
+  DesktopFrame(std::unique_ptr<webrtc::DesktopFrame> frame) : frame(std::move(frame)) {};
```
xianshijing-lk: Remove the ; at the end

webrtc-sys/src/desktop_capturer.cpp
```diff
@@ -0,0 +1,107 @@
...
+    case webrtc::DesktopCapturer::Result::ERROR_PERMANENT:
+      ret_result = CaptureResult::ErrorPermanent;
+      break;
+    case webrtc::DesktopCapturer::Result::ERROR_TEMPORARY:
+      ret_result = CaptureResult::ErrorTemporary;
+      break;
+    default:
+      break;
+  }
+  callback->on_capture_result(ret_result,
```
xianshijing-lk: The current code will always create a DesktopFrame and call on_capture_result regardless if it is success or not. 

Is there any reason for doing so ? 

Maybe you should do 
std::unique_ptr<DesktopFrame> desktopFrame;

switch (result) {
    case webrtc::DesktopCapturer::Result::SUCCESS:
      ret_result = CaptureResult::Success;
      std::make_unique<DesktopFrame>(std::move(frame);
    case ...
      // Don't create the DesktopFrame
      
  
callback->on_capture_result(ret_result, desktopFrame);

webrtc-sys/src/desktop_capturer.cpp
```diff
@@ -0,0 +1,107 @@
...
+void DesktopCapturer::start(
+    rust::Box<DesktopCapturerCallbackWrapper> callback) {
+  this->callback = std::move(callback);
+  capturer->Start(this);
+}
+
+void DesktopCapturer::OnCaptureResult(
+    webrtc::DesktopCapturer::Result result,
+    std::unique_ptr<webrtc::DesktopFrame> frame) {
+  CaptureResult ret_result = CaptureResult::Success;
```
xianshijing-lk: in the default section, you don't change the ret_result, so you should initialize it to an error as default, like 
CaptureResult::ErrorPermanent

webrtc-sys/src/desktop_capturer.cpp
```diff
@@ -0,0 +1,107 @@
...
+    DesktopCapturerOptions options) {
+  webrtc::DesktopCaptureOptions webrtc_options =
+      webrtc::DesktopCaptureOptions::CreateDefault();
+#if defined(WEBRTC_MAC) && !defined(WEBRTC_IOS)
+  webrtc_options.set_allow_sck_capturer(true);
+  webrtc_options.set_allow_sck_system_picker(options.allow_sck_system_picker);
+#endif /* defined(WEBRTC_MAC) && !defined(WEBRTC_IOS) */
+#ifdef _WIN64
+  if (options.window_capturer) {
+    webrtc_options.set_allow_wgc_screen_capturer(true);
```
xianshijing-lk: this looks weird.

it set set_allow_wgc_screen_capturer when window_capturer is true

and set set_allow_wgc_window_capturer otherwise. 

I think it should be opposite

webrtc-sys/src/desktop_capturer.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // https://github.com/webrtc-sdk/webrtc/blob/m137_release/modules/desktop_capture/desktop_capture_options.h#L133-L142
+    webrtc_options.set_enumerate_current_process_windows(false);
+  }
+  webrtc_options.set_allow_directx_capturer(true);
+#endif /* _WIN64 */
+#ifdef WEBRTC_USE_PIPEWIRE
+  webrtc_options.set_allow_pipewire(true);
+#endif /* WEBRTC_USE_PIPEWIRE */
+
+  webrtc_options.set_prefer_cursor_embedded(options.include_cursor);
```
xianshijing-lk, line 54: if you know what this set_prefer_cursor_embedded mean, and why it gets mapped to include_cursor here

please add a comment. I don't really understand this settings from the names :)

webrtc-sys/src/desktop_capturer.cpp
xianshijing-lk: Add a sanity check at the top of the function 
if (!callback) {
  return;
}

webrtc-sys/src/desktop_capturer.rs
```diff
@@ -0,0 +1,98 @@
...
+            result: CaptureResult,
+            frame: UniquePtr<DesktopFrame>,
+        );
+    }
+}
+
+impl_thread_safety!(ffi::DesktopCapturer, Send + Sync);
+
+pub trait DesktopCapturerCallback: Send {
+    fn on_capture_result(&mut self, result: CaptureResult, frame: UniquePtr<DesktopFrame>);
```
xianshijing-lk: Note that DesktopFrame can be null, should we use 
frame: Option<DesktopFrame>
instead ?

libwebrtc/src/native/desktop_capturer.rs
```diff
@@ -0,0 +1,214 @@
...
+impl<T> sys_dc::DesktopCapturerCallback for DesktopCallback<T>
+where
+    T: FnMut(CaptureResult, DesktopFrame) + Send,
+{
+    fn on_capture_result(
+        &mut self,
+        result: sys_dc::ffi::CaptureResult,
+        frame: UniquePtr<sys_dc::ffi::DesktopFrame>,
+    ) {
+        (self.callback)(capture_result_from_sys(result), DesktopFrame::new(frame));
```
xianshijing-lk: + @ladvoc  for suggestion here. 

As errors can happen, which will result in an nullptr frame, we should handle the case better

I wonder if we should do things like 
let result = capture_result_from_sys(result);
let frame = if frame.is_null() {
    None
} else {
    Some(DesktopFrame::new(frame))
};
(self.callback)(result, frame);

xianshijing-lk (commented):
Great work, I have some comments.

webrtc-sys/libwebrtc/build_linux.sh
Be-ing: @iparaskev I have updated the hash for the prebuilt library in https://github.com/gethopp/rust-sdks/pull/5. Let's merge that and see if CI works on this branch.

webrtc-sys/include/livekit/desktop_capturer.h
Be-ing: The Rust API guarantees that the `callback` member is set to a real pointer before it is used by requiring it as a parameter for DesktopCapturer::start.

webrtc-sys/src/desktop_capturer.cpp
Be-ing: I don't see why the Rust callback shouldn't have access to those errors. The application developer may want to do something with them.

webrtc-sys/include/livekit/desktop_capturer.h
xianshijing-lk: I got the usage of it. 

My suggestion is removing the need of doing "callback(rust::Box<DesktopCapturerCallbackWrapper>::from_raw(nullptr))"

one way is to use std::optional<rust::Box> callback;
then we will set the callback in the Start() function.

Alternative, we will make sure the callback is passed in the constructor, like
```
DesktopCapturer(std::unique_ptr<webrtc::DesktopCapturer> capturer,
                  rust::Box<DesktopCapturerCallbackWrapper> callback)
      : capturer(std::move(capturer)),
        callback(std::move(callback)) {}

void start() {
    capturer->Start(this);
}
```


On the Rust side:
```
fn new_desktop_capturer(
    options: DesktopCapturerOptions,
    callback: Box<DesktopCapturerCallbackWrapper>,
) -> UniquePtr<DesktopCapturer>;

fn start(self: Pin<&mut DesktopCapturer>);
```

webrtc-sys/src/desktop_capturer.cpp
xianshijing-lk: Sorry for the confusion. My comment was mainly that we shouldn't create a new desktopFrame from a nullptr

webrtc-sys/include/livekit/desktop_capturer.h
Be-ing: Using a std::optional would work, sure.

The callback was originally passed in the constructor. I changed that because it is inconvenient for the Rust application. A real application likely needs some UI to let the user pick from the list of available CaptureSources. This UI code would instantiate a DesktopCapturer. At that point there is no need to provide a callback. Calling DesktopCapturer::start is likely done in a different part of the application's code from the UI.

libwebrtc/src/native/desktop_capturer.rs
```diff
@@ -0,0 +1,214 @@
...
+        sources
+    }
+}
+
+pub struct DesktopFrame {
+    pub(crate) sys_handle: UniquePtr<sys_dc::ffi::DesktopFrame>,
+}
+
+impl DesktopFrame {
+    pub fn new(sys_handle: UniquePtr<sys_dc::ffi::DesktopFrame>) -> Self {
```
ladvoc: Does this ever get constructed outside the _libwebrtc_ crate? If not, it should be marked `pub(crate)`.

libwebrtc/src/native/desktop_capturer.rs
ladvoc: +1. There is currently UB when a slice is formed with `DesktopFrame::data()` when the pointer is null. We can either do a simple null check here or in `DesktopFrame::new()`—I am ok with either.

libwebrtc/src/desktop_capturer.rs
ladvoc: +1, this adds a reference to a temporary that immediately gets dropped.

ladvoc (commented):
Overall looks great, just a few small comments

libwebrtc/src/native/desktop_capturer.rs
Be-ing: How about instead of passing an Option to the Rust callback, we pass a single argument of `Result<DesktopFrame, CaptureError>`?

webrtc-sys/include/livekit/desktop_capturer.h
iparaskev: I moved `desktop_capture_options.h` but we can't really move `desktop_capturer.h` because we are need `webrtc::DesktopCapturer::Callback` and AFAIU there isn't a way to forward declare a nested class. Let me know if I am missing something.

webrtc-sys/src/desktop_capturer.cpp
iparaskev: Good catch. Actually this doesn't compile. It was a leftover from a previous revision and I didn't get the chance to test the latest change on windows.

webrtc-sys/src/desktop_capturer.cpp
iparaskev: Done. See [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/desktop_capture/desktop_capture_options.h;l=122?q=set_prefer_cursor_embedded&ss=chromium%2Fchromium%2Fsrc) for reference.

libwebrtc/src/native/desktop_capturer.rs
iparaskev: Good catch.

libwebrtc/src/native/desktop_capturer.rs
iparaskev: I liked the `Result` approach and went with it.

iparaskev:
Thanks for the comments everyone. I pushed a new revision with the changes (I still haven't test this on windows).

xianshijing-lk (approved):
lgtm, thanks

xianshijing-lk:
Lets go for it. Thanks for the great work!

PR #727, opened 2025-10-05
fix: do not log 'signal client closed: "stream closed"' on disconnect

davidzhao:
when server sends an explicit LeaveRequest, we would still end up logging

```
livekit::rtc_engine:453:livekit::rtc_engine - received session close: "signal client closed: \"stream closed\"" UnknownReason Resume
```

that's due to the signal client being closed after an explicit Leave is sent

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -70,17 +70,6 @@ jobs:
         with:
           submodules: true
 
-      - name: Cache apt packages
-        if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-24.04-arm' }}
-        uses: actions/cache@v4
-        with:
-          path: |
-            /var/lib/apt/lists
-            /var/cache/apt/archives
-          key: ${{ matrix.os }}-apt-${{ hashFiles('**/builds.yml') }}
-          restore-keys: |
-            ${{ matrix.os }}-apt-
-
       - name: Install linux dependencies
         if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-24.04-arm'}}
         run: |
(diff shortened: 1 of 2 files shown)


ladvoc (approved):
LGTM ✅

ladvoc:
Actually, I just realized this change might have broken `SimulateScenario::SignalReconnect`—it looks like the E2E DC reliability test is hanging.

davidzhao:
oh that's not good. how does the test tear down signal connection?

ladvoc:
The [test](https://github.com/livekit/rust-sdks/blob/main/livekit/tests/data_channel_test.rs) calls `room.simulate_scenario(SimulateScenario::SignalReconnect).await` on both the sending and receiving rooms (after a short delay) to ensure all packets are received despite the connection lapse. With this change, it appears the connection is never reestablished.

livekit/src/rtc_engine/mod.rs
```diff
@@ -450,6 +450,13 @@ impl EngineInner {
                 match action {
                     proto::leave_request::Action::Resume
                     | proto::leave_request::Action::Reconnect => {
+                        let running_handle = self.running_handle.read();
```
ladvoc: It appears reading from the `RwLock` on this line leads to a deadlock—I will have to dig deeper to determine why this is the case.

ladvoc (commented):
See comment

livekit/src/rtc_engine/mod.rs
davidzhao: I see the issue, the read lock wasn't released in time. fixed in https://github.com/livekit/rust-sdks/pull/727/commits/701962c4f7dae00b824ba15026bda07e91607860

PR #728, opened 2025-10-07
Access token via auth header

ladvoc:
Pass access token via auth header (bearer) instead of URL parameter.

### DIFF
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -253,7 +253,7 @@ impl SignalInner {
         let (stream, mut events) = match SignalStream::connect(lk_url.clone(), token).await {
             Ok(stream) => stream,
             Err(err) => {
-                if let SignalError::TokenFormat = err  {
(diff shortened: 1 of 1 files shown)


livekit-api/src/signal_client/signal_stream.rs
```diff
@@ -83,28 +86,13 @@ impl SignalStream {
...
-                    query_pairs.append_pair(&key, &value);
-                }
-
-                query_pairs.append_pair("access_token", "...");
-            }
-
-            log::info!("connecting to {}", url);
-        }
+        log::info!("connecting to {}", url);
+        let mut request = url.clone().into_client_request()?;
```
xianshijing-lk, line 93: nit, do we need to keep the original url ? or we can just directly modify it as 
let mut request = url.into_client_request()?;

I am OK with either

ladvoc: We need to clone in this case since the URL is needed later in the function and `into_client_request` consumes it.

PR #729, opened 2025-10-08
Increase RPC max RT time to 7s

bcherry:
There are some cases where we see ~6s delays. We are separately working to track those down and improve the worst-case speed. The DC reliability work from the summer has addressed the vast majority of true lost packets so it's better to keep waiting than to fail early.

Three changes
- [x] Increase max RT to 7s (it needs to be at least 6s, this felt like a conservative choice for now)
- [x] Increase default response timeout by 15s (keeps the effective handler time to 8s as it was already)
- [x] Add missing response timeout adjustment to account for RT time (other SDKs had this already)

### DIFF
--- a/livekit/src/room/participant/local_participant.rs
+++ b/livekit/src/room/participant/local_participant.rs
@@ -738,7 +738,7 @@ impl LocalParticipant {
     pub async fn perform_rpc(&self, data: PerformRpcData) -> Result<String, RpcError> {
         // Maximum amount of time it should ever take for an RPC request to reach the destination, and the ACK to come back
         // This is set to 7 seconds to account for various relay timeouts and retries in LiveKit Cloud that occur in rare cases
-    
+
         let max_round_trip_latency = Duration::from_millis(7000);
 
         if data.payload.len() > MAX_PAYLOAD_BYTES {


livekit/src/room/participant/local_participant.rs
```diff
@@ -736,7 +736,10 @@ impl LocalParticipant {
     }
 
     pub async fn perform_rpc(&self, data: PerformRpcData) -> Result<String, RpcError> {
-        let max_round_trip_latency = Duration::from_millis(2000);
+        // Maximum amount of time it should ever take for an RPC reqest to reach the destination, and the ACK to come back
+        // This is set to 7 seconds to account for various relay timeouts and retries in LiveKit Cloud that occur in rare cases
+    
+        let max_round_trip_latency = Duration::from_millis(7000);
```
xianshijing-lk, line 742: I don't have any opinion, should we do either 5 or 10 seconds here ? 7 seems a bit less common.

Maybe we can start with 10s to see if that help our customer ?

livekit/src/room/participant/local_participant.rs
```diff
@@ -736,7 +736,10 @@ impl LocalParticipant {
     }
 
     pub async fn perform_rpc(&self, data: PerformRpcData) -> Result<String, RpcError> {
```
xianshijing-lk, line 738: one quick question, should we overload this perform_rpc function like 
pub async fn perform_rpc_with_timeout(&self, data: PerformRpcData, timeout: Duration) -> Result<String, RpcError> 

So that customers can adjust the timeout to meet their needs ?

bcherry: we could - i think it requires a bit of consideration though. currently the "max rt time" variable is used as the connection timeout but is also used to adjust the response timeout reported to the recipient for their handler

e.g. if you have responseTimeout set to 10s and maxRT set to 2s, then the recipient is told they only have 8s to respond to ensure that their response can arrive back at the caller before it times out. So if a user turned their connection timeout up to 10s without adjusting responseTimeout to match, then  their remote handler would have 0s to respond and would timeout immediately

(although it also looks like the rusk SDK is missing this detail - i'll add it in. and I show bump our default response timeout up to match)

that's also another reason to be more conservative here. I picked 7 because the data suggests when things are operating normally 2s is a good number, but we've also identified at least one case that can add a fixed 5s delay. and I didn't want to round up to 10...

livekit/src/room/participant/local_participant.rs
bcherry: see other comment!

livekit/src/room/participant/local_participant.rs
```diff
@@ -736,7 +736,10 @@ impl LocalParticipant {
...
-        let max_round_trip_latency = Duration::from_millis(2000);
+        // Maximum amount of time it should ever take for an RPC reqest to reach the destination, and the ACK to come back
```
boks1971: :nit: request spelling

livekit/src/room/participant/local_participant.rs
bcherry: @xianshijing-lk lmk if you think this makes sense or not - i'll get this PR wrapped up today

xianshijing-lk: I see.  just saw the let effective_timeout = data.response_timeout - max_round_trip_latency below. 

lgtm, thanks for the details.

PR #731, opened 2025-10-09
put examples in root Cargo workspace

Be-ing:
instead of their own workspace. This avoids rebuilding dependencies and wasting space with multiple target directories.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -102,4 +102,5 @@ jobs:
         run: |
           ln -sf $ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/lib/${{ matrix.ndk_arch }}/{libunwind.so,libc++abi.a} $ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/lib/
           cargo install cargo-ndk
-          cargo ndk --target ${{ matrix.target }} build --release -p livekit --workspace -vv
+          # FIXME: wgpu_room does not link on aarch64
+          cargo ndk --target ${{ matrix.target }} build --release -p livekit --workspace --exclude wgpu_room -vv


.github/workflows/tests.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
           sudo apt update -y
-          sudo apt install -y libssl-dev libx11-dev libgl1-mesa-dev libxext-dev libva-dev libdrm-dev libnvidia-decode-570 libnvidia-compute-570 nvidia-cuda-dev
+          sudo apt install -y libasound2-dev libssl-dev libx11-dev libgl1-mesa-dev libxext-dev libva-dev libdrm-dev libnvidia-decode-570 libnvidia-compute-570 nvidia-cuda-dev
```
ladvoc, line 86: nit: Maybe we should add a comment here indicating _libasound2-dev_ is for the _local_audio_ example?

examples/basic_room/src/main.rs
```diff
@@ -5,8 +5,7 @@ use std::env;
 // Connect to a room using the specified env variables
 // and print all incoming events
 
-#[tokio::main]
```
ladvoc, line 8: Thank you for demonstrating the use of alternative executors! One thought: since the _basic_room_ example serves as a starting point for many developers, I'm thinking it might be better to keep it as simple as possible and stick with just Tokio. Perhaps we could turn this into its own example?

examples/send_bytes/Cargo.toml
```diff
@@ -1,15 +1,13 @@
 [package]
 name = "send_bytes"
 version = "0.1.0"
-edition = "2024"
```
ladvoc, line 4: Good catch. Yeah, it doesn't make sense to just use 2024 for this example—we should update everything at once later on.

Cargo.toml
ladvoc, line 1: When building from the workspace root, I get a few warnings:
```
warning: /rust-sdks/livekit/Cargo.toml: `default-features` is ignored for livekit-api, since `default-features` was not specified for `workspace.dependencies.livekit-api`, this could become a hard error in the future
warning: /rust-sdks/examples/basic_room/Cargo.toml: `default-features` is ignored for livekit, since `default-features` was not specified for `workspace.dependencies.livekit`, this could become a hard error in the future
warning: /rust-sdks/examples/basic_room/Cargo.toml: `default-features` is ignored for livekit-api, since `default-features` was not specified for `workspace.dependencies.livekit-api`, this could become a hard error in the future
```

examples/basic_room/src/main.rs
Be-ing: The alternative executors were their own examples. That's incompatible with putting them in the workspace though because they require incompatible Cargo features for livekit-runtime.

The async and dispatcher features don't actually build currently due to a version mismatch of the http crate. They didn't build before either. I think it would be fine to  remove those from the example and only demonstrate with tokio. There's not much different as far as Livekit is concerned between the executors; using different executors should be pretty simple just from looking at the upstream documentation for that executor.

examples/send_bytes/Cargo.toml
Be-ing: CI caught this because it's using some old Rust nightly that doesn't support Rust 2024. It would be good to use `edition.workspace = true` across all crates.

Cargo.toml
Be-ing: fixed

.github/workflows/tests.yml
Be-ing: done

examples/basic_room/src/main.rs
ladvoc: Agreed, thanks for making the change. As for the async and dispatcher features not compiling, we will need to address that in a separate PR (along with testing in CI).

Be-ing:
I got Android builds to work again. The wgpu_room example fails to link on aarch64 Android with a strange error:

```
  = note: ld.lld: error: undefined symbol: __arm_tpidr2_save
          >>> referenced by rotate_sme.cc
          >>>               rotate_sme.o:(TransposeWxH_SME) in archive /home/runner/work/rust-sdks/rust-sdks/target/aarch64-linux-android/release/deps/libwebrtc_sys-8a55d18259959b0a.rlib
```

I don't know what to make of that error; it might take a deep dive into how libwebrtc.a is built for Android (which I don't plan to do) to resolve. I think just skipping building that one crate is fine for now.

Phew, that was a hairy yak to shave :razor:

ladvoc (approved):
LGTM ✅

PR #730, opened 2025-10-08
build_linux.sh: add option to choose toolchain

Be-ing:
This adds a --toolchain CLI option to the script that can be either
  * gnu
  * llvm
  * chromium-llvm

The gnu and llvm options use the system's toolchain. chromium-llvm uses Chromium's prebuilt LLVM toolchain with its Debian sysroot.

Previously, the script tried to combine the system's gcc with libstdc++ from Chromium's Debian sysroot, which failed to find glibc headers:

```
[...]

PR #733, opened 2025-10-10
fix: fix linux so link issue.

cloudwebrtc:
Since rustc is not compatible with libwebrtc.a compiled from llvm clang, we must use gcc/g++ 14 to compile Rust-compatible static libraries. However, g++ 14 will force the GLIBC_USE_ISOC23 macro to be enabled, which forces linking to isoc23_xxx symbols. This causes some older Docker images (which do not support glibc_2.38) to fail to load ffi.so. This PR is used to fix this issue.

### DIFF
--- a/webrtc-sys/libwebrtc/build_linux.sh
+++ b/webrtc-sys/libwebrtc/build_linux.sh
@@ -88,7 +88,7 @@ mkdir -p "$ARTIFACTS_DIR/lib"
 python3 "./src/build/linux/sysroot_scripts/install-sysroot.py" --arch="$arch"
 
 if [ "$arch" = "arm64" ]; then
-  sed -i 's/__GLIBC_USE_ISOC2X[[:space:]]*1/__GLIBC_USE_ISOC2X\t0/' /usr/aarch64-linux-gnu/include/features.h
+  sudo sed -i 's/__GLIBC_USE_ISOC2X[[:space:]]*1/__GLIBC_USE_ISOC2X\t0/' /usr/aarch64-linux-gnu/include/features.h
 fi
 
 debug="false"


webrtc-sys/libwebrtc/build_linux.sh
```diff
@@ -75,13 +75,22 @@ cd build
...
+git apply "$COMMAND_DIR/patches/david_disable_gun_source_macro.patch" -v --ignore-space-change --ignore-whitespace --whitespace=nowarn
+
+cd ../..
 
 mkdir -p "$ARTIFACTS_DIR/lib"
 
 python3 "./src/build/linux/sysroot_scripts/install-sysroot.py" --arch="$arch"
 
+if [ "$arch" = "arm64" ]; then
+  sudo sed -i 's/__GLIBC_USE_ISOC2X[[:space:]]*1/__GLIBC_USE_ISOC2X\t0/' /usr/aarch64-linux-gnu/include/features.h
```
Be-ing, line 91: Needing sudo to run this script is going to be cumbersome. Is there any way around that?

cloudwebrtc: At first I tried to use `__DEFAULT_SOURCE` instead of `__GNU_SOURCE` macro to turn off `ISOC23`, but this is only valid for `x64` platform. and later I used diff to create a patch, but it also requires `sudo`, so I guess there is no better way at present.

Be-ing: A build script shouldn't modify system headers. It would be more appropriate to move this `sed` command to the GitHub Actions workflow that builds libwebrtc.a.

Be-ing: I took care of this in #730.

xianshijing-lk: I was about to suggest but deleted the draft since I didn't want to block the fix for another day. 

But curiously, have you tried setting some compilation flags in the BUILD.gn file ? like -D__GLIBC_USE_ISOC2X=0 and or -U__GLIBC_USE_ISOC2X ? 

or some other build file, will that work ?

PR #738, opened 2025-10-11
Derive from_variants for FFI oneof fields

ladvoc:
This PR derives [_from_variants_](https://docs.rs/from_variants/latest/from_variants/) on generated enums representing oneof fields from the FFI protocol, allowing a message to be converted into its corresponding oneof variant using *into*. This results in a significant boilerplate reduction in the FFI implementation.


ladvoc:
> Makes sense to me!
> 
> Only one minor thought - how common are enum field that contain the same struct, resulting in conflicts in the `from_variants` implementations? Is it worth trying to think about some sort of more scalable solution than just piecemeal adding `from_variants_skip_field` to handle these or are they fairly uncommon?

Good point! I did think about this, but I think this situation is pretty uncommon. A more scalable solution would be creating a custom proc_macro, but I think we should cross that bridge later if it becomes an issue.

PR #745, opened 2025-10-15
Cargo workspace cleanup

Be-ing:
follow up to #731


ladvoc (approved):
LGTM ✅, thanks for the follow-up! We added a new example today, _agent_dispatch_. If you could also add `publish = false` to that one, it’d be great so we can keep everything in a single PR.

Be-ing:
done

PR #749, opened 2025-10-15
[WIP] Rust TokenSource implementation

1egoman:
Talking with @ladvoc, it seemed like it would be a good exercise to port the `TokenSource` to rust as a first pass to see what new additional complexity is exposed in a less client focused context. Plus longer term, this could be useful as part of the broader rust client sdk core project.

For a working example, see `examples/token_source`. You'll want to run it with some environment variables populated - something maybe like this would work: `LIVEKIT_URL=1 LIVEKIT_API_KEY=2 LIVEKIT_API_SECRET=3 SERVER_URL=4 API_KEY=5 API_SECRET=6 RUST_LOG=info cargo run`

> [!WARNING]
> General caveat: I've yet to write a substantial amount of rust professionally, though personally I've written a fair bit. Because of this it's fairly likely I'm doing some things badly / in non idiomatic ways. Feel free to point any of that kind of stuff out!

# High level summary
The `literal`, `endpoint`, `sandboxTokenServer`, and `custom` implementations all work very similar to their web, swift, and android counterparts.

The big new concept this change introduces is the `TokenSourceMinter` / `TokenSourceMinterCustom`. Roughly what they look like:
```rust
let fetch_options = TokenSourceFetchOptions::default().with_agent_name("voice ai quickstart");

// A TokenSourceMinter is a configurable token source that lets a user "mint" new tokens with a given set of options:
let minter = TokenSourceMinter::default();
let _ = minter_literal_custom.fetch(&fetch_options).await;

// There's also a mechanism to allow users to pass different sets of credentials to the minter either from literal
// strings (below), environment variables (the default), or via a manual closure evaluated at fetch time.
let minter_literal = TokenSourceMinter::new(MinterCredentials::new("server url", "api key", "api secret"));
let _ = minter_literal_credentials.fetch(&fetch_options).await;

// One other type of TokenSource I added - I'm not sure if this one is useful / a good idea though...
//
// A TokenSourceCustomMinter is a fixed TokenSource which calls its provided closure with a preconfigured
// access token builder which allows you to manually update and return a new generated token. It seems like
// this might be a better approach for one off use cases while a `TokenSourceMinter` would be better for use
// with a future agents sdk which could pass in an agent name / etc.
let custom_minter =
    TokenSourceCustomMinter::new(|access_token| {
        access_token
            .with_identity("rust-bot")
            .with_name("Rust Bot")
            .with_grants(access_token::VideoGrants {
                room_join: true,
                room: "my-room".to_string(),
                ..Default::default()
            })
            .to_jwt()
    });
let _ = custom_minter.fetch().await;
```

# Large remaining things to work through
- [x] Add some sort of caching / jwt validation logic to `TokenSourceConfigurable` or maybe a `TokenSourceCached` composition layer like swift/android or something like that
- [ ] Get async closures working within `TokenSourceCustom`
- [ ] Get async closures working within `TokenSourceLiteral`
- [x] Error handling
- [ ] Tests

Fixes CLT-3116


livekit-protocol/src/livekit.rs
```diff
@@ -78,6 +78,14 @@ pub struct EventMetric {
...
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct MetricsRecordingHeader {
+    #[prost(string, tag="1")]
+    pub room_id: ::prost::alloc::string::String,
+    #[prost(bool, optional, tag="2")]
+    pub enable_user_data_training: ::core::option::Option<bool>,
+}
 //
 // Protocol used to record metrics for a specific session.
```
1egoman, line 82: I had to update the `livekit-protocol` package to the latest in order to get access to `TokenSourceRequest` / `TokenSourceResponse` from [here](https://github.com/livekit/protocol/pull/1224) - that's where the majority of the diff changes are coming from:

Image

Is this something worth doing as a separate pull request merged in as a prerequisite or is it fine to include this broader change in here?

livekit/src/room/participant/mod.rs
```diff
@@ -44,6 +44,7 @@ pub enum ParticipantKind {
     Egress,
     Sip,
     Agent,
+    Connector,
 }
```
1egoman, line 49: Along with the `livekit-protocol` changes, I had to add a new corresponding `Connector` value to this enum to get it to compile again. I don't know what this pertains to / why this was added, so I wanted to call it out just to make sure somebody else more familiar with the state of this repository can verify I did the right thing here.

livekit/src/room/token_source/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{error::Error, future::{ready, Future}, pin::Pin};
+use livekit_api::access_token::{self, AccessTokenError};
+
+// FIXME: is reexporting these from here a good idea?
+pub use livekit_protocol::{TokenSourceRequest, TokenSourceResponse};
+
```
1egoman: I did a quick search and couldn't find an example of a pattern like this where protobuf structs are being exposed externally like this.

Is the convention that users are supposed to import this kind of stuff from `livekit_protocol` in their code? Or are these structs being re-exported somewhere and I just missed it?

livekit/src/room/token_source/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+
+impl<F: Fn() -> TokenSourceResponse> TokenLiteralGenerator for F {
+    // FIXME: allow this to be an async fn!
+    fn apply(&self) -> TokenSourceResponse {
+        self()
+    }
+}
+
```
1egoman, line 58: TODO: this needs to be changed to accept an async function, so you can do something like:
```rust
let _ = TokenSourceLiteral::new(async || {
  Ok(TokenSourceResponse { /* ... */})
});
```

livekit/src/room/token_source/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+            .with_identity("rust-bot")
+            .with_name("Rust Bot")
+            .with_grants(access_token::VideoGrants {
+                room_join: true,
+                room: "my-room".to_string(),
+                ..Default::default()
+            })
+            .to_jwt()?;
+
+        Ok(TokenSourceResponse { server_url, participant_token })
```
1egoman, line 103: TODO: finish this up by wiring all the `options` values through to the corresponding `with_` prefixed builder functions!

livekit/src/room/token_source/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+        ) -> Pin<Box<dyn Future<Output = Result<TokenSourceResponse, Box<dyn Error>>>>>,
+    > TokenSourceConfigurable for TokenSourceCustom<CustomFn>
+{
+    async fn fetch(
+        &self,
+        options: &TokenSourceFetchOptions,
+    ) -> Result<TokenSourceResponse, Box<dyn Error>> {
+        (self.0)(options).await
+    }
+}
```
1egoman, line 249: This `TokenSourceCustom` is a bit of a mess and still isn't really doing what I want. Ideally, I'd like to be able to pass an `async |options| {}` as `CustomFn` but I can't quite get that to work.

The closest I've been able to get is:
```rust
let custom = TokenSourceCustom::new(|_options| {
    Box::pin(future::ready(Ok(TokenSourceResponse {
        server_url: "...".into(),
        participant_token: "... _options should be encoded in here ...".into(),
    })))
});
```

Maybe somebody else has some ideas?

livekit/src/room/token_source/traits.rs
```diff
@@ -0,0 +1,79 @@
...
+
+/// A Fixed TokenSource is a token source that takes no parameters and returns a completely
+/// independently derived value on each fetch() call.
+///
+/// The most common downstream implementer is TokenSourceLiteral.
+pub trait TokenSourceFixed {
+    // FIXME: what should the error type of the result be?
+    fn fetch(&self) -> impl Future<Output = Result<TokenSourceResponse, Box<dyn Error>>>;
+}
+
```
1egoman, line 26: It's unclear to me what the error type in the `Result`s in these traits really should be. Right now I did `Box<dyn Error>` since these can be implemented by users / the error really could be anything, but I suspect it's probably not that ideomatic. Maybe I should be exploring an error library of some sort?

livekit/src/room/token_source/traits.rs
```diff
@@ -0,0 +1,79 @@
...
+/// A helper trait to more easily implement a TokenSourceFixed which not async.
+pub trait TokenSourceFixedSynchronous {
+    // FIXME: what should the error type of the result be?
+    fn fetch_synchronous(&self) -> Result<TokenSourceResponse, Box<dyn Error>>;
+}
```
1egoman, line 30: Wanted to call out: I added `-Synchronous` versions of the token source traits to try to cut down on the boilerplate in non async downstream implementors.

livekit/src/room/participant/mod.rs
1egoman: Ugh, yea, I didn't do something right here and this is what is causing the ci build to break... 😞

livekit-protocol/src/livekit.rs
ladvoc: The issue is we are using an old version of [_prost_](https://github.com/tokio-rs/prost), and when you install the generator locally, you get the latest by default. Options:
1. Install the correct generator: `cargo install protoc-gen-prost@0.3.1 protoc-gen-prost-serde@0.3.1`
2. Take this PR out of draft, and the CI will generate for you using the correct version

I'm going to create a ticket to remind us we need to upgrade to _prost_ 14.x.x.

livekit/src/room/participant/mod.rs
ladvoc: Ah I see the issue. Since you added the _Connector_ variant in the public _ParticipantKind_ enum, you are getting a build error because this isn't being exposed over FFI. Steps to fix:
1. Add the variant in the FFI protocol in _livekit-ffi/protocol/participant.proto_
2. Add a match arm in _livekit-ffi/src/conversion/participant.rs_ (this is the source of the build error)

ladvoc: I am also curious what a connector participant is

livekit/src/room/token_source/fetch_options.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// Options that can be used when fetching new credentials from a TokenSourceConfigurable
+///
+/// Example:
+/// ```rust
+/// let _ = TokenSourceFetchOptions::default().with_agent_name("my agent name");
+/// ```
+#[derive(Clone, Default)]
+pub struct TokenSourceFetchOptions {
+    pub room_name: Option<String>,
```
ladvoc: praise: I like using the builder pattern here! I think we should leverage the builder pattern for new APIs since it gives us more room to make changes in the public API without breaking.

ladvoc: suggestion: Usually, with the builder pattern, the struct fields are private (_AccessToken_ in the `livekit-api` crate is a good reference for this).

livekit/src/room/token_source/mod.rs
ladvoc: suggestion: You bring up a good point! There are some places where we re-export protocol messages in the public API, however, IMO we should be moving away from this approach for new APIs. I think of the protocol as an implementation detail that should be able to be changed without necessarily causing public API changes—if something is added at the protocol level, it isn't necessarily meant to be in the public API. In this case, I think a better solution would be to define our own _TokenSourceRequest_ and _TokenSourceResponse_ structs that implement _From_ for their protocol equivalents.

livekit/src/room/token_source/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+
+impl TokenSourceConfigurable for TokenSourceEndpoint {
+    async fn fetch(&self, options: &TokenSourceFetchOptions) -> Result<TokenSourceResponse, Box<dyn Error>> {
+        let client = reqwest::Client::new();
+
+        // FIXME: What are the best practices around implementing Into on a reference to avoid the
+        // clone?
+        let request_body: TokenSourceRequest = options.clone().into();
```
ladvoc: thought:
- You currently implement `impl Into<TokenSourceRequest> for TokenSourceFetchOptions`
- Try implementing `impl Into<TokenSourceRequest> for &TokenSourceFetchOptions` instead

livekit/src/room/token_source/traits.rs
ladvoc: thought: I need to better understand the range of errors that can occur, but my initial thought is to define a `TokenSourceError` enum where one of the cases is a "catch all." We can use _thiserror_ to get _From_ implementations for free since we already depend on it:
```rust
#[derive(Debug, thiserror::Error)]
pub enum TokenSourceError {
    // TODO: other error cases (maybe network)

    #[error("Other error: {0}")]
    Other(#[from] Box<dyn Error + Send + Sync>),
}
```

livekit/src/room/token_source/traits.rs
ladvoc: nitpick: What do you think about abbreviating _Synchronous_ to _Sync_?

livekit/src/room/participant/mod.rs
1egoman: Ah, I see - I didn't realize there were separate FFI protobuf implementations which is what was tripping me up. Just made that change and was able to do a build of `livekit-ffi` locally, thanks!

livekit/src/room/token_source/traits.rs
1egoman: I think this makes a lot of sense! I'll give it a shot and see how it goes.

As for the explicit error cases - there are likely some explicit network related onesa as you pointed out, but also probably a wrapped version of `AccessTokenError` too, plus maybe errors related to environment variables not being found (right now these are panics, but that may not be the best behavior, I am unsure).

livekit/src/room/token_source/traits.rs
1egoman: I originally had them ending in `Sync`, but my concern was that it would imply some sort of relationship with the `Sync` trait. It sounds like you don't think that's a concern though?

livekit/src/room/token_source/fetch_options.rs
1egoman: Ah, I see. The reason I had left them public was so users could interrogate the `TokenSourceFetchOptions` and determine what it currently contains but thinking about it more, that might not be a real use case. I'll get rid of the `pub`s for now 👍

livekit/src/room/token_source/mod.rs
1egoman: Makes sense! I think what I was hoping for was some way to transparently handle both `impl Into<TokenSourceRequest> for TokenSourceFetchOptions` and `impl Into<TokenSourceRequest> for &TokenSourceFetchOptions` through some sort of implicit dereferencing or something like that, but it might just not be possible.

That being said, now that I'm thinking about it more - because `TokenSourceRequest` contains `String`s, if I'm converting a `& TokenSourceFetchOptions` into a `TokenSourceRequest`, there's going to have to be some sort of cloning involved at some level unless I did something more crazy with a `Cow` / etc in `TokenSourceRequest`.

Because of this, I'm not sure where that `clone` should happen, but I tend to think forcing the user to do it makes it clear that it has to happen verses it happening implicitly.

livekit-protocol/src/livekit.rs
1egoman: I followed the instructions in [here](https://github.com/livekit/rust-sdks/blob/main/livekit-protocol/generate_proto.sh#L17) so I think I did generate these with the correct version, FYI. That being said I also added a note for myself to add some better documentation (maybe add it to the `README.md`?) on how to generate these, it took me longer than it should have for me to self-discover how.

livekit/src/room/token_source/mod.rs
1egoman: Sounds good - migrated to doing what you suggested.

livekit/src/room/token_source/traits.rs
1egoman: Done, this worked out pretty well.

livekit/src/room/participant/mod.rs
1egoman: It looks like CI is now passing everywhere except for android for an unknown reason. Any thoughts on why? It doesn't at least at surface level seem to be related to my change but I'm not 100% sure.

Failing android build logs:
```
error: failed to run custom build command for `openssl-sys v0.9.109`

Caused by:
  process didn't exit successfully: `CARGO=/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo CARGO_CFG_FEATURE='' CARGO_CFG_PANIC=unwind CARGO_CFG_TARGET_ABI='' CARGO_CFG_TARGET_ARCH=aarch64 CARGO_CFG_TARGET_ENDIAN=little CARGO_CFG_TARGET_ENV='' CARGO_CFG_TARGET_FAMILY=unix CARGO_CFG_TARGET_FEATURE=neon CARGO_CFG_TARGET_HAS_ATOMIC=128,16,32,64,8,ptr CARGO_CFG_TARGET_OS=android CARGO_CFG_TARGET_POINTER_WIDTH=64 CARGO_CFG_TARGET_VENDOR=unknown CARGO_CFG_UNIX='' CARGO_ENCODED_RUSTFLAGS='' CARGO_MANIFEST_DIR=/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/openssl-sys-0.9.109 CARGO_MANIFEST_LINKS=openssl CARGO_MANIFEST_PATH=/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/openssl-sys-0.9.109/Cargo.toml CARGO_PKG_AUTHORS='Alex Crichton <alex@alexcrichton.com>:Steven Fackler <sfackler@gmail.com>' CARGO_PKG_DESCRIPTION='FFI bindings to OpenSSL' CARGO_PKG_HOMEPAGE='' CARGO_PKG_LICENSE=MIT CARGO_PKG_LICENSE_FILE='' CARGO_PKG_NAME=openssl-sys CARGO_PKG_README=README.md CARGO_PKG_REPOSITORY='https://github.com/sfackler/rust-openssl' CARGO_PKG_RUST_VERSION=1.63.0 CARGO_PKG_VERSION=0.9.109 CARGO_PKG_VERSION_MAJOR=0 CARGO_PKG_VERSION_MINOR=9 CARGO_PKG_VERSION_PATCH=109 CARGO_PKG_VERSION_PRE='' DEBUG=false HOST=x86_64-unknown-linux-gnu LD_LIBRARY_PATH='/home/runner/work/rust-sdks/rust-sdks/target/release/deps:/home/runner/work/rust-sdks/rust-sdks/target/release:/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib:/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib' NUM_JOBS=4 OPT_LEVEL=3 OUT_DIR=/home/runner/work/rust-sdks/rust-sdks/target/aarch64-linux-android/release/build/openssl-sys-855bcbdb9481fece/out PROFILE=release RUSTC=/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc RUSTC_LINKER=/home/runner/.cargo/bin/cargo-ndk RUSTDOC=/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustdoc TARGET=aarch64-linux-android /home/runner/work/rust-sdks/rust-sdks/target/release/build/openssl-sys-9005c0badb4c225b/build-script-main` (exit status: 101)
  --- stdout
  cargo:rustc-check-cfg=cfg(osslconf, values("OPENSSL_NO_OCB", "OPENSSL_NO_SM4", "OPENSSL_NO_SEED", "OPENSSL_NO_CHACHA", "OPENSSL_NO_CAST", "OPENSSL_NO_IDEA", "OPENSSL_NO_CAMELLIA", "OPENSSL_NO_RC4", "OPENSSL_NO_BF", "OPENSSL_NO_PSK", "OPENSSL_NO_DEPRECATED_3_0", "OPENSSL_NO_SCRYPT", "OPENSSL_NO_SM3", "OPENSSL_NO_RMD160", "OPENSSL_NO_EC2M", "OPENSSL_NO_OCSP", "OPENSSL_NO_CMS", "OPENSSL_NO_COMP", "OPENSSL_NO_SOCK", "OPENSSL_NO_STDIO", "OPENSSL_NO_EC", "OPENSSL_NO_SSL3_METHOD", "OPENSSL_NO_KRB5", "OPENSSL_NO_TLSEXT", "OPENSSL_NO_SRP", "OPENSSL_NO_RFC3779", "OPENSSL_NO_SHA", "OPENSSL_NO_NEXTPROTONEG", "OPENSSL_NO_ENGINE", "OPENSSL_NO_BUF_FREELISTS", "OPENSSL_NO_RC2"))
  cargo:rustc-check-cfg=cfg(openssl)
  cargo:rustc-check-cfg=cfg(libressl)
  cargo:rustc-check-cfg=cfg(boringssl)
  cargo:rustc-check-cfg=cfg(awslc)
  cargo:rustc-check-cfg=cfg(libressl250)
  cargo:rustc-check-cfg=cfg(libressl251)
  cargo:rustc-check-cfg=cfg(libressl252)
  cargo:rustc-check-cfg=cfg(libressl261)
  cargo:rustc-check-cfg=cfg(libressl270)
  cargo:rustc-check-cfg=cfg(libressl271)
  cargo:rustc-check-cfg=cfg(libressl273)
  cargo:rustc-check-cfg=cfg(libressl280)
  cargo:rustc-check-cfg=cfg(libressl281)
  cargo:rustc-check-cfg=cfg(libressl291)
  cargo:rustc-check-cfg=cfg(libressl310)
  cargo:rustc-check-cfg=cfg(libressl321)
  cargo:rustc-check-cfg=cfg(libressl332)
  cargo:rustc-check-cfg=cfg(libressl340)
  cargo:rustc-check-cfg=cfg(libressl350)
  cargo:rustc-check-cfg=cfg(libressl360)
  cargo:rustc-check-cfg=cfg(libressl361)
  cargo:rustc-check-cfg=cfg(libressl370)
  cargo:rustc-check-cfg=cfg(libressl380)
  cargo:rustc-check-cfg=cfg(libressl381)
  cargo:rustc-check-cfg=cfg(libressl382)
  cargo:rustc-check-cfg=cfg(libressl390)
  cargo:rustc-check-cfg=cfg(libressl400)
  cargo:rustc-check-cfg=cfg(libressl410)
  cargo:rustc-check-cfg=cfg(ossl101)
  cargo:rustc-check-cfg=cfg(ossl102)
  cargo:rustc-check-cfg=cfg(ossl102f)
  cargo:rustc-check-cfg=cfg(ossl102h)
  cargo:rustc-check-cfg=cfg(ossl110)
  cargo:rustc-check-cfg=cfg(ossl110f)
  cargo:rustc-check-cfg=cfg(ossl110g)
  cargo:rustc-check-cfg=cfg(ossl110h)
  cargo:rustc-check-cfg=cfg(ossl111)
  cargo:rustc-check-cfg=cfg(ossl111b)
  cargo:rustc-check-cfg=cfg(ossl111c)
  cargo:rustc-check-cfg=cfg(ossl111d)
  cargo:rustc-check-cfg=cfg(ossl300)
  cargo:rustc-check-cfg=cfg(ossl310)
  cargo:rustc-check-cfg=cfg(ossl320)
  cargo:rustc-check-cfg=cfg(ossl330)
  cargo:rustc-check-cfg=cfg(ossl340)
  cargo:rerun-if-env-changed=AARCH64_LINUX_ANDROID_OPENSSL_LIB_DIR
  AARCH64_LINUX_ANDROID_OPENSSL_LIB_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
  OPENSSL_LIB_DIR unset
  cargo:rerun-if-env-changed=AARCH64_LINUX_ANDROID_OPENSSL_INCLUDE_DIR
  AARCH64_LINUX_ANDROID_OPENSSL_INCLUDE_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
  OPENSSL_INCLUDE_DIR unset
  cargo:rerun-if-env-changed=AARCH64_LINUX_ANDROID_OPENSSL_DIR
  AARCH64_LINUX_ANDROID_OPENSSL_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_DIR
  OPENSSL_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_aarch64-linux-android
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_aarch64_linux_android
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG_ALLOW_CROSS
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS
  cargo:rerun-if-env-changed=PKG_CONFIG_aarch64-linux-android
  cargo:rerun-if-env-changed=PKG_CONFIG_aarch64_linux_android
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_aarch64-linux-android
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_aarch64_linux_android
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG_SYSROOT_DIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR


  Could not find openssl via pkg-config:
  pkg-config has not been configured to support cross-compilation.

  Install a sysroot for the target platform and configure it via
  PKG_CONFIG_SYSROOT_DIR and PKG_CONFIG_PATH, or install a
  cross-compiling wrapper for pkg-config and set it via
  PKG_CONFIG environment variable.

  cargo:warning=Could not find directory of OpenSSL installation, and this `-sys` crate cannot proceed without this knowledge. If OpenSSL is installed and this crate had trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the compilation process. See stderr section below for further information.

  --- stderr


  Could not find directory of OpenSSL installation, and this `-sys` crate cannot
  proceed without this knowledge. If OpenSSL is installed and this crate had
  trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
  compilation process.

  Make sure you also have the development packages of openssl installed.
  For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.

  If you're in a situation where you think the directory *should* be found
  automatically, please open a bug at https://github.com/sfackler/rust-openssl
  and include information about your system as well as this message.

  $HOST = x86_64-unknown-linux-gnu
  $TARGET = aarch64-linux-android
  openssl-sys = 0.9.109
```

ladvoc: Yes, this is a build issue not related to this PR—looking into it now.

PR #750, opened 2025-10-16
UniFFI interface

ladvoc:
Introduces an experimental FFI interface built with UniFFI.

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -77,3 +77,9 @@ changelog_include = [
   "livekit-protocol",
   "webrtc-sys-build",
 ]
+
+[[package]]
+name = "livekit-uniffi"
+changelog_path = "livekit-uniffi/CHANGELOG.md"
+publish = false
+release = true


livekit-uniffi/src/log_forward.rs
```diff
@@ -0,0 +1,107 @@
...
+    log::set_max_level(level);
+}
+
+/// Asynchronously receives a forwarded log entry.
+///
+/// Invoke repeatedly to receive log entries as they are produced
+/// until `None` is returned, indicating forwarding has ended. Clients will
+/// likely want to bridge this to the languages's equivalent of an asynchronous stream.
+///
+/// If log forwarding hasn't been bootstrapped, this will panic.
```
pblazej: Everything works perfectly fine, can we just lift this requirement? We had a similar issue with swift-logging - it means that you cannot inject any default level (e.g. info) as subsequent calls (from the user) will crash.

This will be called before _any logging_.

Sth like
```rust
#[uniffi::export]
fn log_forward_bootstrap(level: LevelFilter) {
    INIT.call_once(|| {
        let logger = LOGGER.get_or_init(|| Logger::new());
        log::set_logger(logger).expect("Failed to set logger");
    });
    // Always update the level, even on subsequent calls
    log::set_max_level(level);
}
```

pblazej: Doesn't seem to work internally

ladvoc: This should be possible, let me check how to make it work.

ladvoc: Done. Also, I made the log level in the Python example debug so you can see the logs.

pblazej: Oh seems like ignoring error works, will test that in spare time 🥲

livekit-uniffi/src/access_token.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub admin: bool,
+    pub call: bool,
+}
+
+/// Claims decoded from a valid access token.
+#[derive(uniffi::Record)]
+pub struct Claims {
+    pub exp: u64,
+    pub iss: String,
+    pub nbf: u64,
```
pblazej, line 112: [nitpick] can we use `SystemTime` instead?

pblazej:
I made a manual pre-release here: https://github.com/livekit/livekit-uniffi-xcframework
Swift SDK support: https://github.com/livekit/client-sdk-swift/pull/822

@ladvoc if you're able to finish the release pipeline here, that would be cool (artifact is OK as mentioned above).

livekit-uniffi/Makefile.toml
```diff
@@ -0,0 +1,206 @@
...
+cargo swift package \
+--accept-all \
+--name ${SPM_NAME} \
+--xcframework-name Rust${SPM_NAME} \
+--platforms macos \
+--platforms ios \
+--platforms maccatalyst \
+--platforms visionos \
+--platforms tvos \
+--lib-type dynamic \
```
pblazej: And one more thing @ladvoc static release is still ~40MB whereas dylib ~1.8MB.

Dynamic one won't pass AppStore validation (we'd need to wrap it in a Framework).

> 90426: Invalid Swift Support. The SwiftSupport folder is missing. Rebuild your app using the current public (GM) version of Xcode and resubmit it.

I'd rather go for stripping the static one further, any ideas?

pblazej: iOS in `--release` profile:

> Ar Archive - 47 MB

pblazej: Okay, looks like `DEAD_CODE_STRIPPING` does the job and the bundle size increase is negligible (0.1 MB), moved back to `static` https://github.com/livekit/rust-sdks/pull/750/commits/729bc101621c37ab8d98671203c3adefd3d1b777

It also passes 🍏 validation 👍 

I'm still open to optimization ideas 😄

livekit-uniffi/Makefile.toml
```diff
@@ -0,0 +1,305 @@
...
+tera  --env --file ${SUPPORT_DIR}/${LANG}/Package@swift-6.0.swift.tera > ${SPM_NAME}/Package@swift-6.0.swift
+"""
+
+[tasks.swift-move-to-packages]
+private = true
+script_runner = "@shell"
+script = """
+# Rename modules inside the xcframework to match the module name (fix for Swift >6.2)
+# Workaroud for cargo-swift v0.10.0 not supporting ffi_module_name
+find "${SPM_NAME}/Rust${SPM_NAME}.xcframework" -depth -type d -name "Rust${SPM_NAME}" -exec bash -c 'mv "$1" "${1%Rust'${SPM_NAME}'}livekit_uniffiFFI"' _ {} \; || true
```
pblazej: Trying to fix it here https://github.com/antoniusnaumann/cargo-swift/pull/87

livekit-uniffi/Makefile.toml
```diff
@@ -0,0 +1,332 @@
...
+LIB_NAME = "lib${CARGO_MAKE_CRATE_FS_NAME}"
+TARGET_DIR = "${CARGO_MAKE_CRATE_TARGET_DIRECTORY}"
+SWIFT_VERSION = "6.0"
+SPM_NAME = "LiveKitUniFFI"
+XCFRAMEWORK_NAME = "Rust${SPM_NAME}"
+NPM_NAME = "@livekit/uniffi"
+
+# Base URL for downloading release builds of the cdylib.
+# Currently, it is assumed the archives will be named "ffi-<platform>-<arch>.zip".
+CDYLIB_DOWNLOAD_BASE_URL = "https://github.com/livekit/example/releases/download/example/build"
```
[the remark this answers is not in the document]
1egoman: @ladvoc FYI I think the release github actions workflows from the aic related copy of this logic needs to be brought over here, and this needs to be updated to point to wherever those new uniffi library builds get published. I can give it a shot but I think you have a better handle on how that logic works since you initially wrote it.

livekit-uniffi/support/node/download-lib.ts.tera
```diff
@@ -0,0 +1,127 @@
...
+function generateLibraryName(platform: TargetPlatform) {
+  switch (platform) {
+    case "windows":
+      return `${PACKAGE_NAME}.dll`;
+    case "macos":
+      return `lib${PACKAGE_NAME}.dylib`;
+    case "linux":
+      return `lib${PACKAGE_NAME}.so`;
+  }
+}
```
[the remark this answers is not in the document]
1egoman: Ah oops, this was a bug in a previous implementation I forgot to copy over. FIxed in 5f918ed.

livekit-uniffi/python_test/main.py
```diff
@@ -0,0 +1,55 @@
...
+
+    async def receive_log_messages():
+        while True:
+            message = await log_forward_receive()
+            if message is None:
+                print("Log forwarding ended")
+                break
+            print(f"Log from Rust: {message}")
+
+    asyncio.run(receive_log_messages())
```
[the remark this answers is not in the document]
1egoman: This is by design.

PR #754, opened 2025-10-20
update tokio-tungstenite and tokio-rustls

Be-ing:
split off from #609

### DIFF
--- a/.github/workflows/gen-node-proto.yml
+++ b/.github/workflows/gen-node-proto.yml
@@ -41,9 +41,6 @@ jobs:
           version: "25.1"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - uses: actions/checkout@v6
-        with:
-          submodules: recursive
       - uses: pnpm/action-setup@v4
         with:
           package_json_file: livekit-ffi-node-bindings/package.json
--- a/livekit-ffi-node-bindings/src/proto/audio_frame_pb.ts
+++ b/livekit-ffi-node-bindings/src/proto/audio_frame_pb.ts
@@ -504,7 +504,7 @@ export class NewAudioSourceResponse extends Message<NewAudioSourceResponse> {
 }
 
 /**
- * Push a frame to an AudioSource 
+ * Push a frame to an AudioSource
  * The data provided must be available as long as the client receive the callback.
  *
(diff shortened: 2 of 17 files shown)


Be-ing:
tokio-rustls 0.24 has been dragging in an old rustls 0.21.12 *in addition* to the rustls 0.23 used by the rest of the dependency tree :see_no_evil:

livekit-api/src/signal_client/signal_stream.rs
```diff
@@ -248,11 +250,13 @@ impl SignalStream {
                             }
 
                             let tls_config = rustls::ClientConfig::builder()
-                                .with_safe_defaults()
```
Be-ing, line 251: It seems like this API was simply removed. I'm unsure if there's anything that needs to be done to replace it.

ladvoc: According to the [release notes](https://github.com/rustls/rustls/releases/tag/v%2F0.22.0) for _rusttls_ v0.22.0, "safe defaults are now implicit," so this should be fine.

ladvoc:
Currently, E2E tests in CI run with a local SFU in dev mode without TLS. In order to make sure this doesn't break anything, we need to verify E2E tests pass with a LiveKit Cloud SFU and the `rustls-tls-native-roots` feature enabled:
```sh
LIVEKIT_URL=wss://my-project.livekit.cloud \
LIVEKIT_API_KEY=xxx \
LIVEKIT_API_SECRET=xxx \
cargo test --features rustls-tls-native-roots,__lk-e2e-test -- --nocapture
```

I'm going to look into what the issue is here, but currently, I get the following error when running E2E tests with TLS:
```
Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled. See the documentation of the CryptoProvider type for more information.
```

Be-ing:
> I think this is the issue: https://github.com/rustls/rustls/issues/1938#issuecomment-2239196348

I think so. livekit-api's rustls-tls-native-roots feature activates reqwest's rustls-tls-native-roots feature, which activates the ring feature of rustls. I think this was not causing a problem before because of the separate old version of rustls being used by livekit-api.

PR #756, opened 2025-10-20
Parse unverified JWT

pblazej:
Enables "caching token source" on the client without 3rd party dependencies.


livekit-api/src/access_token.rs
```diff
@@ -151,6 +151,28 @@ pub struct Claims {
...
 }
 
+impl Claims {
+    pub fn with_unverified(token: &str) -> Result<Self, AccessTokenError> {
+        let mut validation = jsonwebtoken::Validation::new(jsonwebtoken::Algorithm::HS256);
+        validation.validate_exp = true;
+        #[cfg(test)]
+        {
+            validation.validate_exp = false;
+        }
```
1egoman: Out of curiosity, could it be possible to add make the test token have an expiry time way off in the future (like greater than the year `2100`) and get rid of disabling this conditionally?

livekit-api/src/access_token.rs
```diff
@@ -383,4 +405,44 @@ mod tests {
             claims
         );
     }
+
+    #[test]
+    fn test_unverified_token() {
+        let claims = Claims::with_unverified(TEST_TOKEN).expect("Failed to parse token");
```
1egoman: IMO, it would be good to have a 2 other tests:
- Try parse a token that was valid in the past but no longer is valid, and make sure it fails
- Try to parse a token which will become valid in the future and make sure it fails

livekit-api/src/access_token.rs
```diff
@@ -151,6 +151,28 @@ pub struct Claims {
     pub room_config: Option<livekit_protocol::RoomConfiguration>,
 }
 
+impl Claims {
+    pub fn with_unverified(token: &str) -> Result<Self, AccessTokenError> {
```
ladvoc: nitpick: I think `from_unverified` is a more descriptive name (it decodes Claims _from_ an unverified token)

livekit-uniffi/src/access_token.rs
```diff
@@ -179,18 +198,17 @@ pub fn verify_token(
...
+}
+
+/// Parses an access token without verifying its signature.
+///
+/// This is useful when you want to inspect token contents without having the secret.
+/// The token's expiration (exp) and not-before (nbf) times are still validated.
+/// WARNING: Do not use this for authentication - the signature is not verified!
+///
+#[uniffi::export]
+pub fn unverified_token(token: &str) -> Result<Claims, AccessTokenError> {
```
ladvoc: question: Not specific to this PR, but in Swift, symbols will be imported in the global namespace, right? If so, we might consider using a standardized naming scheme for exported symbols from each module. For example, in this module we would have `token_claims_from_unverified` (this method), `token_verify`, and `token_generate`—everything is prefixed with "token" so it is clear which module it comes from.

pblazej: yep I wanted to ask that in Notion actually, it's not possible to create something like `Claims.myNewMethod` at all?

livekit-api/src/access_token.rs
xianshijing-lk: +1,  can we use https://www.jwt.io/ to generate signed tokens that with long expired date for testing purpose ?

livekit-uniffi/src/access_token.rs
1egoman: ^ I've only played around with the python bindings, but at least with those I was able to export a class as an "object" with static methods (I think I had to mark them technically as "alternate constructors") and that seemed to work. But maybe the swift bindings are different.

livekit-api/src/access_token.rs
pblazej: 9d9cdb3f911fd8f133d09a9fb3ce3d3afccf1849

livekit-api/src/access_token.rs
pblazej: 472f566593a6b38fa830f55c81d61b232f336fac

livekit-uniffi/src/access_token.rs
pblazej: I applied the `token_` prefix, I don't think we can do more about namespaces (am I wrong?)

PR #765, opened 2025-10-24
fix: fix unable to locate __arm_tpidr2_save for android ffi.

cloudwebrtc:
Additional changes:
* H265/Simulcast Support for iOS/macOS/Android
* Tested in Unity Android app

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -105,4 +105,4 @@ jobs:
           ln -sf $ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/lib/${{ matrix.ndk_arch }}/{libunwind.so,libc++abi.a} $ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/lib/
           cargo install cargo-ndk
           # FIXME: wgpu_room does not link on aarch64
-          cargo ndk --target ${{ matrix.target }} build --release -p livekit --workspace --exclude wgpu_room -vv
+          cargo ndk --target ${{ matrix.target }} build --release -p livekit --workspace -vv


webrtc-sys/libwebrtc/patches/disable_sme_for_libyuv.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/libyuv.gni
+@@ -24,7 +24,7 @@ declare_args() {
+   # errors on Fuchsia, macOS, and compilation errors on Windows.
+   # TODO: bug 359006069 - Remove the restriction after the linker and
+   # compilation errors are fixed.
+-  libyuv_use_sme = current_cpu == "arm64" && (is_linux || is_android)
++  libyuv_use_sme = current_cpu == "arm64" && (is_linux)
+   libyuv_use_msa =
+       (current_cpu == "mips64el" || current_cpu == "mipsel") && mips_use_msa
+   libyuv_use_mmi =
```
xianshijing-lk, line 13: [question](https://pullpo.io/cc?l=question) 

is it intentional to set    libyuv_use_mmi =
here ? 
should it be    libyuv_use_mmi = flase ? or true

xianshijing-lk (approved):
one question

webrtc-sys/libwebrtc/patches/disable_sme_for_libyuv.patch
cloudwebrtc: I just used `libyuv_use_sme=false` in `gn gen --args="....."`, but it is actually invalid. `__arm_tpidr2_save` is still generated in ffi.so.

Be-ing:
You should be able to remove the `--exclude wgpu_room` from `cargo ndk` in .github/workflows/builds.yml

PR #768, opened 2025-10-28
Examples for local camera publishing

chenosaurus:


examples/local_video/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+name = "publisher"
+path = "src/publisher.rs"
+
+[[bin]]
+name = "subscriber"
+path = "src/subscriber.rs"
+
+[dependencies]
+tokio = { version = "1", features = ["full", "parking_lot"] }
+livekit = { workspace = true, features = ["rustls-tls-native-roots"] }
```
Be-ing, line 17: What is the need for the rustls-tls-native-roots feature for this example? This will force the feature on the whole workspace, not just this example.

examples/local_video/src/publisher.rs
```diff
@@ -0,0 +1,331 @@
...
+    info!("Target frame interval: {:.2} ms", target.as_secs_f64() * 1000.0);
+
+    // Timing accumulators (ms) for rolling stats
+    let mut sum_get_ms = 0.0;
+    let mut sum_decode_ms = 0.0;
+    let mut sum_convert_ms = 0.0;
+    let mut sum_capture_ms = 0.0;
+    let mut sum_sleep_ms = 0.0;
+    let mut sum_iter_ms = 0.0;
+    loop {
```
Be-ing, line 226: I suggest to let the user break this loop with Ctrl + C like https://github.com/Be-ing/rust-sdks/commit/81acf7cf53e3f3367d6ec096f52606952e639c4f

PR #771, opened 2025-10-31
feat: add outbound trunk config for create_sip_participant.

cloudwebrtc:


livekit-api/src/services/sip.rs
```diff
@@ -417,13 +417,15 @@ impl SIPClient {
         call_to: String,
         room_name: String,
         options: CreateSIPParticipantOptions,
+        outbound_trunk: Option<proto::SipOutboundConfig>,
```
xianshijing-lk: suggestion, use outbound_trunk_config rather than outbound_trunk,

xianshijing-lk (approved):
lgtm

PR #772, opened 2025-11-01
(WIP) Audio core module with UniFFI bindings

ladvoc:


pblazej:
Swift → Rust experiment

- Starting point: `UnsafeMutablePointer<Float>`
  - https://developer.apple.com/documentation/avfaudio/avaudiopcmbuffer/floatchanneldata
  - https://github.com/livekit/client-sdk-swift/blob/main/Sources/LiveKit/Types/AudioBuffer.swift#L45
- Rust APIs:
  - `fn push(&self, input: &[i16])` → `[Int16]` - requires copy on the Swift side already (`Array` owns its data buffer) 🔴 
  - `fn push(&self, input: AudioBufferBytes)` → `Data` 🟠 
    - does not require copy on the Swift side due to https://developer.apple.com/documentation/foundation/data/init(bytesnocopy:count:deallocator:)
    - does copy over FFI boundary because of the actual `lower` impl

```rust
pub struct AudioBufferBytes {
    pub data: Vec<u8>,
}
```

```swift
var writer = createWriter()
write(value, into: &writer)
return RustBuffer(bytes: writer) // here comes the copy
```

- brute-force `fn push_from_ptr(&self, data_ptr: u64, len: u32)` - works, no copy on either side, no ownership, no safety 🟢 

```swift
let frameCount = audioBuffer.frames
let channelBuffer = audioBuffer.rawBuffer(forChannel: 0)

// Arithmetics...
try resampler.pushFromPtr(dataPtr: UInt64(bitPattern: Int64(Int(bitPattern: channelBuffer))), len: UInt32(frameCount * MemoryLayout<Float>.size))
```

Conclusion: it boils down to using `u64` under the hood, where the conversion really happens (bindgen or Swift) probably is not _that_ important

```rust
#[derive(uniffi::Record)]
pub struct AudioBuffer {
    /// Pointer value as u64 (points to Float32 data in Swift memory)
    pub pointer: u64,
    /// Length in bytes
    pub len: u32,
}

impl AudioBuffer {
    /// Get the pointer as *const u8
    pub fn as_ptr(&self) -> *const u8 {
        self.pointer as *const u8
    }
    
    /// Get length in bytes
    pub fn byte_len(&self) -> usize {
        self.len as usize
    }
}
```

PR #778, opened 2025-11-04
Fix the fast path in capture_frame function, without buffering

xianshijing-lk:
As context:
queue_size_ms=0 is supposed to disable the buffer:
https://github.com/livekit/rust-sdks/blob/0773bcec4e24812b382e71166b2ab359bf1a9384/webrtc-sys/src/audio_track.cpp#L203-L226

But it is incorrectly handled in audio_source.rs:
https://github.com/livekit/rust-sdks/blob/0773bcec4e24812b382e71166b2ab359bf1a9384/libwebrtc/src/native/audio_source.rs#L87

Calling chunks with 0 causes panic.

This PR implements the fast path that push the frame to webrtc directly

### DIFF
--- a/libwebrtc/src/native/audio_source.rs
+++ b/libwebrtc/src/native/audio_source.rs
@@ -27,6 +27,25 @@ pub struct NativeAudioSource {
 }
 
 impl NativeAudioSource {
+    /// Creates a new [`NativeAudioSource`].
+    ///
+    /// # Arguments
+    /// * `options` – Configuration options for the source (e.g. echo cancellation, noise suppression).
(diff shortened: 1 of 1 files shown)


libwebrtc/src/native/audio_source.rs
ladvoc, line 49: I think this could be a good opportunity to include a doc comment on this constructor; for the `queue_size_ms` argument, it should mention that this value can be zero (indicating no buffering) and that it must be a multiple of 10 if non-zero.

xianshijing-lk: Good idea. Done.

PR #779, opened 2025-11-04
AudioResampler: initialize frame_.samples_per_channel_

Be-ing:
This change is needed for the same reason as
b776e64a3e1a33b92ffe94ffba389a318ae87f41 (#722):

A change in the resampler API introduced in libwebrtc m137 caused the output frame's samples_per_channel_ property to remain uninitialized. As a result, the length of the output slice was always calculated as zero, resulting in NativeAudioStream yielding audio frames without any sample data.


xianshijing-lk:
I just landed it. Sorry for missing the email notification.

PR #781, opened 2025-11-10
change the livekit-ffi to output a static lib for cpp SDK

xianshijing-lk:
Otherwise the latest commit can't be linked by the client-sdk-cpp


ladvoc:
I've triggered the [FFI builds workflow](https://github.com/livekit/rust-sdks/actions/runs/19225212928) to make sure this doesn't break anything.

Issue #481, opened 2024-11-05
Reconnection attempts do not honor the reconnection interval 


TomaszMysliwiec:
Based on my internal tests the reconnection attempts do not honor the reconnection interval or at least in most cases will not honor it.

This might be related to the fact that `reconnecting_interval` is spawned after connecting and while in the session it's constantly stacking up events causing multiple instant resolutions inside the reconnect task depending on how much time passed:
https://github.com/livekit/rust-sdks/blob/06997356c083b01da766c4c93e8c4e8354d871ce/livekit/src/rtc_engine/mod.rs#L695

It looks like the custom interval doesn't support different `MissedTickBehavior`:
https://github.com/livekit/rust-sdks/blob/06997356c083b01da766c4c93e8c4e8354d871ce/livekit-runtime/src/async_std.rs#L11-L12

so potential solution would be to add `inner.reconnecting_interval.lock().await.reset();` before starting the reconnection task:
https://github.com/livekit/rust-sdks/blob/06997356c083b01da766c4c93e8c4e8354d871ce/livekit/src/rtc_engine/mod.rs#L599-L600
but that's just an initial, potentially dirty, quick fix.

Please take a look into this issue as it reduces the designed 50-second reconnection window down to 10-15 seconds.

TMysliwiec:
This is still not fixed. I created a PR with a fix here:
https://github.com/livekit/rust-sdks/pull/785

TMysliwiec:
This is still not fixed. I need one more approval on the PR, and we are good to go
https://github.com/livekit/rust-sdks/pull/786

PR #785, opened 2025-11-15
Fix reconnection logic to not exhaust retries prematurely

TMysliwiec:
This is a fix for: https://github.com/livekit/rust-sdks/issues/481, which was just closed again, likely due to inactivity.

**Problem:**
The reconnection logic doesn't properly honor the reconnection interval. The interval starts counting from engine creation, not from when reconnection begins. If the engine runs for over 50 seconds (RECONNECT_ATTEMPTS × RECONNECT_INTERVAL), the reconnection loop immediately executes all attempts without waiting, causing it to fail after a couple of seconds. That is because we are using Tokio's default interval, which utilizes [MissedTickBehavior::Burst](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Burst). Any accumulated ticks will fire immediately.
https://github.com/livekit/rust-sdks/blob/06997356c083b01da766c4c93e8c4e8354d871ce/livekit/src/rtc_engine/mod.rs#L52

**Solution**:
By changing the interval missed tick behavior to [MissedTickBehavior::Delay](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Delay), we can achieve the potentially initially desired await before making another reconnection attempt. That way, the reconnection window is significantly larger, thereby increasing the likelihood of reconnection.

**Reproduction steps**:
The problem is quite simple to reproduce. You need to run the `basic_room` example, wait a minute, disconnect the internet, and observe how fast the reconnection attempts are exhausted.

### DIFF
--- a/livekit/src/rtc_engine/mod.rs
+++ b/livekit/src/rtc_engine/mod.rs
@@ -17,7 +17,7 @@ use std::{borrow::Cow, fmt::Debug, sync::Arc, time::Duration};
 use libwebrtc::prelude::*;
 use livekit_api::signal_client::{SignalError, SignalOptions};
 use livekit_protocol as proto;
-use livekit_runtime::{interval, Interval, JoinHandle};
+use livekit_runtime::{interval, Interval, JoinHandle, MissedTickBehavior};
 use parking_lot::{RwLock, RwLockReadGuard};
 use thiserror::Error;
 use tokio::sync::{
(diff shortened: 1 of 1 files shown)


TMysliwiec:
@davidzhao not sure if there is anything more I can do in terms of the CLA. When I click recheck the status it reroutes me to this page and on their side it says it's already signed.
There is a lot of similar issues like that on the web so seems I am not alone on that one.

TMysliwiec:
I think I have mixed commits and submit using my 2 accounts. Let me recreate the PR

TMysliwiec:
@davidzhao created new one: https://github.com/livekit/rust-sdks/pull/786

PR #790, opened 2025-11-18
feat(connector): initial service impl

anunaym14:


livekit-api/src/services/connector.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    pub fn new(host: &str) -> ServiceResult<Self> {
+        let (api_key, api_secret) = get_env_keys()?;
+        Ok(Self::with_api_key(host, &api_key, &api_secret))
+    }
+
+    /// Dials a WhatsApp call
+    ///
+    /// # Arguments
+    /// * `phone_number_id` - The number of the business initiating the call
```
davidzhao: is this a phone number? or an identifier? if a phone number, what is the format? is it E.164?

livekit-api/src/services/connector.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// * `options` - Additional options for the call
+    ///
+    /// # Returns
+    /// Information about the dialed call including the WhatsApp call ID and room name
+    pub async fn dial_whatsapp_call(
+        &self,
+        phone_number_id: impl Into<String>,
+        to_phone_number: impl Into<String>,
+        api_key: impl Into<String>,
+        cloud_api_version: impl Into<String>,
```
davidzhao, line 117: is API version required only for the initial dial API? I don't see it in subsequent requests.

livekit-api/src/services/connector.rs
anunaym14: it's an identifier

livekit-api/src/services/connector.rs
anunaym14: yes, we only added it to initial requests since api version is unlikely to change in the middle of a call

anunaym14: connect does not do any api calls, so it's just disconnect that's missing it

livekit-api/src/services/connector.rs
davidzhao: maybe update the comment to say `identifier of the number for business initiating the call`

davidzhao (approved):
just a small nit on comment for clarity

PR #799, opened 2025-12-01
fix: fix mute/unmute events for LocalTrack.

cloudwebrtc:
The room is unable to emit mute/unmute events. This is because  `publication.set_track` for local tracks was not called to register event handlers, and there is another minor bug.

### DIFF
--- a/examples/local_audio/src/main.rs
+++ b/examples/local_audio/src/main.rs
@@ -339,15 +339,13 @@ async fn stream_audio_to_livekit_with_shared_apm(
     Ok(())
 }
 
-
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -7,6 +7,18 @@
             "name": "wgpu_room(debug)",
             "program": "${workspaceFolder}/examples/target/debug/wgpu_room",
             "preLaunchTask": "build"
+        },
```
ladvoc, line 10: I don't have a strong opinion, but do you think we should commit this launch configuration?

PR #613, opened 2025-03-27
Expose WebRTC's audio_mixer

ConradIrwin:
This allows one to use the apm to cancel echo in multiway calls.

Co-Authored-By: Mikayla <mikayla@zed.dev>
Co-Authored-By: Zed AI <ai+claude-3.5@zed.dev>

PR #806, opened 2025-12-04
Expose WebRTC's audio_mixer

Be-ing:
This allows one to use the apm to cancel echo in multiway calls.

This is a rebase of https://github.com/livekit/rust-sdks/pull/613 with merge conflicts fixed.

### DIFF
--- a/libwebrtc/src/native/audio_mixer.rs
+++ b/libwebrtc/src/native/audio_mixer.rs
@@ -27,7 +27,7 @@ pub use ffi::AudioFrameInfo;
 pub trait AudioMixerSource {
     fn ssrc(&self) -> i32;
     fn preferred_sample_rate(&self) -> u32;
-    fn get_audio_frame_with_info<'a>(&self, target_sample_rate: u32) -> Option<AudioFrame>;
+    fn get_audio_frame_with_info(&self, target_sample_rate: u32) -> Option<AudioFrame>;
 }
 
 struct AudioMixerSourceImpl<T> {
@@ -42,7 +42,7 @@ impl<T: AudioMixerSource> sys::AudioMixerSource for AudioMixerSourceImpl<T> {
         self.inner.preferred_sample_rate() as i32
     }
 
-    fn get_audio_frame_with_info<'a>(
+    fn get_audio_frame_with_info(
         &self,
         target_sample_rate: i32,
         native_frame: sys::NativeAudioFrame,
--- a/webrtc-sys/src/audio_mixer.rs
+++ b/webrtc-sys/src/audio_mixer.rs
@@ -69,10 +69,10 @@ pub mod ffi {
 pub trait AudioMixerSource {
     fn ssrc(&self) -> i32;
     fn preferred_sample_rate(&self) -> i32;
-    fn get_audio_frame_with_info<'a>(
+    fn get_audio_frame_with_info(
         &self,
         target_sample_rate: i32,
-        frame: NativeAudioFrame<'a>,
+        frame: NativeAudioFrame,
     ) -> AudioFrameInfo;
 }
 


Be-ing:
@xianshijing-lk This is an older PR that Zed has been using downstream for a while. I rebased it and ran clang-format on the C++ code. It would be great to get this merged.

libwebrtc/src/native/audio_mixer.rs
```diff
@@ -0,0 +1,108 @@
...
+                "sample rate must match target_sample_rate"
+            );
+            assert_eq!(
+                frame.samples_per_channel as usize, samples_count,
+                "frame must contain 10ms of samples"
+            );
+            assert_eq!(
+                frame.data.len(),
+                samples_count * frame.num_channels as usize,
+                "slice must contain 10ms of samples"
```
xianshijing-lk, line 63: [question](https://pullpo.io/cc?l=question) 

does the audio mixer work with samples of multiple 10ms ? or it only works 10ms of sample ?

libwebrtc/src/native/audio_mixer.rs
```diff
@@ -0,0 +1,108 @@
...
+pub struct AudioMixer {
+    sys_handle: UniquePtr<ffi::AudioMixer>,
+}
+
+pub use ffi::AudioFrameInfo;
+
+pub trait AudioMixerSource {
+    fn ssrc(&self) -> i32;
+    fn preferred_sample_rate(&self) -> u32;
+    fn get_audio_frame_with_info<'a>(&self, target_sample_rate: u32) -> Option<AudioFrame>;
```
xianshijing-lk: curiously, do you need 'a ? 
or just 
fn get_audio_frame_with_info(&self, target_sample_rate: u32) -> Option<AudioFrame>;

libwebrtc/src/native/audio_mixer.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    pub fn remove_source(&mut self, ssrc: i32) {
+        unsafe {
+            self.sys_handle.pin_mut().remove_source(ssrc);
+        }
+    }
+
+    pub fn mix(&mut self, num_channels: usize) -> &[i16] {
+        unsafe {
+            let len = self.sys_handle.pin_mut().mix(num_channels);
```
xianshijing-lk, line 104: My understanding here is that, it is expected that the len is samples ? rather than bytes ? 

can you check it ?

libwebrtc/src/native/audio_mixer.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn preferred_sample_rate(&self) -> i32 {
+        self.inner.preferred_sample_rate() as i32
+    }
+
+    fn get_audio_frame_with_info<'a>(
+        &self,
+        target_sample_rate: i32,
+        native_frame: sys::NativeAudioFrame,
+    ) -> AudioFrameInfo {
+        if let Some(frame) = self.inner.get_audio_frame_with_info(target_sample_rate as u32) {
```
xianshijing-lk, line 50: [suggestion](https://pullpo.io/cc?l=suggestion) 
this will call back to Rust, is there any way to avoid it ?

libwebrtc/src/native/audio_mixer.rs
Be-ing: This function is returning info about one specific frame, so it wouldn't make much sense to return one AudioFrameInfo struct for multiple frames.

libwebrtc/src/native/audio_mixer.rs
Be-ing: Good catch, this lifetime generic is not needed.

webrtc-sys/src/audio_mixer.rs
```diff
@@ -0,0 +1,106 @@
...
+        fn preferred_sample_rate(self: &AudioMixerSourceWrapper) -> i32;
+        fn get_audio_frame_with_info(
+            self: &AudioMixerSourceWrapper,
+            target_sample_rate: i32,
+            frame: Pin<&mut NativeAudioFrame>,
+        ) -> AudioFrameInfo;
+    }
+}
+
+pub trait AudioMixerSource {
```
xianshijing-lk, line 69: How is the thread safety for the audio mixer ? 

I feel like they are not safe but my limited rust knowledge is not enough to get deep into it.

@ladvoc, could you please help review the code and look at the thread safety ?

libwebrtc/src/native/audio_mixer.rs
```diff
@@ -0,0 +1,108 @@
...
+            unsafe {
+                native_frame.update_frame(
+                    0,
+                    frame.data.as_ptr(),
+                    frame.samples_per_channel as usize,
+                    frame.sample_rate as i32,
+                    frame.num_channels as usize,
+                );
+            }
+            return ffi::AudioFrameInfo::Normal;
```
ladvoc: nitpick: The return keyword isn't needed here or in the else block.

libwebrtc/src/native/audio_mixer.rs
```diff
@@ -0,0 +1,108 @@
...
+        self.inner.preferred_sample_rate() as i32
+    }
+
+    fn get_audio_frame_with_info<'a>(
+        &self,
+        target_sample_rate: i32,
+        native_frame: sys::NativeAudioFrame,
+    ) -> AudioFrameInfo {
+        if let Some(frame) = self.inner.get_audio_frame_with_info(target_sample_rate as u32) {
+            let samples_count = (frame.sample_rate as usize / 100) as usize;
```
ladvoc: nitpick: The final cast to `usize` is redundant.

webrtc-sys/src/audio_mixer.rs
Be-ing: Great question. I merely rebased this code and did not write it, so I was not very familiar before you asked. The thread safety of livekit::AudioMixer on the C++ side is not good. livekit::AudioMixer stores its own `std::vector<std::shared_ptr<AudioMixerSource>> sources_` member. I don't see the reason for this. libwebrtc's webrtc::AudioMixerImpl stores its own list of sources internally which are guarded by a mutex. So livekit::AudioMixer seems to add a redundant list on top of that but without the thread safety, despite the manual impl's for Send + Sync traits.

This seems to not have been a practical issue for Zed because Zed wraps its AudioMixer with `Arc<Mutex>` on the Rust side. However, just looking at the Rust API exposed by the libwebrtc crate here, there is no reason for anyone who didn't write this code to suspect that `Arc<Mutex>` is necessary without documentation. There are also no SAFETY comments in the higher level Rust wrappers in the libwebrtc crate for the unsafe blocks documenting why it is safe to call the unsafe lower level functions.

From a quick look, I think there are several other parts of these new APIs that are redundant or overcomplicated.

I will talk with @ConradIrwin about how to proceed here. We have a pair coding call scheduled for Friday.

ConradIrwin: I think I did it that way to make it easier to implement remove, but there are probably other approaches we could take.

libwebrtc/src/native/audio_mixer.rs
xianshijing-lk: sounds good, thanks for the clarification.

libwebrtc/src/native/audio_mixer.rs
xianshijing-lk: what do you think here ? 

I think the current code should "work", but it might allocate more memory than it needs.

webrtc-sys/src/audio_mixer.rs
xianshijing-lk: Oh, my previous replies were lost. 

I think std::vector<std::shared_ptr<AudioMixerSource>> sources_ is needed, livekit::AudioMixer  AddSouce / RemoveSouce does call into webrtc::AudioMixerImpl, but they just pass the raw pointers there, thus WebRTC lib does not own the object. That is the reason that std::vector<std::shared_ptr<AudioMixerSource>> sources_ solve the problem as we need a owner to the AudioMixerSource. 
And we need to keep them. 

I think Zed solved the racing issue by using Arc<Mutex>. 

My suggestion is to make the C++   livekit::AudioMixer thread safe, something like 

std::mutex sources_mutex_;

```
void AudioMixer::add_source(rust::Box<AudioMixerSourceWrapper> source) {
  auto native_source = std::make_shared<AudioMixerSource>(std::move(source));
  {
    std::lock_guard<std::mutex> lock(sources_mutex_);
    audio_mixer_->AddSource(native_source.get());
    sources_.push_back(native_source);
  }
}

void AudioMixer::remove_source(int source_ssrc) {
  std::lock_guard<std::mutex> lock(sources_mutex_);
  auto it = std::find_if(
      sources_.begin(), sources_.end(),
      [source_ssrc](const auto& s) { return s->Ssrc() == source_ssrc; });

  if (it != sources_.end()) {
    audio_mixer_->RemoveSource(it->get());
    sources_.erase(it);
  }
}
```

libwebrtc/src/native/audio_mixer.rs
Be-ing: Oh, good catch. `len` should be number of samples but the C++ code was calculating the number of bytes by multiplying by `sizeof(int16_t)`

libwebrtc/src/native/audio_mixer.rs
ConradIrwin: The inner value is a trait to let apps bring their own sources (kinda like how WebRTC structures its source stuff).

That's nice because you can be flexible in buffer management, and thinking about where the source comes from, rather than having to use a concrete type provided by the live kit crate.

We could move the source implementation from the Zed codebase into the live kit crate and give it a concrete type, but it seems like it'd reduce flexibility, and still end up calling rust code under the hood.

libwebrtc/src/native/audio_mixer.rs
ConradIrwin: Good catch! I think this actually could lead to out-of-bounds memory reads as the slice claims access to twice as much memory as it should, we'll fix. (Luckily these buffers are only passed to other functions with similarly restricted sizes so they don't do that!)

libwebrtc/src/native/audio_mixer.rs
xianshijing-lk: I see, thanks for the clarification. I am ok to keep it.

webrtc-sys/src/audio_mixer.rs
Be-ing: Good point about the lifetime; AudioMixerImpl in libwebrtc does keep its own vector of sources internally, but those are AudioMixerImpl::SourceStatus structs which have a plain pointer Source* member. So it is helpful to have this vector of std::shared_pointers above it.

We added a mutex guarding add/remove_source to resolve the thread safety issue with that. Unless you see any more thread safety issues, I think it's okay now.

xianshijing-lk:
I think you can merge it now

xianshijing-lk:
I see. Sorry, I missed that. And just pressed the button here.

PR #814, opened 2025-12-17
fix: lazy loading for additional dependencies.

cloudwebrtc:


licenserc.toml
```diff
@@ -4,6 +4,9 @@ excludes = [
     "soxr-sys/src",
     "webrtc-sys/libwebrtc",
     "webrtc-sys/test",
+    "webrtc-sys/src/lazy_load_so/desktop_capturer",
+    "webrtc-sys/src/lazy_load_so/vaapi",
+    "webrtc-sys/src/lazy_load_so/nvidia",
     "webrtc-sys/src/vaapi",
     "webrtc-sys/src/nvidia",
```
lukasIO, line 9: Just to clarify, why is this still needed even with the lazy loading?

cloudwebrtc: These files are used to isolate the direct dependency of liblivekit_ffi.so on libX11.so during linking. At runtime, the actual libX11.so is lazily loaded through proxy function calls in these files.

cloudwebrtc: I used this tool to generate these files automatically:

https://github.com/yugr/Implib.so

In addition, we must find the relevant function dependencies in libwebrtc, wrap them one by one into proxy functions with dlopen, or separate the desktop capturer into a separate .so module, but this would make the project fragmented. We would need to maintain more binary packages, flags, and other control parameters.

lukasIO: I meant these ones
```
 "webrtc-sys/src/vaapi",
 "webrtc-sys/src/nvidia"
 ```
 
 My naive assumption was that these wouldn't be required anymore if we do the lazyloading approach

webrtc-sys/src/lazy_load_deps_for/generate_implibs.sh
lukasIO, line 1: thank you for adding it 🙏

PR #820, opened 2025-12-22
improving initial video quality by setting x-google-start-bitrate for svc codecs

xianshijing-lk:
And we should do it for Desktop only

### DIFF
--- a/livekit/src/rtc_engine/peer_transport.rs
+++ b/livekit/src/rtc_engine/peer_transport.rs
@@ -140,8 +140,8 @@ impl PeerTransport {
         if ultimate_kbps == 0 {
             return None;
         }
-        // JS uses ~80% of ultimate; 100% is also reasonable per feedback.
-        let start_kbps = (ultimate_kbps as f64 * 0.8).round() as u32;
+        // JS / Flutter uses ~70% of ultimate; 100% is also reasonable per feedback.
+        let start_kbps = (ultimate_kbps as f64 * 0.7).round() as u32;
 
         // Clamp: avoid silly low/high values
         Some(start_kbps.clamp(300, ultimate_kbps))


livekit/src/rtc_engine/peer_transport.rs
```diff
@@ -26,6 +26,12 @@ use super::EngineResult;
 
 pub type OnOfferCreated = Box<dyn FnMut(SessionDescription) + Send + Sync>;
 
+#[cfg(any(target_os = "windows", target_os = "macos", target_os = "linux"))]
+const DEFAULT_VP9_START_BITRATE_KBPS: u32 = 2500;
```
davidzhao: I think we should use the currently negotiated bitrate? instead of a fixed amount? the JS logic uses 80% of ultimate bitrate. I think it's ok to go to 100% as well

livekit/src/rtc_engine/peer_transport.rs
```diff
@@ -127,6 +133,52 @@ impl PeerTransport {
         Ok(answer)
     }
 
+    fn munge_x_google_start_bitrate(sdp: &str, start_bitrate_kbps: u32) -> String {
```
davidzhao, line 150: should we use a proper sdp parser/rewriter? instead of string changes? https://crates.io/crates/sdp

in case we'd need to do more munging for other things

livekit/src/rtc_engine/peer_transport.rs
```diff
@@ -151,7 +203,25 @@ impl PeerTransport {
             return Ok(());
         }
 
-        let offer = self.peer_connection.create_offer(options).await?;
+        let mut offer = self.peer_connection.create_offer(options).await?;
+        let start_bitrate_kbps = DEFAULT_VP9_START_BITRATE_KBPS;
+        let sdp = offer.to_string();
+        // TODO, we should extend the codec support to AV1 ?
```
davidzhao: worth testing if AV1 has this issue or not.. the JS logic is applying to AV1 as well

livekit/src/rtc_engine/peer_transport.rs
xianshijing-lk: I will address the comment

livekit/src/rtc_engine/peer_transport.rs
xianshijing-lk: @ladvoc @cloudwebrtc , what do you think here ? 

If this is the only place that we will need to parse sdp, I would prefer keeping this handmade parser rather than introducing a new dep

livekit/src/rtc_engine/peer_transport.rs
xianshijing-lk: Do we have negotiated birate in SDP ? any one can point me to the relevant code or give me some hint ? 

@ladvoc  @cloudwebrtc

cloudwebrtc: In Flutter, the preset `maxBitrate * 0.7` is used as the `startBitrate` here.

https://github.com/livekit/client-sdk-flutter/blob/ee938fd1451ae80f41b5fb25bca90988df5e8aa5/lib/src/core/transport.dart#L207

cloudwebrtc: Similar changes from JS:

https://github.com/livekit/client-sdk-js/blob/a7b19395d944439170a0649a3d5eca9bb4b49f09/src/room/PCTransport.ts#L320-L342

xianshijing-lk: thanks, I addressed the comments.

livekit/src/rtc_engine/peer_transport.rs
davidzhao: JS munges SDP for several purposes, including enabling stereo publishing, and a few other things

livekit/src/rtc_engine/peer_transport.rs
davidzhao: why are we using 70 or 80? is there a good reason to? I'd like to avoid magic numbers if possible.

livekit/src/rtc_engine/peer_transport.rs
ladvoc: I think brining in the SDP crate you have linked to do munging is ok. The crate is lightweight (minimal [transitive dependencies](https://crates.io/crates/sdp/0.10.0/dependencies)) and it comes from the webrtc-rs project.

livekit/src/rtc_engine/peer_transport.rs
xianshijing-lk: I was using the same number for Rust to be consistent with the web

@lukasIO @1egoman , any idea why we are using a 70% of target bitrate here ? 

/* The svc codec (av1/vp9) would use a very low bitrate at the begining and
increase slowly by the bandwidth estimator until it reach the target bitrate. The
process commonly cost more than 10 seconds cause subscriber will get blur video at
the first few seconds. So we use a 70% of target bitrate here as the start bitrate to
eliminate this issue.
*/
const startBitrateForSVC = 0.7;

PR #822, opened 2026-01-02
fix: ensure Room.creation_time is ms

davidzhao:
when we initially connect to the room, we are parsing the creation_time incorrectly. it gets updated later with the correct ms values, but will remain wrong immediately after connect.


ladvoc (approved):
LGTM ✅

Issue #823, opened 2026-01-03
Participant's permission does not exposed to FFI clients


GuuJiang:
Since admins can update participant permissions via the server API to remotely enable or disable microphone/camera, FFI clients need to be aware of these permission changes to update their UI accordingly just like how the web client handles this.

Currently, the FFI protocol does not expose the `permission` field from `ParticipantInfo`, making it impossible for FFI clients to:
- Detect when permissions change
- Determine which track sources a participant is allowed to publish
- Update UI elements based on current permissions

PR #824, opened 2026-01-03
Expose participant's permission to ffi layer

GuuJiang:
fix for [ISSUE-823](https://github.com/livekit/rust-sdks/issues/823)

### DIFF
--- a/livekit/src/room/mod.rs
+++ b/livekit/src/room/mod.rs
@@ -509,6 +509,7 @@ impl Room {
             pi.metadata,
             pi.attributes,
             e2ee_manager.encryption_type(),
+            pi.permission,
         );
 
         let dispatcher = Dispatcher::<RoomEvent>::default();
@@ -661,6 +662,7 @@ impl Room {
                     pi.name,
                     pi.metadata,
                     pi.attributes,
+                    pi.permission,
                 )
             };
             participant.update_info(pi.clone());
@@ -1023,6 +1025,7 @@ impl RoomSession {
                         pi.name,
                         pi.metadata,
                         pi.attributes,
+                        pi.permission,
                     )
                 };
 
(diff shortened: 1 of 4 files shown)


livekit-ffi/protocol/room.proto
```diff
@@ -509,6 +510,11 @@ message ParticipantNameChanged {
   required string name = 2;
 }
 
+message ParticipantPermissionChanged {
+  required string participant_identity = 1;
+  optional ParticipantPermission permission = 2;
```
davidzhao, line 515: should this be optional? seems like it should always be there

livekit/src/room/mod.rs
```diff
@@ -153,6 +153,10 @@ pub enum RoomEvent {
         participant: Participant,
         is_encrypted: bool,
     },
+    ParticipantPermissionChanged {
+        participant: Participant,
+        permission: Option<proto::ParticipantPermission>,
```
davidzhao, line 158: same concern here

davidzhao (commented):
great work! just a couple of nits and I think we can merge this one.

livekit/src/room/mod.rs
GuuJiang: I had the same concerns from the beginning and have been struggling with this decision.

The root cause is that the protocol layer uses proto3 syntax, which makes the generated nested structs optional by default. Even though we know in practice it will never be empty (and we can get the semantic default value even if it's empty in the proto3 world), after careful consideration, I decided to make the permission field optional in the `ParticipantInfo` struct in the FFI layer to avoid potential breaking changes.

I completely agree that it should be required in the `ParticipantPermissionChanged` event. However, I initially thought it might be better to keep consistency with the `ParticipantInfo` struct. I'm happy to change it to required if you think that's the better approach.

BTW, I made a new commit to initialize the permission field(It should be discovered earlier if choose required instead, LOL)

PR #829, opened 2026-01-07
Add node ffi bindings

lukasIO:


livekit-ffi-node-bindings/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+edition = "2021"
+name = "rtc-node-ffi-bindings"
+version = "0.0.1"
+
```
ladvoc, line 6: This should have `publish = false` so that it doesn't get published to Crates.io when the workspace is published.

lukasIO:
This PR will most likely need a follow up, merging it into main now to verify and test the publishing workflow for the node bindings

PR #834, opened 2026-01-12
node binding generation follow up

lukasIO:
follow up for #829 

- fixes the node ffi build action
- enables rust caching across our build workflows
- fixes some licensing headers

### DIFF
--- a/.github/workflows/ffi-builds.yml
+++ b/.github/workflows/ffi-builds.yml
@@ -211,11 +211,6 @@ jobs:
           name: ffi-builds-${{ matrix.target }}
           path: ${{ matrix.name }}.zip
 
-  build_node:
-    name: Build node bindings
-    runs-on: ubuntu-latest
-    steps:
-
   release:
     name: Release to GH (Draft)
     runs-on: ubuntu-latest


.github/workflows/node-builds.yml
```diff
@@ -205,9 +197,27 @@ jobs:
...
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git add .
+          git commit -m "chore: update livekit-ffi-node-bindings version to $(echo "${{ env.TAG_NAME }}" | sed 's/.*@//')" || echo "No changes to commit"
+          git push origin HEAD
+
       - name: Build TS parts
         working-directory: livekit-ffi-node-bindings
-        run: pnpm build:ts
+        run: |
```
davidzhao: should this run before the commit portion?

livekit-ffi-node-bindings/package.json
```diff
@@ -65,7 +65,7 @@
   },
   "scripts": {
     "build:ts": "tsup --onSuccess \"tsc --declaration --emitDeclarationOnly\" && cp -r src/napi dist/ && cp -r src/napi/* dist/",
-    "build": "./generate_proto.sh && pnpm build:ts && napi build --platform --release --dts native.d.ts --js native.cjs --pipe \"prettier -w\" src/napi",
```
davidzhao, line 68: we don't need this step anymore? when doing local dev?

.github/workflows/node-builds.yml
lukasIO: Was debating whether or not to move the generate_proto step to before the commit, but felt it was a bit unexpected that only the release workflow would commit the generated proto files. 
This currently errors out if the committed protos aren't up to date on main. 

I can see three approaches that have their own tradeoffs:
- keep it as is and error out if not up to date
- run on release and commit any changes without double checking that the output is sane
- run `./generate_proto.sh` on every PR instead and commit (this is what python is doing)


thinking about it again now, I prefer the third option of running it on every PR/build step, does that sound good to you, too?

livekit-ffi-node-bindings/package.json
lukasIO: I added this part in the previous PR, but running it like this errors out on windows builds, so I kept it as a separate/manual step in the CI as it was before

.github/workflows/node-builds.yml
lukasIO: I went with the approach that python also does now and commit the proto stubs on every PR

PR #860, opened 2026-02-02
Use unwind panic for tier III Apple platforms

ladvoc:
Use unwind panic for tier III Apple platforms when building _livekit-uniffi_. This will no longer be necessary once these platforms become tier II (see rust-lang/rust#151705).


ladvoc:
Closing as tvOS/watchOS/visionOS have been promoted to tier 2.

PR #885, opened 2026-02-10
add support for rockchip h264/h265 video encoders

chenosaurus:


examples/local_video/src/publisher.rs
```diff
@@ -259,6 +259,8 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
     let mut sum_sleep_ms = 0.0;
     let mut sum_iter_ms = 0.0;
     let mut logged_mjpeg_fallback = false;
+    let mut consecutive_capture_errors: u32 = 0;
+    const MAX_CONSECUTIVE_CAPTURE_ERRORS: u32 = 30;
```
ladvoc: question: What leads to these intermittent capture errors? Is 30 the right limit to set?

webrtc-sys/build.rs
```diff
@@ -204,6 +204,29 @@ fn main() {
                 );
             }
 
+            if arm {
```
ladvoc: comment (non-blocking): I'm wondering if we should put all hardware encoders behind crate features. Ideally, we shouldn't be compiling hardware-specific code on platforms that can't take advantage of it. @cloudwebrtc, did you consider this when adding support for NVIDIA?

ladvoc (approved):
LGTM ✅, this should be tested on-device if it hasn't been already before merging.

webrtc-sys/build.rs
chenosaurus: isn't this gated by build.rs finding the right headers on the system?

PR #890, opened 2026-02-12
Add support for frame level user timestamp

chenosaurus:
- Add support to attach/parse frame level timestamps to VideoTracks as a custom payload trailer.


webrtc-sys/src/frame_cryptor.cpp
```diff
@@ -22,12 +22,64 @@
...
+
+  void OnTransformedFrame(
+      std::unique_ptr<webrtc::TransformableFrameInterface> frame) override {
+    second_->Transform(std::move(frame));
+  }
+
+ private:
+  rtc::scoped_refptr<webrtc::FrameTransformerInterface> first_;
+  rtc::scoped_refptr<webrtc::FrameTransformerInterface> second_;
+  rtc::scoped_refptr<webrtc::TransformedFrameCallback> callback_;
```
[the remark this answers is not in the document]
chenosaurus: removed the unused ref

theomonnom:
Is there a way to ensure that any client can still decode videos, even if they don’t implement user timestamps?

libwebrtc/src/native/packet_trailer.rs
```diff
@@ -0,0 +1,147 @@
+// Copyright 2025 LiveKit, Inc.
```
xianshijing-lk: nit, 2026 ?

libwebrtc/src/native/packet_trailer.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn enabled(&self) -> bool {
+        self.sys_handle.enabled()
+    }
+
+    /// Lookup the frame metadata for a given RTP timestamp (receiver side).
+    /// Returns `Some((user_timestamp_us, frame_id))` if found, `None` otherwise.
+    /// The entry is removed from the map after a successful lookup.
+    pub fn lookup_frame_metadata(&self, rtp_timestamp: u32) -> Option<(i64, u32)> {
+        let ts = self.sys_handle.lookup_timestamp(rtp_timestamp);
+        if ts >= 0 {
```
xianshijing-lk: maybe simply this block to something like 
```
if ts < 0 {
    return None;
}

let frame_id = self.sys_handle.last_lookup_frame_id();
if ts > 2_000_000_000_000_000 {
    log::warn!(
         "[PacketTrailer-FFI] C++ returned bad ts={} (0x{:016x}) fid={} rtp_ts={}",
          ts, ts, frame_id, rtp_timestamp
    );
}

Some((ts, frame_id))
```

.changeset/add_support_for_frame_level_packet_trailer.md
```diff
@@ -0,0 +1,19 @@
...
+livekit-protocol: minor
+livekit-api: minor
+livekit-wakeword: no changelog additions
+soxr-sys: no changelog additions
+webrtc-sys-build: no changelog additions
+webrtc-sys: minor
+livekit-ffi: minor
+yuv-sys: no changelog additions
+libwebrtc: minor
+imgproc: no changelog additions
```
ladvoc: issue: I think "no changelog additions" still triggers a release for those packages. You can just remove the lines for unaffected packages.

libwebrtc/src/video_frame.rs
```diff
@@ -59,9 +59,19 @@ where
 {
     pub rotation: VideoRotation,
     pub timestamp_us: i64, // When the frame was captured in microseconds
+    /// Optional user timestamp in microseconds, if available.
+    pub user_timestamp_us: Option<i64>,
```
ladvoc: comment: The equivalent field for data tracks is unsigned.

libwebrtc/src/video_frame.rs
```diff
@@ -59,9 +59,19 @@ where
...
+    /// Optional user-supplied frame identifier.
+    pub frame_id: Option<u32>,
```
ladvoc: suggestion: Adding new fields to an exhaustive struct is a breaking change, so this should be mentioned in the changelog.

livekit/src/room/options.rs
```diff
@@ -84,10 +85,14 @@ pub struct TrackPublishOptions {
...
     pub red: bool,
     pub simulcast: bool,
+    /// Custom simulcast layer presets (low, mid). When set, these override the
+    /// SDK's built-in defaults which reduce fps on lower layers.
+    pub simulcast_layers: Option<Vec<VideoPreset>>,
     // pub name: String,
     pub source: TrackSource,
     pub stream: String,
     pub preconnect_buffer: bool,
+    pub packet_trailer_features: Vec<PacketTrailerFeature>,
```
ladvoc: suggestion: Although we do expose them in some places, it is better to keep Protobuf types as implementation details. I would recommend a non-exhaustive struct with Boolean flags (also more natural for the user). For example:
```rust
#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
#[non_exhaustive]
pub struct PacketTrailerFeatures {
    pub user_timestamp: bool,
    pub frame_id: bool,
}

impl PacketTrailerFeatures {
    // Internal helper to convert to protocol-eqv.
    pub(crate) fn to_proto() -> Vec<proto::PacketTrailerFeature> {
        todo!()
    }
}
```
Adding new fields here in the future would also be a non-breaking change since the struct is declared _non_exhaustive_ and implements _Default_. Alternatively, this could use a builder pattern, but that might be overkill.

ladvoc (commented):
Looking good so far, some initial comments.

livekit/src/room/options.rs
chenosaurus: good call, updated.

.changeset/add_support_for_frame_level_packet_trailer.md
```diff
@@ -0,0 +1,16 @@
+---
+livekit: minor
+livekit-protocol: minor
+livekit-api: minor
+webrtc-sys: minor
+livekit-ffi: minor
+libwebrtc: minor
+imgproc: no changelog additions
```
ladvoc: nitpick: _imgproc_ shouldn't get a release for this change.

libwebrtc/src/video_frame.rs
```diff
@@ -52,16 +52,36 @@ pub enum VideoBufferType {
...
 }
 
+/// Metadata carried alongside a video frame via the packet trailer mechanism.
+///
+/// Each field corresponds to an independently negotiable packet trailer feature
+/// (`PTF_USER_TIMESTAMP`, `PTF_FRAME_ID`), so individual fields are `Option`.
+#[derive(Debug, Clone, Copy)]
+pub struct FrameMetadata {
+    /// Wall-clock capture time in microseconds, when `PTF_USER_TIMESTAMP` is enabled.
+    pub user_timestamp_us: Option<u64>,
```
ladvoc: question: The equivalent field for data tracks doesn't prescribe a unit and is just called (`user_timestamp`) since the way it is used is determined by the application. Maybe it should be the same here for consistency?

examples/local_video/src/subscriber.rs
```diff
@@ -236,13 +295,11 @@ async fn handle_track_subscribed(
         sc.active_quality = None;
         sc.publication = Some(publication.clone());
     }
-    let simulcast2 = simulcast.clone();
     std::thread::spawn(move || {
```
ladvoc: issue: _NativeVideoStream_ implements _Stream_ and is intended to be consumed in an async context. The frame sink loop should be a _tokio::spawn_ task rather than a dedicated OS thread that immediately calls _rt.block_on()_.

livekit/src/room/participant/local_participant.rs
```diff
@@ -361,6 +369,23 @@ impl LocalParticipant {
 
         track.set_transceiver(Some(transceiver));
 
+        if !options.packet_trailer_features.is_empty() {
+            if let LocalTrack::Video(video_track) = &track {
+                log::info!("packet_trailer enabled for local video track {}", publication.sid(),);
```
ladvoc, line 374: suggestion: Maybe this should be at debug level?

livekit/src/room/track/local_video_track.rs
```diff
@@ -123,6 +126,18 @@ impl LocalVideoTrack {
         self.source.clone()
     }
 
+    /// Returns the packet trailer handler associated with this track, if any.
+    /// When present on the sender side, callers can store per-frame user
+    /// timestamps which will be embedded into encoded frames.
+    pub fn packet_trailer_handler(&self) -> Option<PacketTrailerHandler> {
```
ladvoc: question: Shouldn't this also be _pub(crate)_?

ladvoc (approved):
Generally looks good to me! I would consider adding two E2E tests: one that validates frame metadata arrives intact and another to ensure it doesn't break with end-to-end encryption enabled. This will give us confidence future changes don't break this feature.

libwebrtc/src/video_frame.rs
chenosaurus: I can update this since now it's in it's own metadata struct.  Originally it was a property on the VideoFrame struct and I added the `_us` to be consistent with `timestamp_us`

examples/local_video/src/publisher.rs
```diff
@@ -230,6 +381,7 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
     let mut frame = VideoFrame {
         rotation: VideoRotation::VideoRotation0,
         timestamp_us: 0,
+        frame_metadata: None,
```
ladvoc, line 384: suggestion(non-blocking): Sorry to bring this up late in the review process, but taking another look at this, I'm wondering if these fields should be inlined into `VideoFrame` instead. `rotation` and `timestamp_us` are essentially metadata too, so it might be awkward that these newer metadata fields are nested in `frame_metadata`.

ladvoc (approved):

PR #996, opened 2026-04-07
Send publisher offer with join request to accelerate connection

cnderrauber:
Align with js sdk: https://github.com/livekit/client-sdk-js/pull/1846

Test result (100ms rtt, oss sfu with ice-lite, 10 rounds)

| Test Case | Avg | Min | Max |
|--------|--------|--------|--------|
| main-v0 (dual PC) | 925.6ms | 894.6ms | 962.0ms |
| main-v1 (single PC, no early offer) | 926.6ms | 884.9ms | 977.4ms |
| publisher_offer-v1 (single PC, early offer) | 824.2ms | 740.3ms | 948.9ms |


livekit-api/src/signal_client/mod.rs
```diff
@@ -715,13 +732,29 @@ fn create_join_request_param(
     // Serialize JoinRequest to bytes
     let join_request_bytes = join_request.encode_to_vec();
 
-    // Create WrappedJoinRequest (JS doesn't explicitly set compression, so default is NONE)
+    // Use gzip compression when publisher offer is included (SDP makes payload large)
```
xianshijing-lk, line 735: does gzip introduce meaningful improvement here ? I thought it would be useful only for large data, like > 1KB

cnderrauber: Numbers from a local run with the current early offer (data-channel m-section only): JoinRequest encodes to ~524 B, gzip to ~430 B — ~17% saved (90 B). You are right, it is not helpful in this size.
Use gzip because the publisher offer in the join was designed to include pre-allocated recvonly transceivers, where the SDP balloons to ~3 KB (7 m-lines including video codecs). But I found rust-sdk(libwebrtc) has a problem reusing the pre-allocated transceiver which works well in js-sdk(browser), so I commented it out. Left gzip as the CPU cost is negligible, and we would use it in the future to include more bytes.

xianshijing-lk (approved):
lgtm

PR #998, opened 2026-04-08
Revert patch fix versioning for consistent ffi-versions

lukasIO:
knope bot appears to not run on the repo since we did this intermediate patch fix for the node bindings. 
running `knope --validate` showed that all versions have to be the same for a particular package, which makes sense.


MaxHeimbrock (approved):
LGTM

PR #1003, opened 2026-04-10
Add dynacast support

chenosaurus:
- properly handle `SubscribedQualityUpdate` events from SFU and update published layers.
- Add e2e tests for dynacast

Heres a video of dynacast working. We can see that if we get server commands down, we properly respond
https://github.com/user-attachments/assets/41bb2da5-c58d-47cb-8399-a767d987c353

### DIFF
--- a/examples/local_video/src/publisher.rs
+++ b/examples/local_video/src/publisher.rs
@@ -75,6 +75,10 @@ struct Args {
     /// Use H.265/HEVC encoding if supported (falls back to H.264 on failure)
     #[arg(long, default_value_t = false)]
     h265: bool,
+
+    /// Enable dynacast (pause unused simulcast layers based on subscriber demand)
+    #[arg(long, default_value_t = false)]
+    dynacast: bool,
 }
 
 fn list_cameras() -> Result<()> {
@@ -137,6 +141,7 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
     info!("Connecting to LiveKit room '{}' as '{}'...", args.room_name, args.identity);
     let mut room_options = RoomOptions::default();
     room_options.auto_subscribe = true;
+    room_options.dynacast = args.dynacast;
     let (room, _) = Room::connect(&url, &token, room_options).await?;
     let room = std::sync::Arc::new(room);
     info!("Connected: {} - {}", room.name(), room.sid().await);
@@ -419,11 +424,24 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
             let secs = last_fps_log.elapsed().as_secs_f64();
             let fps_est = frames as f64 / secs;
             let n = frames.max(1) as f64;
+            let layers = track.publishing_layers();
+            let layers_str = if layers.is_empty() {
+                "n/a".to_string()
+            } else {
+                layers
+                    .iter()
+                    .map(|(rid, quality, active)| {
+                        format!("{}({})={}", rid, quality, if *active { "ON" } else { "off" })
+                    })
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            };
             info!(
-                "Publishing video: {}x{}, ~{:.1} fps | avg ms: get {:.2}, decode {:.2}, convert {:.2}, capture {:.2}, sleep {:.2}, iter {:.2} | target {:.2}",
+                "Publishing video: {}x{}, ~{:.1} fps | layers: [{}] | avg ms: get {:.2}, decode {:.2}, convert {:.2}, capture {:.2}, sleep {:.2}, iter {:.2} | target {:.2}",
                 width,
                 height,
                 fps_est,
+                layers_str,
                 sum_get_ms / n,
                 sum_decode_ms / n,
                 sum_convert_ms / n,
--- a/examples/local_video/src/publisher.rs
+++ b/examples/local_video/src/publisher.rs
@@ -238,6 +238,10 @@ struct Args {
     #[arg(long, default_value_t = false)]
     attach_timestamp: bool,
 
+    /// Enable dynacast (pause unused simulcast layers based on subscriber demand)
+    #[arg(long, default_value_t = false)]
+    dynacast: bool,
+
     /// Burn the attached timestamp into each video frame; does nothing unless --attach-timestamp is also enabled
     #[arg(long, default_value_t = false)]
     burn_timestamp: bool,
@@ -924,8 +928,8 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
 
     info!("Connecting to LiveKit room '{}' as '{}'...", args.room_name, args.identity);
     let mut room_options = RoomOptions::default();
-    room_options.auto_subscribe = false;
-    room_options.dynacast = true;
+    room_options.auto_subscribe = true;
+    room_options.dynacast = args.dynacast;
 
     // Configure E2EE if an encryption key is provided
     if let Some(ref e2ee_key) = args.e2ee_key {
@@ -1253,6 +1257,7 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
                 let capture_task = tokio::spawn(run_capture_loop(
                     capture_config,
                     ctrl_c_received.clone(),
+                    track.clone(),
                     rtc_source,
                     video_input,
                     width,
@@ -1277,6 +1282,7 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
                 let capture_result = run_capture_loop(
                     capture_config,
                     ctrl_c_received,
+                    track,
                     rtc_source,
                     video_input,
                     width,
@@ -1297,6 +1303,7 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
 async fn run_capture_loop(
     config: CaptureConfig,
     ctrl_c_received: Arc<AtomicBool>,
+    track: LocalVideoTrack,
     rtc_source: NativeVideoSource,
     mut video_input: VideoInput,
     width: u32,
@@ -1646,11 +1653,29 @@ async fn run_capture_loop(
         if last_fps_log.elapsed() >= std::time::Duration::from_secs(2) {
             let secs = last_fps_log.elapsed().as_secs_f64();
             let fps_est = frames as f64 / secs;
+            let layers = track.publishing_layers();
+            let layers_str = if layers.is_empty() {
+                "n/a".to_string()
+            } else {
+                layers
+                    .iter()
+                    .map(|layer| {
+                        format!(
+                            "{}({})={}",
+                            layer.rid,
+                            layer.quality,
+                            if layer.active { "ON" } else { "off" }
+                        )
+                    })
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            };
             info!(
-                "Video status: {}x{} | ~{:.1} fps | target {:.2} ms",
+                "Video status: {}x{} | ~{:.1} fps | layers: [{}] | target {:.2} ms",
                 width,
                 height,
                 fps_est,
+                layers_str,
                 target.as_secs_f64() * 1000.0,
             );
             info!("{}", format_timing_line(&timings));
--- a/examples/local_video/src/publisher.rs
+++ b/examples/local_video/src/publisher.rs
@@ -1664,7 +1664,7 @@ async fn run_capture_loop(
                             "{}({})={}",
                             layer.rid,
                             layer.quality,
-                            if layer.active { "ON" } else { "off" }
+                            if layer.active { "on" } else { "off" }
                         )
                     })
                     .collect::<Vec<_>>()
--- a/libwebrtc/src/native/rtp_parameters.rs
+++ b/libwebrtc/src/native/rtp_parameters.rs
@@ -39,7 +39,13 @@ impl From<sys_rp::ffi::RtpParameters> for RtpParameters {
         Self {
             codecs: value.codecs.into_iter().map(Into::into).collect(),
             header_extensions: value.header_extensions.into_iter().map(Into::into).collect(),
+            encodings: value.encodings.into_iter().map(Into::into).collect(),
             rtcp: value.rtcp.into(),
+            transaction_id: value.transaction_id,
+            mid: value.mid,
+            has_degradation_preference: value.has_degradation_preference,
+            // Safety: DegradationPreference is #[repr(i32)]
+            degradation_preference: unsafe { std::mem::transmute(value.degradation_preference) },
         }
     }
 }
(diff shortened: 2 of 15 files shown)


livekit/src/room/track/local_video_track.rs
```diff
@@ -146,4 +146,88 @@ impl LocalVideoTrack {
     pub(crate) fn update_info(&self, info: proto::TrackInfo) {
         super::update_info(&self.inner, &Track::LocalVideo(self.clone()), info);
     }
+
+    /// Returns a snapshot of each simulcast layer's RID, quality label, and active state.
+    /// Useful for diagnostics / HUD display. Returns an empty vec if no transceiver is set.
```
xianshijing-lk: nit, I think it also returns empty vec if there is no encodings (transceiver) yet.

/// Returns an empty vec if no transceiver is set, or if the sender has no encodings yet.

livekit/src/room/track/local_video_track.rs
```diff
@@ -146,4 +146,88 @@ impl LocalVideoTrack {
...
+    pub fn publishing_layers(&self) -> Vec<(String, String, bool)> {
+        let Some(transceiver) = self.transceiver() else {
+            return Vec::new();
```
xianshijing-lk, line 339: I wonder if we should log a debug message here ? 

btw, how likely that publishing_layers will be triggered before we have a transceiver ?

livekit/src/room/track/local_video_track.rs
```diff
@@ -146,4 +146,88 @@ impl LocalVideoTrack {
...
+        };
+        let params = transceiver.sender().parameters();
```
xianshijing-lk, line 341: any chance that params.len is 0 ?

livekit/src/room/mod.rs
```diff
@@ -1814,6 +1817,94 @@ impl RoomSession {
...
+                })
+        } else {
+            let qs: Vec<String> = update
+                .subscribed_qualities
+                .iter()
+                .map(|q| {
+                    format!(
+                        "{:?}={}",
+                        proto::VideoQuality::try_from(q.quality)
+                            .unwrap_or(proto::VideoQuality::High),
```
xianshijing-lk: It looks like we have multiple places to default the quality to proto::VideoQuality::High when things are unset ? 

Should we make a flag or macro instead ? and please add comments to explain why proto::VideoQuality::High is preferred

livekit/src/room/track/local_video_track.rs
```diff
@@ -146,4 +146,88 @@ impl LocalVideoTrack {
...
+
+            if encoding.active != should_active {
+                changed = true;
+                encoding.active = should_active;
+            }
+        }
+
+        if !any_enabled {
+            for encoding in &mut params.encodings {
+                encoding.active = false;
```
xianshijing-lk: should you also set changed to true here ?

livekit/src/room/track/local_video_track.rs
```diff
@@ -146,4 +146,88 @@ impl LocalVideoTrack {
...
+            .iter()
+            .map(|e| {
+                let quality = crate::options::video_quality_for_rid(&e.rid)
+                    .unwrap_or(proto::VideoQuality::High);
+                let state = if e.active { "ON" } else { "off" };
+                format!("{}({:?})={}", e.rid, quality, state)
+            })
+            .collect();
+
+        if changed {
```
xianshijing-lk, line 401: [question](https://pullpo.io/cc?l=question) 

Nit, I wonder if we should print the logs out after sender.set_parameters(params), and only log it when the set_parameters is successfully ?

livekit/src/room/mod.rs
```diff
@@ -1814,6 +1817,94 @@ impl RoomSession {
...
+                log::debug!(
+                    "dynacast: track {} is not a local video track, ignoring quality update",
+                    track_sid
+                );
+                return;
+            }
+        };
+
+        let qualities: Vec<proto::SubscribedQuality> = if !update.subscribed_codecs.is_empty() {
+            let codec = publication.publish_options().video_codec.as_str().to_lowercase();
```
xianshijing-lk, line 1919: curiously, is publication.publish_options().video_codec the codec that being used ? or the codec that is preferred or requested ? 

I wonder if there is any chance that the actual codec is not publication.publish_options().video_codec ? in that case, there might be a mismatch

livekit/src/room/track/local_video_track.rs
```diff
@@ -146,4 +146,88 @@ impl LocalVideoTrack {
     pub(crate) fn update_info(&self, info: proto::TrackInfo) {
         super::update_info(&self.inner, &Track::LocalVideo(self.clone()), info);
     }
+
+    /// Returns a snapshot of each simulcast layer's RID, quality label, and active state.
+    /// Useful for diagnostics / HUD display. Returns an empty vec if no transceiver is set.
+    pub fn publishing_layers(&self) -> Vec<(String, String, bool)> {
```
ladvoc: nitpick: Defining a struct for this would be cleaner than returning a three-member tuple.

libwebrtc/src/native/rtp_parameters.rs
```diff
@@ -39,7 +39,13 @@ impl From<sys_rp::ffi::RtpParameters> for RtpParameters {
         Self {
             codecs: value.codecs.into_iter().map(Into::into).collect(),
             header_extensions: value.header_extensions.into_iter().map(Into::into).collect(),
+            encodings: value.encodings.into_iter().map(Into::into).collect(),
             rtcp: value.rtcp.into(),
+            transaction_id: value.transaction_id,
+            mid: value.mid,
+            has_degradation_preference: value.has_degradation_preference,
+            // Safety: DegradationPreference is #[repr(i32)]
+            degradation_preference: unsafe { std::mem::transmute(value.degradation_preference) },
```
ladvoc: suggestion: This can be done without unsafe using `value.degradation_preference.repr`. This applies in a few other places too.

livekit/src/room/track/local_video_track.rs
stephen-derosa: Good catch -- this can also be empty when the sender has no encodings yet, not only when there is no transceiver. I’ll update the doc comment.

livekit/src/room/track/local_video_track.rs
stephen-derosa: Yeah -- params.encodings can be empty. I added an early return / debug for empty encodings.

livekit/src/room/mod.rs
stephen-derosa: Agreed., should be centralized. Added a small helper/comment.

livekit/src/room/track/local_video_track.rs
stephen-derosa: The first loop already sets changed whenever an active encoding is turned off, so we dont need to set again. That said, the extra loop is confusing/redundant -- I’ll clean it up.

livekit/src/room/track/local_video_track.rs
stephen-derosa: yeah for sure, ill add a log

livekit/src/room/mod.rs
stephen-derosa: publish_options().video_codec is the requested/adver­tised codec AND its used when setting codec preferences and populating simulcast_codecs. It should match the codec key the sfu sends back. If it somehow does not, the current fallback to the first subscribed codec handles it with a warning. I’ll add a comment.

livekit/src/room/track/local_video_track.rs
stephen-derosa: Agreed -- adding

libwebrtc/src/native/rtp_parameters.rs
stephen-derosa: nice, done

livekit/src/room/track/local_video_track.rs
stephen-derosa: I wasnt exacty sure, but an agent gave the following cases:

Scenario | Likelihood
-- | --
Track created but not published yet | Expected — always None until publish finishes
After successful publish_track() | Very unlikely — transceiver is set before the call returns
After unpublish_track() | Guaranteed if you still call publishing_layers() on that handle
Publish fails after add_track but before create_sender | Rare edge case — track may have a SID but no transceiver
Race during publish from another thread | Extremely unlikely — set_transceiver happens synchronously right after create_sender in the same task

stephen-derosa: i added a debug log

livekit/src/room/track/local_video_track.rs
```diff
@@ -38,6 +38,13 @@ pub use libwebrtc::native::packet_trailer::{PublishTimingEvent, PublishTimingSta
 
 const PUBLISH_TIMING_BUFFER: usize = 256;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PublishingLayer {
+    pub rid: String,
+    pub quality: String,
```
ladvoc: issue: Quality should be defined as an enum matching the Protobuf cases.

libwebrtc/src/rtp_parameters.rs
```diff
@@ -33,7 +33,22 @@ pub struct RtpHeaderExtensionParameters {
 pub struct RtpParameters {
     pub codecs: Vec<RtpCodecParameters>,
     pub header_extensions: Vec<RtpHeaderExtensionParameters>,
+    pub encodings: Vec<RtpEncodingParameters>,
     pub rtcp: RtcpParameters,
+    /// Opaque token used by WebRTC to pair getParameters/setParameters calls.
+    /// Must be preserved when round-tripping through set_parameters().
+    pub(crate) transaction_id: String,
```
ladvoc, line 40: issue: Adding new fields to a non-exhaustive struct is technically a breaking change, and mixing `pub(crate)` fields in means the consumer can no longer construct with struct literal syntax (`RtpParameters { ... }`). The changeset should reflect there is a minor breaking change for this crate.

livekit/src/room/track/local_video_track.rs
stephen-derosa: ty. made `PublishingLayer.quality` -> `PublishingLayerQuality`, an enum matching the proto

stephen-derosa: also added from impls:
`impl From<proto::VideoQuality> for PublishingLayerQuality {`
`impl From<PublishingLayerQuality> for proto::VideoQuality {`
to go proto <>  sdk enum

libwebrtc/src/rtp_parameters.rs
stephen-derosa: ty, updated

livekit/tests/dynacast_test.rs
```diff
@@ -0,0 +1,462 @@
...
+                layers
+            ));
+        }
+        time::sleep(Duration::from_millis(250)).await;
+    }
+}
+
+/// Waits for the publisher to expose `expected_count` local video tracks.
+#[cfg(feature = "__lk-e2e-test")]
+async fn wait_for_publisher_video_tracks(
```
ladvoc: nitpick: These helpers can be made event driven to avoid polling (see _wait_for_remote_track_ in _data_track_test.rs_).

livekit/src/room/track/local_video_track.rs
```diff
@@ -266,4 +318,81 @@ impl LocalVideoTrack {
...
+                format!("{}({:?})={}", e.rid, quality, state)
+            })
+            .collect();
+
+        sender
+            .set_parameters(params)
+            .map_err(|e| RoomError::Internal(format!("failed to set sender parameters: {}", e)))?;
+
+        if changed {
+            log::info!("dynacast: layers changed -> [{}]", layers.join(", "));
```
ladvoc: suggestion: Consider making this debug level.

livekit/src/room/options.rs
```diff
@@ -347,14 +347,27 @@ pub fn spatial_layers_from_scalability_mode(mode: &str) -> u32 {
     1
 }
 
+pub(crate) fn default_video_quality() -> proto::VideoQuality {
```
ladvoc: suggestion: Consider defining this as a constant.

examples/local_video/src/publisher.rs
```diff
@@ -1646,11 +1653,29 @@ async fn run_capture_loop(
...
+                "n/a".to_string()
+            } else {
+                layers
+                    .iter()
+                    .map(|layer| {
+                        format!(
+                            "{}({})={}",
+                            layer.rid,
+                            layer.quality,
+                            if layer.active { "ON" } else { "off" }
```
ladvoc: nitpick: Inconsistent casing.

livekit/src/room/track/local_video_track.rs
```diff
@@ -38,6 +38,58 @@ pub use libwebrtc::native::packet_trailer::{PublishTimingEvent, PublishTimingSta
...
+            PublishingLayerQuality::Medium => Self::Medium,
+            PublishingLayerQuality::High => Self::High,
+            PublishingLayerQuality::Off => Self::Off,
+        }
+    }
+}
+
+impl Display for PublishingLayerQuality {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
```
ladvoc: suggestion: Generally, it isn't recommended to rely on debug to implement display. I would recommend writing it out explicitly:
```rust
impl Display for PublishingLayerQuality {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        match self {
            Self::Low => write!(f, "low"),
            Self::Medium => write!(f, "medium"),
            Self::High => write!(f, "high"),
            Self::Off => write!(f, "off")
        }
    }
}
```
This also gives us lowercase display strings which is consistent with other types in this codebase.

Alternatively, we could use [`strum::Display`](https://docs.rs/strum/latest/strum/derive.Display.html) to avoid the boilerplate, but I'd say we should adopt that across the codebase in a separate PR if we are going to pull in the new dep.

ladvoc:
One more thing: the README should be updated to check the "Dynacast" box.

livekit/src/room/track/local_video_track.rs
```diff
@@ -38,6 +38,58 @@ pub use libwebrtc::native::packet_trailer::{PublishTimingEvent, PublishTimingSta
...
+/// Video quality for a publishing layer.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PublishingLayerQuality {
+    /// Low video quality.
+    Low,
+    /// Medium video quality.
+    Medium,
+    /// High video quality.
+    High,
+    /// Disabled video quality.
```
xianshijing-lk: what does it mean by Disabled video quality (Off) ? or it means auto ? could you please update the comment to explain it better?

xianshijing-lk: and is it used at all ?

xianshijing-lk (approved):
some questions / nits. lgtm

livekit/src/room/track/local_video_track.rs
stephen-derosa: done

examples/local_video/src/publisher.rs
stephen-derosa: fixed

livekit/src/room/options.rs
stephen-derosa: done

livekit/src/room/track/local_video_track.rs
stephen-derosa: ty, done

livekit/tests/dynacast_test.rs
stephen-derosa: great call, done

livekit/src/room/track/local_video_track.rs
stephen-derosa: it just exists to align with the `VideoQuality` proto, it is never actually used though. I added a comment:
```
    /// Mirrors `VideoQuality::OFF` from the LiveKit protocol. In practice this
    /// is not emitted by [`LocalVideoTrack::publishing_layers`] (a disabled
    /// layer is instead reflected by [`PublishingLayer::active`] being
    /// `false`); it exists to fully model the protocol enum and for conversions
    /// to/from it.
```

ladvoc (approved):

PR #1013, opened 2026-04-13
Support for large RPC messages using data streams

1egoman:
This is port of the identically named client-sdk-js pull request https://github.com/livekit/client-sdk-js/pull/1832, adding support for RPC protocol version 2, via feeding [the spec](https://github.com/livekit/client-sdk-js/blob/f69d8a05a5e6db5bdbe984766be1faa5ade6911e/RPC_SPEC.md) to a LLM.

> [!WARNING]
> This pull request was LLM generated and has only been lightly reviewed by a human. The author has tested this and confirms it works in the happy path, but no other validation has been done.
> 
> A more thorough review of this needs to occur before it could be merged. The plan is to hand this off to a rust sdk maintainer to get it fully across the line.


livekit/src/room/rpc/tests.rs
```diff
@@ -0,0 +1,873 @@
...
+async fn test_v2_v2_caller_happy_path_short() {
+    let client = Arc::new(RpcClientManager::new());
+    let transport = Arc::new(
+        MockTransport::new()
+            .with_remote_protocol("dest", CLIENT_PROTOCOL_DATA_STREAM_RPC),
+    );
+
+    let handle = spawn_perform_rpc(
+        client.clone(),
+        transport.clone(),
```
1egoman, line 241: Note to any reviewers: The tests here don't use the larger e2e test infrastructure in this repository purposely, because some test need to fully emulate the behavior  of an old client (ie, doing things like always sending `RpcRequest` / `RpcResponse` even with a newer `client_protocol`, etc).

I am not sure if this is a good idea or not. Halfway through the process, the LLM basically said it couldn't generate the tests with the e2e infrastructure and proposed this as an alternative so I let it go for it.

livekit-api/src/signal_client/mod.rs
```diff
@@ -83,11 +89,19 @@ pub enum SignalError {
 pub struct SignalSdkOptions {
     pub sdk: String,
     pub sdk_version: Option<String>,
+    /// Override the client_protocol advertised during join. If `None`, falls back
+    /// to `CLIENT_PROTOCOL_DEFAULT` (0). The SDK's default constructor sets this
+    /// to `CLIENT_PROTOCOL_DATA_STREAM_RPC` (1) to advertise data-stream RPC support.
+    pub client_protocol: Option<i32>,
```
xianshijing-lk: Nit
why not just make it a i32 and default to 0 , rather than optional

livekit/src/room/rpc/client.rs
```diff
@@ -0,0 +1,422 @@
+// Copyright 2025 LiveKit, Inc.
```
xianshijing-lk: nit, 2026

livekit/src/room/rpc/client.rs
```diff
@@ -0,0 +1,422 @@
...
+        let send_result = if use_v2 {
+            self.send_v2_request(
+                transport,
+                &data.destination_identity,
+                &id,
+                &data.method,
+                &data.payload,
+                effective_timeout,
+            )
+            .await
```
xianshijing-lk, line 104: [question](https://pullpo.io/cc?l=question) 

do we need .map_err(|e| { for v2 ?

livekit/src/room/rpc/client.rs
```diff
@@ -0,0 +1,422 @@
...
+            }
+            Ok(Ok(payload)) => {
+                // Successful response
+                Ok(payload)
+            }
+        }
+    }
+
+    /// Send an RPC request as a v2 text data stream.
+    async fn send_v2_request(
```
xianshijing-lk, line 196: [suggestion](https://pullpo.io/cc?l=suggestion) 

curiously, any chance this send_v2_request can be called in parallel by multiple threads ? 

for instance, if I have two threads, and both call perform_rpc at the same time. 
if ATTR_REQUEST_ID / ATTR_METHOD are the same, will there be any conflict ?

livekit/src/room/rpc/mod.rs
```diff
@@ -0,0 +1,263 @@
+// Copyright 2025 LiveKit, Inc.
```
xianshijing-lk: nit, 2026

livekit/src/room/rpc/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+pub use server::RpcServerManager;
+
+use crate::data_stream::{StreamResult, StreamTextOptions, TextStreamInfo};
+use crate::room::id::ParticipantIdentity;
+use livekit_protocol::RpcError as RpcError_Proto;
+use std::{error::Error, fmt::Display, future::Future, time::Duration};
+
+// RPC protocol version constants (distinct from client_protocol; this is the
+// version field on RpcRequest / v2 stream attributes).
+pub(crate) const RPC_VERSION_V1: u32 = 1;
```
xianshijing-lk, line 31: nit, use bool ?

livekit/src/room/rpc/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    fn server_version(&self) -> Option<String> {
+        self.0
+            .rtc_engine
+            .session()
+            .signal_client()
+            .join_response()
+            .server_info
+            .and_then(|info| {
+                if info.version.is_empty() {
```
xianshijing-lk: nit 
fn server_version(&self) -> Option<String> {
    self.0
        .rtc_engine
        .session()
        .signal_client()
        .join_response()
        .server_info
        .and_then(|info| (!info.version.is_empty()).then(|| info.version))
}

livekit/src/room/rpc/server.rs
```diff
@@ -0,0 +1,334 @@
+// Copyright 2025 LiveKit, Inc.
```
xianshijing-lk: nit, 2026

livekit/src/room/rpc/server.rs
```diff
@@ -0,0 +1,334 @@
...
+        + Send
+        + Sync,
+>;
+
+/// Manages incoming RPC requests (handler/server side).
+///
+/// Stores registered method handlers and dispatches incoming requests
+/// to the appropriate handler. Handles both v1 packet and v2 data stream
+/// request formats.
+pub struct RpcServerManager {
```
ladvoc: nitpick: Maybe this could just be called `RpcManager`?

livekit/src/room/rpc/server.rs
```diff
@@ -0,0 +1,334 @@
...
+
+    pub(crate) fn get_handler(&self, method: &str) -> Option<RpcHandlerFn> {
+        self.handlers.lock().get(method).cloned()
+    }
+
+    /// Handle an incoming v1 RPC request (received as a DataPacket).
+    ///
+    /// Sends ACK, invokes the registered handler, and sends the response
+    /// as a v1 RPC response packet.
+    pub(crate) async fn handle_request(
```
ladvoc: suggestion: Although this pattern is used in other places in this codebase, ideally we shouldn't have functions with this many arguments; I would suggest breaking this out into an internal struct.

livekit/src/room/rpc/server.rs
```diff
@@ -0,0 +1,334 @@
...
+pub struct RpcServerManager {
+    handlers: Mutex<HashMap<String, RpcHandlerFn>>,
+}
+
+impl RpcServerManager {
+    pub fn new() -> Self {
+        Self { handlers: Mutex::new(HashMap::new()) }
+    }
+
+    pub fn register_method(
```
ladvoc, line 56: suggestion: I have a WIP PR (#858) to allow RPC method registration prior to room connection. I haven't merged this yet since the FFI portion is incomplete and requires some larger architectural changes, but the core change in the `livekit` crate is finished and could potentially be integrated here.

livekit/src/room/rpc/tests.rs
ladvoc: There's also been some divergence for the single peer connection tests; in a follow-up, we should ensure all the E2EE tests are using common testing utilities (and make sure they can support any configuration needed by the test), but I would say for now it's better to include the tests even if they use different infrastructure.

livekit/src/room/rpc/mod.rs
1egoman: This should be able to able to be incremented beyond two values, so I think keeping it some kind of uint makes sense.

reenboog:
Postponing my review until 1832.

livekit/src/room/rpc/client.rs
1egoman: I don't think this is needed - `send_v2_request`'s returned error type is `RpcError`, and `perform_rpc` (the function this is in) returns `Result<String, RpcError>`.

livekit/src/room/rpc/client.rs
1egoman: `ATTR_REQUEST_ID` should be unique for each rpc call - if it isn't, then yes two interleaved rpc requests will interfere with each other.

This should always be unique in practice for the way that this `send_v2_request` function is being called right now, [since `id` is a random uuid](https://github.com/livekit/rust-sdks/blob/d37f3ae95c48c0f46ebe8c761dfa72e4e84bae65/livekit/src/room/rpc/client.rs#L87) - this code was already built this way for RPC v1 so I opted to leave it alone. If you can think of a different way to abstract this (maybe put the id generation within `send_v2_request` / `publish_rpc_request`?) I could do that instead.

livekit/src/room/rpc/server.rs
1egoman: Nice! There's enough conflicts on that linked pull request where I think it probably makes more sense to integrate these as part of inevitably fixing the merge conflicts when that pull request is rebased.

livekit/src/room/rpc/client.rs
xianshijing-lk: Yes, I think it will be great if abstract them into a helper function and make it explicit that some those params are unique, that will be very helpful.

xianshijing-lk:
I think the code in general looks clean and easy to follow for me. what is the plan of landing it ?

livekit-api/src/signal_client/mod.rs
1egoman: Looking at this again, I don't really understand why this even has to be a struct member. I think it should match the js implementation and just be a constant called something like `ADVERTISED_CLIENT_PROTOCOL`. Updated to use this constant approach instead in 4e638072.

1egoman:
@reenboog FWIW, the plan was to merge this and https://github.com/livekit/client-sdk-js/pull/1832 at the same time - I don't want to get into the situation where the web version gets merged and this one gets blocked for a while and then it's a web only feature. Also, holding off on merging the web one means we can make protocol updates to v2 which are non backwards compatible during the review, if required. So, feel free to take a look at this now.

livekit/src/room/rpc/mod.rs
1egoman: Huh, today I learned that `.then` was a thing, I never had come across that before. Cool, makes sense, updated.

livekit/src/room/rpc/server.rs
1egoman: Good point - done in cc75a94.

livekit/src/room/rpc/client.rs
1egoman: I went down the road of trying to reabstract this, but it turns out to be fairly hard - the id is needed to append entries into `self.pending_acks` / `self.pending_responses`, and if you trace down the dependencies of those calls you end up with like 80% of the `perform_rpc` method broken out into another function and it has basically the same problem.

I think one thing I can do to make it a little better though is to rename `send_v2_request` to something like `send_v2_request_payload` or similar to imply it's just doing the sending, not actually crafting the request. Also the v1 version of the operation should be a method, not a standalone function (and should follow the same naming pattern). I think those things together should help at least a little...

If you have a better suggestion feel free to surface it!

xianshijing-lk: then lets keep the way as is. Thanks for the exploration.

1egoman:
In regards to all the ci failures - this looks to be a github problem, NOT a problem with this branch. See here for another example of this `arduino/setup-protoc@a8b67ba40b37d35169e222f3bb352603327985b6` action being unable to be downloaded: https://github.com/livekit/rust-sdks/pull/1049

I'll rerun these in a few hours and hopefully this is transient / that will fix it.

xianshijing-lk (approved):
lgtm

1egoman:
Heads up that in 7bf4818a0391ce71cb26182df777bbfd8a34c606 I added back in some unused structs that were accidentally removed from the public api interface - `RpcRequest`, `RpcResponse`, and `RpcAck`. These are unrelated to the identically named `livekit-protocol` structs, and I can't really figure out what they are supposed to be used for, in `main` they also seem to be unused...

So, I added them back in with a deprecation - so builds which did expect these to exist won't break.

ladvoc:
> Heads up that in [7bf4818](https://github.com/livekit/rust-sdks/commit/7bf4818a0391ce71cb26182df777bbfd8a34c606) I added back in some unused structs that were accidentally removed from the public api interface - `RpcRequest`, `RpcResponse`, and `RpcAck`. These are unrelated to the identically named `livekiit-protocol` structs, and I can't really figure out what they are supposed to be used for, in `main` they also seem to be unused...

Yeah, these should not be public—depreciating for now is the right call I think.

PR #1016, opened 2026-04-15
Proposed fix for tokio runtime missing when called from Unity .NET GC

MaxHeimbrock:
First commit introduces a new test that repros the crash I see in Unity and tests fails: 
https://github.com/livekit/rust-sdks/actions/runs/24567408053
With second commit the bug is fixed and tests are green again, as seen in pipeline now.


xianshijing-lk (approved):

xianshijing-lk:
lgtm

lukasIO (approved):

PR #1019, opened 2026-04-15
TEL-464: reduce redundant resampling in audio filter

hechen-eng:
## Problem

When an audio filter (Krisp NC or ai-coustics) is active, audio gets resampled multiple times unnecessarily. The root cause is that the `livekit-ffi` audio filter plugin interface only supports a single `sample_rate`, used for both input and output. The WebRTC sink resamples audio to the target rate *before* passing it to the filter, and the filter may resample again internally to match its model's native rate.

### Current pipeline (agents, default `sample_rate=24000`)

```
WebRTC track (48000 Hz)
  → WebRTC native audio sink resamples (48000 → 24000)
  → Audio filter receives frames at 24000
  → Filter internally resamples to model rate and back (24000 → model_rate → 24000)
  → Agents receive frames at 24000
```

Krisp model native rates (per [Krisp SDK model selection guide](https://sdk-docs.krisp.ai/docs/krisp-audio-sdk-model-selection-guide)):

| Model | File | Native rate |
|-------|------|-------------|
| NC | `c8.f.s.026300-1.0.0_3.1.kef` | 32kHz |
| BVC | `hs.c6.f.m.75df8f.kef` | 32kHz |
| BVC Telephony | `inb.bvc.hs.c6.w.s.23cdb3.kef` | 16kHz |

Current resampling with agents default `sample_rate=24000`:

| Model | Current pipeline | After fix |
|-------|-----------------|-----------|
| Krisp NC (32kHz) | 48→24→32→24 (3 steps) | 48→32→24 (2 steps) |
| Krisp BVC (32kHz) | 48→24→32→24 (3 steps) | 48→32→24 (2 steps) |
| Krisp BVC Telephony (16kHz) | 48→24→16→24 (3 steps) | 48→16→24 (2 steps) |
| ai-coustics QuailL (16kHz) | 48→24→16→24 (3 steps) | no change (v1 fallback) |

ai-coustics model rate per `resolve_for_sample_rate` and [ai-coustics docs](https://docs.ai-coustics.com/guides/models).

### Ideal pipeline

```
WebRTC track (48000 Hz)
  → Audio filter receives frames at 48000, outputs at 24000
  → Filter internally: 48000 → model_rate → 24000
  → Agents receive frames at 24000
```

The WebRTC resampling step is eliminated. The filter handles rate conversion as part of its processing.

### Same problem existed in SIP (https://github.com/livekit/cloud-sip/pull/337)

This is the same issue that was fixed in `cloud-sip` + `sip`. The SIP pipeline had:
```
audioIn(A) → external resampler(A→B) → Krisp(B→B) → downstream(B)
```
Fixed to:
```
audioIn(A) → Krisp(input=A, output=B) → downstream(B)
```
The Krisp C SDK already supports separate input/output rates via `krispAudioNcCreateSession(inputSampleRate, outputSampleRate, ...)`.

## Solution

Add v2 plugin ABI symbols (`audio_filter_create_v2`, `audio_filter_process_int16_v2`, `audio_filter_process_float_v2`) that accept separate input and output sample rates. The FFI server probes for v2 symbols at plugin load time. When present, it creates the WebRTC sink at 48000 Hz (WebRTC's internal pipeline rate) so audio passes through without resampling, and the filter handles the rate conversion. When absent (v1 plugins), behavior is unchanged.

## Further Work: ai-coustics

ai-coustics currently processes audio in-place at a single rate (`process(&mut frame)`). Their SDK does not support separate input and output sample rates, so ai-coustics cannot take advantage of the v2 plugin ABI today. The FFI server falls back to v1 behavior (WebRTC resamples first, then ai-coustics processes at the resampled rate, potentially resampling again internally to the model's native 16kHz).

Two options to eliminate the redundant resampling for ai-coustics:

1. **Request ai-coustics to support separate input/output sample rates** in their SDK (similar to how Krisp's `krispAudioNcCreateSession` accepts both). The ai-coustics plugin would then export v2 symbols and the fix works the same way as Krisp.

2. **Handle the resampling ourselves in the plugin wrapper**. Resample from the input rate (48kHz) to the model's native rate (e.g., 16kHz for QuailL) before calling `process()`, then resample the output to the desired output rate (24kHz). This avoids the unnecessary 48→24 step in WebRTC while still using the existing in-place ai-coustics API. The trade-off is that we own the resampling logic and need to keep it in sync with whichever model is selected.

### DIFF
--- a/livekit-ffi/src/server/audio_stream.rs
+++ b/livekit-ffi/src/server/audio_stream.rs
@@ -109,14 +109,12 @@ impl FfiAudioStream {
                 // When the audio filter supports separate rates (v2), create
                 // the WebRTC sink at 48000 (WebRTC's internal pipeline rate)
                 // to skip resampling — the filter handles rate conversion.
-                let input_sample_rate = if audio_filter
-                    .as_ref()
-                    .is_some_and(|f| f.supports_separate_rates())
-                {
-                    48000
-                } else {
-                    output_sample_rate
-                };
+                let input_sample_rate =
+                    if audio_filter.as_ref().is_some_and(|f| f.supports_separate_rates()) {
+                        48000
+                    } else {
+                        output_sample_rate
+                    };
 
                 let native_stream = NativeAudioStream::with_options(
                     rtc_track,
@@ -289,14 +287,12 @@ impl FfiAudioStream {
                     }
                 });
 
-                let input_sample_rate = if filter
-                    .as_ref()
-                    .is_some_and(|f| f.supports_separate_rates())
-                {
-                    48000
-                } else {
-                    output_sample_rate
-                };
+                let input_sample_rate =
+                    if filter.as_ref().is_some_and(|f| f.supports_separate_rates()) {
+                        48000
+                    } else {
+                        output_sample_rate
+                    };
 
                 let (mut audio_filter_session, info) = match &filter {
                     Some(filter) => match &request.audio_filter_options {
--- a/livekit-ffi/src/server/audio_stream.rs
+++ b/livekit-ffi/src/server/audio_stream.rs
@@ -107,11 +107,11 @@ impl FfiAudioStream {
                 };
 
                 // When the audio filter supports separate rates (v2), create
-                // the WebRTC sink at 48000 (WebRTC's internal pipeline rate)
-                // to skip resampling — the filter handles rate conversion.
+                // the WebRTC sink at the codec's native rate to skip
+                // resampling — the filter handles rate conversion.
                 let input_sample_rate =
                     if audio_filter.as_ref().is_some_and(|f| f.supports_separate_rates()) {
-                        48000
+                        ffi_track.track.codec_clock_rate().unwrap_or(48000)
                     } else {
                         output_sample_rate
                     };
@@ -289,7 +289,7 @@ impl FfiAudioStream {
 
                 let input_sample_rate =
                     if filter.as_ref().is_some_and(|f| f.supports_separate_rates()) {
-                        48000
+                        track.codec_clock_rate().unwrap_or(48000) as i32
                     } else {
                         output_sample_rate
                     };
--- a/livekit-ffi/src/server/audio_stream.rs
+++ b/livekit-ffi/src/server/audio_stream.rs
@@ -98,7 +98,7 @@ impl FfiAudioStream {
             #[cfg(not(target_arch = "wasm32"))]
             proto::AudioStreamType::AudioStreamNative => {
                 let audio_stream = Self { handle_id, stream_type, self_dropped_tx };
-                let sample_rate = new_stream.sample_rate.unwrap_or(48000);
+                let output_sample_rate = new_stream.sample_rate.unwrap_or(48000);
                 let num_channels = new_stream.num_channels.unwrap_or(1);
                 let options = NativeAudioStreamOptions {
                     queue_size_frames: new_stream
@@ -106,16 +106,29 @@ impl FfiAudioStream {
                         .map(|capacity| capacity as usize),
                 };
 
+                // When the audio filter supports separate rates (v2), create
+                // the WebRTC sink at 48000 (WebRTC's internal pipeline rate)
+                // to skip resampling — the filter handles rate conversion.
+                let input_sample_rate = if audio_filter
+                    .as_ref()
+                    .is_some_and(|f| f.supports_separate_rates())
+                {
+                    48000
+                } else {
+                    output_sample_rate
+                };
+
                 let native_stream = NativeAudioStream::with_options(
                     rtc_track,
-                    sample_rate as i32,
+                    input_sample_rate as i32,
                     num_channels as i32,
                     options,
                 );
 
                 let stream = if let Some(audio_filter) = &audio_filter {
                     let session = audio_filter.clone().new_session(
-                        sample_rate,
+                        input_sample_rate,
+                        output_sample_rate,
                         new_stream.audio_filter_options.unwrap_or("".into()),
                         info.as_ref().map(|i| i.stream_info.clone()).unwrap(),
                     );
@@ -126,7 +139,8 @@ impl FfiAudioStream {
                                 native_stream,
                                 session,
                                 Duration::from_millis(10),
-                                sample_rate,
+                                input_sample_rate,
+                                output_sample_rate,
                                 num_channels,
                             );
                             AudioStreamKind::Filtered(stream)
@@ -149,7 +163,7 @@ impl FfiAudioStream {
                     true,
                     info,
                     new_stream.frame_size_ms,
-                    sample_rate.try_into().unwrap(),
+                    output_sample_rate.try_into().unwrap(),
                     num_channels.try_into().unwrap(),
                 ));
                 server.watch_panic(handle);
@@ -253,7 +267,7 @@ impl FfiAudioStream {
                 let (c_tx, c_rx) = oneshot::channel::<()>();
                 let (handle_dropped_tx, handle_dropped_rx) = oneshot::channel::<()>();
                 let (done_tx, mut done_rx) = oneshot::channel::<()>();
-                let sample_rate = request.sample_rate.unwrap_or(48000) as i32;
+                let output_sample_rate = request.sample_rate.unwrap_or(48000) as i32;
                 let num_channels = request.num_channels.unwrap_or(1) as i32;
                 let track_sid = track.sid();
                 let options = NativeAudioStreamOptions {
@@ -275,6 +289,15 @@ impl FfiAudioStream {
                     }
                 });
 
+                let input_sample_rate = if filter
+                    .as_ref()
+                    .is_some_and(|f| f.supports_separate_rates())
+                {
+                    48000
+                } else {
+                    output_sample_rate
+                };
+
                 let (mut audio_filter_session, info) = match &filter {
                     Some(filter) => match &request.audio_filter_options {
                         Some(options) => {
@@ -293,7 +316,8 @@ impl FfiAudioStream {
                             };
 
                             let session = filter.clone().new_session(
-                                sample_rate as u32,
+                                input_sample_rate as u32,
+                                output_sample_rate as u32,
                                 &options,
                                 info.stream_info.clone(),
                             );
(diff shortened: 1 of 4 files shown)


livekit-ffi/src/server/audio_stream.rs
```diff
@@ -98,24 +98,35 @@ impl FfiAudioStream {
...
                 let options = NativeAudioStreamOptions {
                     queue_size_frames: new_stream
                         .queue_size_frames
                         .map(|capacity| capacity as usize),
                 };
 
+                // When the audio filter supports separate rates (v2), create
+                // the WebRTC sink at 48000 (WebRTC's internal pipeline rate)
+                // to skip resampling — the filter handles rate conversion.
+                let input_sample_rate =
```
xianshijing-lk, line 112: It looks like we are trying to avoid a webrtc resampling if the audio filter supports_separate_rates() is true ?

Is there any reason that we want to use the audio filter resampling rather than the webrtc one ? 

BTW, I think it depends on the codec, if OPUS 48K is being used, webrtc is using 48K internally; but if other codecs, like ISAC 16K or 32K is being used, it might not be 48K. It has been a while, maybe I remember it wrong.

hechen-eng: The goal is to reduce number of resamplings. It can be done by resampling using webrtc resampling, i.e. `webrtc resampling to model's native sampling rate` -> `AI model` -> `resampling to target 24kHz`. This would work for ai-coustics as well. But we would need to maintain the native sampling rate of the models used.

The current approach seems to be more straightforward since it only needs to specify input and output sampling rate. That being said, ai-coustics doesn't support this. See `Future Work: ai-coustics` in the description.

> BTW, I think it depends on the codec, if OPUS 48K is being used, webrtc is using 48K internally; but if other codecs, like ISAC 16K or 32K is being used, it might not be 48K.

I think this is true. Committed another fix. Feel free to take another look 👍

xianshijing-lk (approved):

livekit-ffi/src/server/audio_stream.rs
xianshijing-lk: Curiously, does Krisp plugin output 24K audio after processing ? or they will resample to what the audio stream is using ? 

For AIC, I think their default sample rate is 48K, so it is good to keep using 48K.
In case we are feeing audio with other sample rate to AIC models, they will internally resample to 48K, process it, then resample the audio back to original sample rate, so it is better to run with their models' native sample rate.

hechen-eng: The output sampling rate is what the TTS expects, which is 24kHz by default I believe. This is passed to Krisp plugin as `output_sample_rate` and this is what it will output.

For AIC, the default is quailL (see [here](https://github.com/livekit/plugins-ai-coustics-internal/blob/main/crates/plugins-ai-coustics-uniffi/support/node/processor.ts#L78)) and the sample rate is 16kHz (see [here](https://docs.ai-coustics.com/guides/models)). The user can choose other models though, and different models may have different native sample rate. I think we should push them to support the API that accepts both input sample rate and output sample rate, which shouldn't be a complex change on their end..

xianshijing-lk: + @1egoman  Ryan, maybe the AIC SDK has been upgraded and now using 16K as default ? 

what do you think about it ? should we pass such feature request to AIC ?

1egoman: > maybe the AIC SDK has been upgraded and now using 16K as default ?

Yes, all of the aic models all operate at 16khz natively both going in and coming out. Some of the models (which we aren't currently using) are advertised as operating on other frequencies, but from discussions with them they have said that those alternate frequencies aren't fundamentally different models, they are just resampling in their sdk back to 16hkz before running the signal through.

I agree that if it's important for some use cases to have different sample rates going in and coming out, we should ask them for it, I also agree I don't think this would be super difficult on their end. I'm not sure if the quality of resampling matters here, but if it does then it might be worth ensuring that their resampler isn't introducing more artifacts than sox would otherwise on an equivilent signal.

> Handle the resampling ourselves in the plugin wrapper.

If it is helpful, I have a [pull request](https://github.com/livekit/plugins-ai-coustics-internal/pull/67) where I actually implemented something like this for a different reason. It didn't end up getting merged (the aic folks recommended we go in a different direction for that specific issue) but in theory actually getting the livekit part of the plugin to do this resampling internally I think would be fairly trivial.

livekit/src/plugin.rs
```diff
@@ -44,9 +44,17 @@ type CreateFn = unsafe extern "C" fn(
...
+type CreateV2Fn = unsafe extern "C" fn(
+    input_sample_rate: u32,
+    output_sample_rate: u32,
+    options: *const c_char,
+    stream_info: *const c_char,
+) -> *mut c_void;
 type DestroyFn = unsafe extern "C" fn(*const c_void);
 type ProcessI16Fn = unsafe extern "C" fn(*const c_void, usize, *const i16, *mut i16);
+type ProcessI16V2Fn = unsafe extern "C" fn(*const c_void, usize, *const i16, usize, *mut i16);
 type ProcessF32Fn = unsafe extern "C" fn(*const c_void, usize, *const f32, *mut f32);
```
1egoman, line 56: Just generally, this worries me a little bit given the aim of moving away from this audio filter interface and instead the `FrameProcessor` in [node](https://github.com/livekit/node-sdks/blob/2bbf15403a5baeabdbab472651780ea3a3ddf9c8/packages/livekit-rtc/src/frame_processor.ts#L37-L50) and [python](https://github.com/livekit/python-sdks/blob/fbbe6f3a9941749aa19f2702cd4c5f66026bc154/livekit-rtc/livekit/rtc/frame_processor.py#L10-L33) (which is what the aic plugin was built on top of).

If it's useful I can go into more background on why we introduced the `FrameProcessor` over continuing to invest in this, but the short version is that it means we can expose this over uniffi and use the same code paths across all sdks rather than independent solutions per platform like Krisp.

The hope with introducing `FrameProcessor` was to deprecate this existing audio filter interface for everything, including krisp, eventually. Adding a new v2 version of this audio filter interface might be the right short term move to make this more performant but it is introducing some technical debt along the way.

livekit-ffi/src/server/audio_stream.rs
hechen-eng: >Yes, all of the aic models all operate at 16khz natively both going in and coming out.

Can you help to confirm this? Their user guide does say different models have different native sample rate. If non native sample rate is given, it'll be resampled to the native sample rate internally and then resample back to the input sample rate https://docs.ai-coustics.com/guides/models#using-models-with-non-native-sample-rates.

> I agree that if it's important for some use cases to have different sample rates going in and coming out, we should ask them for it, I also agree I don't think this would be super difficult on their end. I'm not sure if the quality of resampling matters here, but if it does then it might be worth ensuring that their resampler isn't introducing more artifacts than sox would otherwise on an equivilent signal.

Shall we ask them to support it and hear what they say? Supporting asymmetric input/output sampling rates is an industry-standard feature for audio pipelines, and it is a relatively lightweight change for their engineers to implement IMHO.

The thing is, there will be other models / vendors in the future. We probably don't want to keep track of the native sample rate of different models. So it's better if they can just support asymmetric input/output sampling rate..

livekit/src/plugin.rs
hechen-eng: the deprecation may take a long time especially if there are already users using this.. but I'll leave this up to team to decide whether this optimization is needed 👍

cc @xianshijing-lk

1egoman: Yep totally - sorry I am realizing my original comment wasn't very actionable. I am fine with what you have here for the time being, but I think this use case needs to be taken into account more fully in the `FrameProcessor` interface to make migration to it across the board (including for telephony use cases ) fully possible.

I think I'll set up some time so we can chat about this some more offline and I can get some more context on how krisp is integrated into telephony currently!

livekit-ffi/src/server/audio_stream.rs
1egoman: > Can you help to confirm this?

Ah, you know what, I think this has changed with some of their more recent model updates - they might have models which are built around different frequencies now. That being said right now, all the models we are using (and for the foreseeable future, will probably continue to use) operate at 16khz so resampling will have to happen somewhere.

> Shall we ask them to support it and hear what they say?

I have some news on this, I will reach out on slack and we can chat.

hechen-eng:
https://github.com/livekit/krisp-noise-filter/pull/174 btw, this also needs review @xianshijing-lk @1egoman 🙏

xianshijing-lk:
> [livekit/krisp-noise-filter#174](https://github.com/livekit/krisp-noise-filter/pull/174) btw, this also needs review @xianshijing-lk @1egoman 🙏

it looks good to me. @1egoman  do you have any concern of landing this PR ? 

How about you @alexlivekit  ?

reenboog:
In general, all looks fine, but those unsafe blocks make me a bit nervous. Could it be possible to at least isolate them as functions with internal checks/asserts and unit-test?

hechen-eng:
> In general, all looks fine, but those unsafe blocks make me a bit nervous. Could it be possible to at least isolate them as functions with internal checks/asserts and unit-test?

@reenboog I did some locally testing with an agent using the old/new versions. Here are the procedures and results https://github.com/hechen-eng/sip-bot/blob/main/README.md#observed-results. Everything looks good. I'd prefer to keep the way it is if possible since it's more consistent with existing code. Let me know if you strongly against that..

xianshijing-lk:
@hechen-eng , lets land it. Thanks for the patience.

alexlivekit:
> How about you @alexlivekit ?

Oof, sorry folks, I totally lost track of this one.

imo the reasoning here is great and awesome to see this done.
Today:
- 48000 → 24000: Webrtc
- 24000 → model_rate → 24000: Krisp (and if model_rate == 24000, none at all)

After this change:
- 48000 → model_rate → 24000: Krisp

The one concern I may have here centers on which component does the resampling. `48000 -> model_rate (Krisp)` has not been tested as extensively, compared to `48000 -> 24000 (Webrtc) -> model_rate (Krisp)`, and the results may differ in quality (not necessarily a bad thing!) when comparing these two.

Now, this just means we would want to be attentive to this during the rollout. So a very late LGTM here :)

PR #1021, opened 2026-04-16
webrtc-sys: add armv7 (arm-linux-gnueabihf) support

neilgarb:
- Add "arm" to arch tables for sysroot and glib header paths
- Use -std=c++2a -fpermissive instead of c++20 (compat with GCC 9 in cross-rs)
- Skip desktop_capturer lazy-load stubs for armv7 (only x64/arm64 stubs exist)
- Add -fPIC on Linux so .o files link into Rust PIE binaries


cloudwebrtc (approved):
lgtm!

PR #1032, opened 2026-04-22
Data track send queue improvement

chenosaurus:
the unbounded send queue in `dc_sender.rs` can continue to grow when there is congestion, causing linearly increasing latency to build up.  This PR changes the queue behavior to be bounded and will drop oldest packets if buffer is full.


ladvoc (approved):
LGTM ✅

PR #1034, opened 2026-04-22
Data tracks UniFFI

ladvoc:
Summary of changes:
- Exposes data tracks core functionality through `livekit-uniffi`
  - This will eventually enable the following clients to share the Rust implementation: Swift, Kotlin, React Native, Flutter
- Minor changes to the data tracks crate to support this

Resolves CLT-2472


livekit-uniffi/src/data_track/local.rs
```diff
@@ -0,0 +1,242 @@
...
+    ///
+    /// This must be invoked for the following response types in order for the
+    /// manager to function properly:
+    ///
+    /// - `RequestResponse`
+    /// - `PublishDataTrackResponse`
+    ///
+    /// If a signal response type not listed above is provided, the result is an error.
+    ///
+    pub fn handle_signal_response(&self, res: &[u8]) -> Result<(), HandleSignalResponseError> {
```
pblazej: I think we should take a closer look at this from the perspective of signaling migration (or "hybrid" signaling if we decide to move some features earlier e.g. data streams, RPC or any other small piece) cc @reenboog 

Let's assume that _both_ client language and rust can handle messages, there are 2 general approaches:
- strict chain of responsibility so `if !rust_handled(signal) { other_handled(signal) }` - the main problem with this approach is there's no _central_ place in rust to decide that the message was indeed handled, so hypothetically we'd need to pass it to each data track manager, etc. (and build another chain inside rust)
- dispatch - client deserializes and knows what to do, passes `rust_handle_specific_message` down

What I wanna avoid is we re-wrap messages that we already deserialize e.g. in swift (and know what to do with them) just to pass inside this `handle_signal_response(&self, res: &[u8])`, this should be more specific like `fn handle_publish_response` etc.

The same may apply to the "receiving" side.

This problem won't exist as we move the signaling to rust ofc, but let's make sure we can do that smoothly.

pblazej (commented):
I don't see too much glue code for Swift now, it's definitely doable; I'm mostly worried about the "intermediate" layer so some awkward signaling patterns/states etc.

livekit-uniffi/src/data_track/local.rs
```diff
@@ -0,0 +1,242 @@
...
+        self.0.wait_for_unpublish().await
+    }
+
+    /// Information about the data track.
+    pub fn info(&self) -> DataTrackInfo {
+        self.0.info().into()
+    }
+
+    /// Try pushing a frame to subscribers of the track.
+    pub fn try_push(&self, frame: DataTrackFrame) -> Result<(), PushFrameErrorReason> {
```
pblazej, line 55: This and other "throwing" methods: separate error type can't be mapped directly to `LiveKitError`, so we need to "rethrow" in Swift (probably others as well), which is yet another typical problem for such wrappers.

On the other hand, this whole pattern of 1 global error type (+code) is outdated, and we've got counterexamples like `StreamError` already in the public API.

I'd rather use typed throws for Swift here directly (even if inconsistent with the rest of the SDK) or leave it as-is.

livekit-uniffi/src/data_track/local.rs
```diff
@@ -0,0 +1,242 @@
...
+        // `PushFrameError` returns ownership of the unpublished frame to the caller;
+        // since this isn't applicable in an FFI context, just provide the reason.
+        self.0.try_push(frame).map_err(|err| err.reason())
+    }
+
+    /// Unpublishes the track.
+    pub fn unpublish(&self) {
```
pblazej, line 62: I like the new APIs a lot, the only asymmetry for me is that you cannot surface the fact that unpublished track is "invalid" just via language constructs ("consuming", scopes, etc.).

I'm thinking of sth like:
```swift
  public func withDataTrack<T>(
      name: String,
      body: (LocalDataTrack) async throws -> T
  ) async throws -> T {
      let track = try await publishDataTrack(name: name)
      defer { track.unpublish() }
      return try await body(track)
  }
```

So you can _pipe_ streams e.g.
```swift
  // Instead of handing back an object that can go stale
  try await localParticipant.withDataTrack(name: "game-state") { track in
      // track is guaranteed valid here
      for await state in gameStates {
          track.tryPush(frame: state.toFrame())
      }
  }
  // automatically unpublished when scope exits (cancellation, throw, or return)
```

Or even:
```swift
  // Publishing: AsyncSequence → DataTrack (sink)
  try await localParticipant.publishDataTrack(name: "game", sending:
      gameLoop.map { $0.toFrame() }
  )

  // Subscribing: DataTrack → AsyncSequence (source)
  for await frame in stream {
      process(frame)
  }
```

It's purely additive, so take it with a grain of 🧂 but IMO provides a great DX on top of data tracks if you stick to certain stream-like constructs.

livekit-uniffi/src/data_track/remote.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/// A stream of [`DataTrackFrame`]s received from a [`RemoteDataTrack`].
+#[derive(uniffi::Object)]
+pub struct DataTrackStream(Mutex<livekit_datatrack::api::DataTrackStream>);
+
+#[uniffi::export]
+impl DataTrackStream {
+    /// Returns the next received frame or `None` if the subscription has ended.
+    pub async fn next(&self) -> Option<DataTrackFrame> {
```
pblazej, line 96: This could be mapped in a generic way; the impl is trivial in Swift, basically using `AsyncStream(unfolding: next)`, as we do for logs.

livekit-uniffi/src/data_track/local.rs
pblazej: Sorry there's also a 3rd way - passing everything at data (raw proto) level to _both_ client and rust, not carrying about overlaps at all, let's call it message bus.

livekit-uniffi/src/common.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use bytes::Bytes;
+
+uniffi::custom_type!(Bytes, Vec<u8>, { remote });
```
pblazej, line 17: General: do we plan to introduce some helpers for encoding/decoding into that or leaving that to experienced users?

livekit-datatrack/src/frame.rs
```diff
@@ -32,8 +32,8 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 ///
 #[derive(Clone, Default)]
 pub struct DataTrackFrame {
-    pub(crate) payload: Bytes,
-    pub(crate) user_timestamp: Option<u64>,
+    pub payload: Bytes,
+    pub user_timestamp: Option<u64>,
```
pblazej: Nit: shall we expose `duration_since_timestamp` e.g. for benchmarks?

livekit-datatrack/src/frame.rs
pblazej, line 1: If this is not an `uniffi::Object` we won't get `withTimestampNow` etc. for free, right? As it basically maps to raw struct:

```swift
public struct DataTrackFrame: Equatable, Hashable, Sendable {
  public var payload: Data
  public var userTimestamp: UInt64?
}
```

livekit-uniffi/src/data_track/remote.rs
```diff
@@ -0,0 +1,241 @@
...
+        // TODO: avoid mutex?
+        self.0.lock().await.next().await
```
pblazej: cc @reenboog talking about async 😄

livekit-datatrack/src/frame.rs
pblazej: Applies to other "DTOs" as well, so good to discuss that now.

pblazej (commented):
I still don't see any real blockers here, would be nice to discuss some common cases like errors and records now.

pblazej:
Hm, I see one more gap in the runtime after introducing 90cf95ac0d8aa5a043da11671054bf84d0125e5e

UniFFI polls async futures from Swift's thread via continuation callbacks. That thread has no tokio runtime context. When the future hits `tokio::time::timeout` (line 453 in `manager.rs`), tokio tries to register a timer with a runtime that doesn't exist on that thread → panic.

The naive claude solution was:

```rust
pub(crate) fn contextualize<F: std::future::Future>(
    future: F,
) -> impl std::future::Future<Output = F::Output> {
    use std::pin::Pin;
    use std::task::{Context, Poll};

    struct Contextualized<F> {
        inner: Pin<Box<F>>,
    }

    impl<F: std::future::Future> std::future::Future for Contextualized<F> {
        type Output = F::Output;

        fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
            let _enter = runtime().enter();
            self.inner.as_mut().poll(cx)
        }
    }

    Contextualized { inner: Box::pin(future) }
}
```

then e.g.

```rust
crate::runtime::contextualize(self.input.publish_track(options))
            .await
            .map(LocalDataTrack)
```

reenboog:
It's even more generic: at some point, a reactor might be expected (timers, network, any io), and none found. So, I think we should wrap all publicly exportable functions with something like this:

```rust
// a cancelable version could be introduced as well
pub async fn on_runtime<F, Res>(future: F) -> Res
where
    F: Future<Output = Res> + Send + 'static,
    Res: Send + 'static,
{
    get_runtime()
        .spawn(future)
        .await
        .expect("A background Tokio task panicked")
}

...

// when exporting, wrap with on_runtime
#[uniffi::export]
pub async fn ffi_do_work() -> Result<MyFfiType, FfiErr> {
    on_runtime(async {
        let res = core::do_work(params).await?;
        Ok(res)
    })
    .await
}
```

I already started drafting this in my work related to signaling.

livekit-uniffi/src/data_track/common.rs
```diff
@@ -0,0 +1,56 @@
...
+        Self { sid: info.sid(), name: info.name().to_string(), uses_e2ee: info.uses_e2ee() }
+    }
+}
+
+/// Signal response crossing the FFI boundary could not be processed.
+#[derive(uniffi::Error, thiserror::Error, Debug)]
+#[uniffi(flat_error)]
+pub enum HandleSignalResponseError {
+    #[error("Response decoding failed: {0}")]
+    Decode(prost::DecodeError),
```
1egoman, line 71: thought: It is worth making a unique error type here rather than using `prost::DecodeError` so that the internal protobuf implementation type doesn't leak through the interface?

livekit-uniffi/src/data_track/e2ee.rs
```diff
@@ -0,0 +1,103 @@
...
+    ///
+    fn decrypt(
+        &self,
+        payload: EncryptedPayload,
+        sender_identity: String,
+    ) -> Result<Bytes, DecryptionError>;
+}
+
+/// Adapts [`DataTrackEncryptionProvider`] to implement [`EncryptionProvider`].
+pub(super) struct FfiEncryptionProvider(pub(super) Arc<dyn DataTrackEncryptionProvider>);
```
1egoman: thought: Do you need a `uniffi` attr macro of some sort here and on `FfiDecryptionProvider`?

livekit-uniffi/src/data_track/local.rs
1egoman: Generally makes sense to me! Though IMO since this is somewhat swift specific, maybe this makes sense to expose in the swift specific code above this rather than in the uniffi definition. Also because I don't _think_ sending a closure over uniffi is possible.

livekit-uniffi/src/data_track/local.rs
```diff
@@ -0,0 +1,244 @@
...
+        let encryption_provider = encryption_provider
+            .map(|p| Arc::new(FfiEncryptionProvider(p)) as Arc<dyn EncryptionProvider>);
+        let manager_options = local::ManagerOptions { encryption_provider };
+
+        let (manager, input, output) = local::Manager::new(manager_options);
+
+        let rt = crate::runtime::runtime();
+
+        // TODO: in a follow-up PR, refactor manager to work with cancellation tokens directly, eliminating the
+        // need for this additional task.
```
1egoman, line 117: (Just a note that the web implementation has quite a bit of of logic dealing with this, so if nothing else there's a lot of good test cases to pull in)

livekit-uniffi/src/data_track/local.rs
1egoman: Given each client implementation pulls in the protobufs today anyway and that isn't likely to change, I think I'd vote for 2 (`dispatch - client deserializes and knows what to do, passes rust_handle_specific_message down`). This was generally what I did on the web (I realize it's not exactly the same set of problems though) and it worked quite well. If there's a desire to eventually do all protobuf serialization / deserialization at the manager level, then I think it's fairly reasonable to then move to 1 or 3 once we're more confident which pattern makes the most sense.

livekit/src/room/e2ee/data_track.rs
```diff
@@ -37,7 +37,7 @@ impl dt::EncryptionProvider for DataTrackEncryptionProvider {
         let encrypted = self
             .manager
             .encrypt_data(payload.into(), &self.sender_identity, key_index)
-            .map_err(|_| dt::EncryptionError)?;
+            .map_err(|_| dt::EncryptionError::Failed)?;
```
1egoman, line 40: question: Is this a breaking api change to existing rust sdk users?

livekit-uniffi/src/data_track/local.rs
pblazej: > rather than in the uniffi definition

yes, I'm talking purely about wrapping here.

livekit-uniffi/src/data_track/local.rs
pblazej: For now, at least for swift - we must deserialize anyway, as the "handlers" are spread across different layers - this problem will disappear/move as we move the signaling to rust.

ladvoc: Switched to option 2 in [b449306](https://github.com/livekit/rust-sdks/pull/1034/commits/b44930608fbf929581f8436885e094af5ce672fd).

livekit-uniffi/src/data_track/local.rs
ladvoc: I don't think UniFFI supports generating Swift functions with typed throws yet, but we will get that for free once it does.

livekit-uniffi/src/data_track/local.rs
ladvoc: Something to be aware of is the client doesn't have sole control over the publish state of a data track (the SFU reserves the right to unpublish at any time without the client initiating it) which is why I didn't do something similar in Rust (e.g., modeling with type state).

livekit-uniffi/src/common.rs
ladvoc: This macro handles the conversion from `Bytes` to `Vec<u8>` automatically (under the hood it implements `impl uniffi::FfiConverter<crate::UniFfiTag> for Bytes`), so Swift can just pass `Data` as expected wherever the API on the Rust side accepts `Bytes`.

livekit-uniffi/src/data_track/common.rs
ladvoc: Since this uses the `#[uniffi(flat_error)]` macro ([docs](https://mozilla.github.io/uniffi-rs/0.27/proc_macro/index.html)), the associated value for each case gets converted to a string via display for the purposes of crossing the FFI boundary so none of the inner error types are exposed—but we still get the enum cases.

livekit-uniffi/src/data_track/e2ee.rs
ladvoc: No, since this type never crosses the FFI boundary. The current approach for exporting the E2EE traits is a bit awkward; I had to introduce FFI-only wrapper traits since nothing like `#[uniffi::remote(Trait)]` exists. However, once we move the UniFFI macros to the `livekit-datatrack` crate, this should no longer be necessary—`#[uniffi::export(with_foreign)]` can be applied directly to the trait definitions then.

livekit-uniffi/src/data_track/local.rs
ladvoc: I think this will map nicely; [_CancellationToken_](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) is roughly equivalent to _AbortSignal_ on web.

livekit-datatrack/src/frame.rs
ladvoc: It seems this type of "DTO" is more naturally modeled as a value type on the Swift side. I don't think you can currently have associated functions on a `uniffi::Record` (even though this makes sense in a Swift context), so I see two options:
1. Use `uniffi::Object`
2. Export standalone helper functions (e.g., `fn with_user_timestamp(frame: DataTrackFrame) → DataTrackFrame`) and define an `extension DataTrackFrame` on the Swift side to make it an associated function

livekit/src/room/e2ee/data_track.rs
ladvoc: No, since this type is never exposed in the public API in the _livekit_ crate.

ladvoc:
> It's even more generic: at some point, a reactor might be expected (timers, network, any io), and none found. So, I think we should wrap all publicly exportable functions with something like this [...]

@reenboog, is this something you would want to add in the _livekit_runtime_ crate?

livekit-datatrack/src/frame.rs
ladvoc: Good news: UniFFI [v0.31.0](https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md#v0310-backend-crates-v0310---2026-01-14) added support for methods on records and enums!

livekit-uniffi/src/data_track/local.rs
pblazej: Yes, this "problem" may exist in other langs like Kotlin if we use "centralized error type"

livekit-uniffi/src/data_track/local.rs
pblazej: Ah so if someone "unpublished it for me" it will just throw and break the loop?

livekit-uniffi/src/common.rs
pblazej: Yeah, I just mean "something higher-level than `Data`" like `Encodable` types (JSON etc.)

livekit-datatrack/src/frame.rs
pblazej: Yup let's avoid handwritten extensions and just try bumping UniFFI.

pblazej: Edit: tried that locally, and it does not work for `#[uniffi::remote(Record)]` e.g. `DataTrackFrame`

reenboog:
>is this something you would want to add in the livekit_runtime crate?
@ladvoc, yes, the idea is all async public functions, exported with uniffi, should be wrapped with that `on_runtime` I described earlier, to make it actually run on a runtime instead of piggybacking on whatever each platform provides. No manual `enter` is required, nor manual polling. I'm wrapping up a small PR to expose a runtime, so that both, uniffi-based modules and the legacy (protobuf-based) ones could use that global runtime.

pblazej:
@davidliu tagging you as it would be great to get some approval _before_ the actual "0.1" release.

Let's make sure the DTOs etc. look fine at least on 2 platforms, the async things work, etc.

davidliu:
@pblazej looks good, got it compiled for Android and played around with the API. Seems like everything would integrate fairly seamlessly into the Android SDK.

@ladvoc I did have to add the "macros" feature to [this line in the Cargo.toml to get it building](https://github.com/livekit/rust-sdks/blob/1848e77a32ca2f302a58323a7a21858165d81aad/livekit-datatrack/Cargo.toml#L15):
```
tokio = { workspace = true, default-features = false, features = ["macros", "sync"] }
```

I'm not sure if this is android specific or affects other builds as well.

pblazej:
Wiring up the Swift SDK — hit two gaps in the binding surface for join/reconnect @ladvoc @1egoman:

**Join-time tracks.** No join handler on `RemoteDataTrackManager`, even though the reference SDK runs `event_from_join` on connect (`rtc_session.rs:531`) — so a foreign client joining a room with existing data tracks won't see them. Either add a `handle_sfu_join`, or just feed `join.other_participants` through the existing `handle_sfu_participant_update` (what JS effectively does). Identical under the hood — both hit `event_from_participant_info`. Preference?

**Quick-reconnect sync state.** No way to build `SyncState.publishDataTracks` from the bindings. `query_tracks()` is exposed, but the uniffi `DataTrackInfo` drops `pub_handle`, so I can't construct `PublishDataTrackResponse` client-side the way JS/Rust do (their info carries the handle). Either expose `pub_handle` on the uniffi `DataTrackInfo`, or add a `publish_responses_for_sync_state()` returning the serialized responses (keeps the handle internal — probably cleaner).

**Caveats:** no released livekit-server (incl. 1.13.2) fills `ParticipantInfo.data_tracks` in the join, so join discovery isn't verifiable e2e atm — is the SFU side still unreleased?

ladvoc:
> Quick-reconnect sync state. No way to build SyncState.publishDataTracks from the bindings. query_tracks() is exposed, but the uniffi DataTrackInfo drops pub_handle, so I can't construct PublishDataTrackResponse client-side the way JS/Rust do (their info carries the handle). Either expose pub_handle on the uniffi DataTrackInfo, or add a publish_responses_for_sync_state() returning the serialized responses (keeps the handle internal — probably cleaner).

Makes sense, I implemented the latter in [4f18c9c](https://github.com/livekit/rust-sdks/pull/1034/commits/4f18c9c010e293f28d242491abe05f28c6f40e7b). This replaces the `query_tracks` method that was previously exported.

ladvoc:
> Join-time tracks. No join handler on RemoteDataTrackManager, even though the reference SDK runs event_from_join on connect (rtc_session.rs:531) — so a foreign client joining a room with existing data tracks won't see them. Either add a handle_sfu_join, or just feed join.other_participants through the existing handle_sfu_participant_update (what JS effectively does). Identical under the hood — both hit event_from_participant_info. Preference?

Rust has these additional helpers to keep the interface between the rest of the client SDK and the data tracks managers as uniform as possible; instead of caring about specific fields in each response type, you hand over the whole response, and the logic of how the required information is extracted is kept as an implementation detail. A few other advantages:
- If the data track managers ever need additional information not extracted from participant info, no update to this interface will be needed.
- Local participant identity isn't needed when handling a join response but it is when handling a participant update; this gets modeled in the signature of each method.

I implemented `handle_sfu_join_response` in [8abcf90](https://github.com/livekit/rust-sdks/pull/1034/commits/8abcf90f128e19ce49743258c5d451a207d7bb32). While these helpers don't add any overhead in the Rust client integration since they accept the responses by reference, in an FFI context, there is additional overhead because the whole response must be serialized rather than just the relevant fields—I'm open to changing this if the additional overhead is a concern.

ladvoc:
> Caveats: no released livekit-server (incl. 1.13.2) fills ParticipantInfo.data_tracks in the join, so join discovery isn't verifiable e2e atm — is the SFU side still unreleased?

Data tracks is fully released in OSS and Cloud now so this would be unexpected. Do you have any repro steps for this?

pblazej:
> Do you have any repro steps for this?

No repro — false alarm on my side. The join **does** carry `data_tracks` on 1.13.2; my test was discarding the returned `LocalDataTrack` (`_ = try await publishDataTrack(...)`), which unpublishes it on drop, so the track was already gone before the late joiner arrived. Holding it, join-time discovery works. Both `handle_sfu_join_response` and `publish_responses_for_sync_state` are wired up and green e2e now.

That RAII-on-drop is really the last open question: should FFI consumers mirror Rust's semantics (drop = unpublish) or go "publication-like" (JS keeps the publication in its manager until an explicit unpublish)? We went publication-like on the Swift side so callers aren't forced to retain the handle — this diff shows it: https://github.com/livekit/client-sdk-swift/pull/975/changes/3e3043ec3fba8f6d4618dd9fc421bd6405062fb4

Framing it as "what should the convention be" rather than "should Rust change": the core's RAII is correct/idiomatic for Rust consumers — the divergence is about what the FFI bindings present. JS and (now) Swift both land on publication-like, so there's a de-facto convention worth making explicit.

For concreteness, the Swift lifecycle surface:

| API | Semantics |
|---|---|
| `publishDataTrack(name:)` | publication-like — SDK retains it; discarding the handle does not unpublish |
| `track.unpublish()` / `await track.waitForUnpublish()` | explicit unpublish + await |
| `withDataTrack(name:) { … }` | scoped RAII — auto-unpublishes when the block exits |

`withDataTrack` is the compromise: it keeps Rust's RAII where it's genuinely a feature (an explicit lexical scope where auto-cleanup is exactly what you want), while the default `publishDataTrack` stays publication-like so discarding the handle doesn't surprise-unpublish.

pblazej (approved):
I don't see any functional gaps here preventing Swift [integration](https://github.com/livekit/client-sdk-swift/pull/975)

The rest is mostly consumer-side discussion (like the RAII pattern mentioned above).

Before merging please update both `SPM_SIZE_LIMIT_BYTES` and `ANDROID_SIZE_LIMIT_BYTES` with the final sizes.

Approximate sizes @ the latest commit:
- `SPM_SIZE_LIMIT_BYTES = 1152136`
- `ANDROID_SIZE_LIMIT_BYTES = 1250710`

pblazej:
@hiroshihorie cc what's your take on applying that in flutter (as the infra is ~complete)?

livekit-datatrack/src/e2ee.rs
```diff
@@ -52,6 +66,17 @@ pub trait DecryptionProvider: Send + Sync + Debug {
     fn decrypt(
         &self,
         payload: EncryptedPayload,
-        sender_identity: &str,
+        sender_identity: String,
     ) -> Result<Bytes, DecryptionError>;
```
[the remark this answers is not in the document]
ladvoc: Noted in a comment this should be addressed later.

ladvoc:
I have temporarily disable Dart tests in CI since they are currently failing. In the interest of getting this merged, we can address in a follow-up PR.

PR #1037, opened 2026-04-22
get webrtc adm into rust

xianshijing-lk:
Summary

  This PR implements Platform Audio support for the LiveKit Rust SDK, enabling WebRTC's built-in audio device handling with microphone capture and speaker playout. The implementation introduces a handle-based PlatformAudio API that coexists with the existing NativeAudioSource for manual audio pushing.

  Key Features

  - Two Audio Source Types:
    - RtcAudioSource::Native (default): Manual audio push via NativeAudioSource for TTS, file streaming, agents
    - RtcAudioSource::Device: WebRTC handles mic capture & speaker playout with echo cancellation (AEC)
  - Handle-based API: Create PlatformAudio instances that enable ADM recording; drop to release
  - Reference Counting: Multiple PlatformAudio instances share the same underlying ADM
  - Device Enumeration & Selection: List and select recording/playout devices
  - Hot-swap Device Switching: switch_recording_device() / switch_playout_device() for changing devices during active sessions
  - Audio Processing Configuration: AEC, AGC, NS with hardware/software preference
  - WebRTC Patching: external_audio_source.patch prevents audio mixing conflicts between device and manual sources

  Design Document

  See docs/ADM_PROXY_DESIGN.md for full architecture details including:
  - Recording gate pattern
  - WebRTC patching explanation
  - FFI API documentation

  API Overview

  use livekit::prelude::*;

  // Create PlatformAudio instance (enables ADM recording)
  let audio = PlatformAudio::new()?;

  // Enumerate and select devices
  for i in 0..audio.recording_devices() as u16 {
      println!("Mic [{}]: {}", i, audio.recording_device_name(i));
  }
  audio.set_recording_device(0)?;

  // Connect and publish
  let (room, _) = Room::connect(&url, &token, RoomOptions::default()).await?;
  let track = LocalAudioTrack::create_audio_track("mic", audio.rtc_source());
  room.local_participant().publish_track(LocalTrack::Audio(track), opts).await?;

  // Cleanup - just drop the handle
  room.close().await?;
  drop(audio);  // ADM recording disabled when all handles released

  Testing

  Run Standalone Tests (no LiveKit server required)

  # Set custom WebRTC build path
  export LK_CUSTOM_WEBRTC="/path/to/webrtc-sys/libwebrtc/mac-arm64-debug"

  # Run standalone PlatformAudio tests
  cargo test -p livekit --test platform_audio_test test_platform_audio_standalone -- --nocapture

  # Run FFI request handler tests
  cargo test -p livekit-ffi requests::tests -- --nocapture

  Run E2E Integration Tests (requires LiveKit server)

  # Start a local LiveKit server first, then:
  LIVEKIT_URL=ws://localhost:7880 \
  LIVEKIT_API_KEY=devkey \
  LIVEKIT_API_SECRET=secret \
  cargo test -p livekit --test platform_audio_test --features __lk-e2e-test -- --nocapture

  Test Coverage
             Category            │ Tests │                         Description                          │
  
   Standalone - Creation         │ 1     │ PlatformAudio creation, device enumeration
   Standalone - Ref Counting     │ 1     │ Clone, sharing, drop behavior
   Standalone - Device Selection │ 1     │ Set devices, invalid index handling 
   Standalone - Processing       │ 1     │ AEC/AGC/NS configuration, hardware availability
   Standalone - Reset            │ 1     │ reset_platform_audio() function
   Standalone - Lifecycle        │ 1     │ Full create→configure→use→release cycle
   FFI - Handlers                │ 6     │ NewPlatformAudio, GetDevices, SetDevice, handle lifecycle
   E2E - Room Connection         │ 4+    │ Platform audio with room, two participants, device switching
  
All tests handle missing audio devices gracefully (CI-friendly).
  Run the Example

  List Audio Devices

  cargo run -p basic_room -- --list-devices

  Connect with Platform Audio (microphone capture)

  LIVEKIT_URL=wss://your-server.livekit.cloud \
  LIVEKIT_API_KEY=your-key \
  LIVEKIT_API_SECRET=your-secret \
  cargo run -p basic_room -- --platform-audio

  Connect with File Audio

  cargo run -p basic_room -- --file path/to/audio.raw

  Connect with Both Platform Audio and File

  cargo run -p basic_room -- --platform-audio-and-file path/to/audio.raw

  WebRTC Build Requirements

  The external_audio_source.patch must be applied to WebRTC. The patch is automatically applied by all platform build scripts:
  - build_macos.sh
  - build_ios.sh
  - build_linux.sh
  - build_android.sh
  - build_windows.cmd

  For local development, set LK_CUSTOM_WEBRTC to point to your patched WebRTC build.

  Known Limitations
  
       Limitation      │                                  Description 
   Process-global      │ Audio configuration affects all rooms in the process  
   Device indices      │ May change on hot-plug; match by name for persistence 
   Single device track │ One device audio track per ADM (use NativeAudioSource for additional streams)


libwebrtc/src/audio_source.rs
```diff
@@ -23,21 +21,140 @@ pub struct AudioSourceOptions {
...
+    }
+
+    /// Get the sample rate.
+    /// Note: For `Device` source, returns 48000 (default WebRTC sample rate).
+    pub fn sample_rate(&self) -> u32 {
+        match self {
+            #[cfg(not(target_arch = "wasm32"))]
+            RtcAudioSource::Native(source) => source.sample_rate(),
+            #[cfg(not(target_arch = "wasm32"))]
+            RtcAudioSource::Device => 48000, // Default WebRTC sample rate
```
ladvoc: nitpick: This should probably be defined as a constant somewhere.

livekit/src/rtc_engine/lk_runtime.rs
```diff
@@ -51,6 +63,153 @@ impl LkRuntime {
...
+
+    // ===== ADM Management Methods =====
+    // These methods allow runtime control of the Audio Device Module
+
+    /// Enable platform ADM (WebRTC's built-in audio device management)
+    ///
+    /// When enabled, WebRTC handles audio device enumeration, selection,
+    /// and audio capture/playout automatically.
+    ///
+    /// Note: This is an internal method used by FFI. Platform ADM is not
```
ladvoc: comment: Generally we should avoid FFI-only methods in the public API—although we do have them in some places. If that is unavoidable in this case, I would recommend annotating with `#[doc(hidden)]`.

livekit/src/rtc_engine/lk_runtime.rs
```diff
@@ -14,17 +14,29 @@
...
+#[cfg(not(target_arch = "wasm32"))]
+use libwebrtc::peer_connection_factory::native::PeerConnectionFactoryExt;
+
 lazy_static! {
     static ref LK_RUNTIME: Mutex<Weak<LkRuntime>> = Mutex::new(Weak::new());
 }
 
+/// Tracks the number of active room connections.
+/// Used to prevent audio mode switching while rooms are connected.
+static ACTIVE_ROOM_COUNT: AtomicUsize = AtomicUsize::new(0);
```
ladvoc: suggestion: A potentially cleaner way to handle this is to have every room hold a `Arc<()>` and leverage [`strong_count`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.strong_count) to learn the number of active rooms—no need to manually decrement.

livekit/tests/audio_manager_test.rs
```diff
@@ -0,0 +1,845 @@
...
+    // Set to Synthetic mode
+    let result = audio.set_mode(AudioMode::Synthetic);
+    assert!(result.is_ok());
+
+    assert_eq!(audio.current_mode(), AdmDelegateType::Synthetic);
+}
+
+/// Test setting Platform mode.
+#[test]
+#[serial]
```
ladvoc: comment (non-blocking): Currently in CI, all tests are run serially. If we switch over to Nextest (outdated PR, #816), we can configure which tests are run in serial through that config and run everything else in parallel.

docs/ADM_PROXY_DESIGN.md
```diff
@@ -0,0 +1,940 @@
...
+- Audio is pushed via `capture_frame()`
+- `is_external_source() = true` prevents AudioState interference
+- AEC does NOT work (no valid playout reference)
+
+**Suitable for:**
+- Server-side agents
+- Text-to-speech (TTS) audio
+- Audio from files or network streams
+- Testing without audio hardware
+
```
MaxHeimbrock, line 844: This is the original audio input right? Existing Unity clients who want to keep the "Unity" style microphone management would also use this.

docs/ADM_PROXY_DESIGN.md
```diff
@@ -0,0 +1,940 @@
...
+| Remote audio to speakers | Manual via `NativeAudioStream` | Automatic via ADM |
+| Application code needed | Create stream + route to speaker | None |
+| Latency | Depends on app implementation | Optimized by WebRTC |
+| Audio mixing | Application handles | WebRTC handles |
+| Device selection | Application handles | `set_playout_device()` |
+| AEC reference | Not available | Available |
+
+### Hybrid Approach
+
+You can combine both approaches - use `PlatformAudio` for automatic speaker playback while also creating `NativeAudioStream` for audio processing/analysis:
```
MaxHeimbrock, line 1026: Is this also possible from a Unity client? As we discussed, for the lip sync animation Unity clients might want read access to the audio data, but still want output through the platform audio.

docs/ADM_PROXY_DESIGN.md
```diff
@@ -0,0 +1,940 @@
...
+
+message SetRecordingDeviceResponse {
+  optional string error = 1;
+}
+
+// Set playout device
+message SetPlayoutDeviceRequest {
+  uint64 platform_audio_handle = 1;
+  uint32 index = 2;
+}
```
MaxHeimbrock, line 1204: How does it handle switching the device at runtime?

docs/ADM_PROXY_DESIGN.md
```diff
@@ -0,0 +1,940 @@
...
+- `recording_enabled_ = false` (default)
+- ADM recording operations are no-ops
+- Audio is pushed via `capture_frame()`
+- `is_external_source() = true` prevents AudioState interference
+- AEC does NOT work (no valid playout reference)
+
+**Suitable for:**
+- Server-side agents
+- Text-to-speech (TTS) audio
+- Audio from files or network streams
```
MaxHeimbrock, line 842: Or screen share audio right?

livekit-ffi-node-bindings/proto/audio_frame_pb.d.ts
```diff
@@ -136,9 +136,19 @@ export declare enum AudioStreamType {
  */
 export declare enum AudioSourceType {
```
xianshijing-lk, line 137: @ladvoc , I change the audio_frame proto in this PR, should I commit this pb.d.ts change? not sure why the node proto code is in the rust SDK

livekit/tests/audio_manager_test.rs
xianshijing-lk: fixed.

livekit/src/rtc_engine/lk_runtime.rs
xianshijing-lk: fixed.

livekit/src/rtc_engine/lk_runtime.rs
xianshijing-lk: done with adding #[doc(hidden)].

libwebrtc/src/audio_source.rs
xianshijing-lk: Done

docs/ADM_PROXY_DESIGN.md
xianshijing-lk: Yes, that is possible with Unity client for such lip sync animation + mic / speaker audio via PlatformAudio

docs/ADM_PROXY_DESIGN.md
xianshijing-lk: webrtc handles the runtime device switching automatically. 

If the current device is removed during a call, it will automatically switch over to "default device", and the call should still be on-going.

docs/ADM_PROXY_DESIGN.md
xianshijing-lk: Yes

examples/basic_room/src/main.rs
```diff
@@ -1,31 +1,337 @@
...
+//
+// Example with GUID-based device selection:
+//   cargo run -p basic_room -- --list-devices                         # Find the GUID you want
+//   cargo run -p basic_room -- --platform-audio --mic-guid "your-guid-here"
 
 #[tokio::main]
 async fn main() {
     env_logger::init();
 
+    let args: Vec<String> = env::args().collect();
```
ladvoc: suggestion: Since we already use `clap` (with derive) for argument parsing in the other examples, it might be a good idea to use the same approach here. We also would get `--help` for free if the args have doc comments.

examples/basic_room/src/main.rs
```diff
@@ -1,31 +1,337 @@
...
 use std::env;
+use std::path::Path;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::signal;
 
-// Connect to a room using the specified env variables
-// and print all incoming events
+// Usage:
```
ladvoc: nitpick: Other examples put usage guide in a README in the example directory.

livekit/src/platform_audio/mod.rs
```diff
@@ -0,0 +1,1110 @@
...
+
+    /// Returns the number of available recording (microphone) devices.
+    ///
+    /// # Example
+    ///
+    /// ```rust,ignore
+    /// let audio = PlatformAudio::new()?;
+    /// println!("Found {} microphones", audio.recording_devices());
+    /// ```
+    pub fn recording_devices(&self) -> i16 {
```
ladvoc: suggestion: Index-based property accessors like this are not typical in Rust APIs. I would recommend encapsulating these info fields in a struct and making the API iterator based:
```rust
struct RecordingDeviceInfo {
    pub index: u16, // or new type
    pub name: String,
    pub guid: String // or new type
}
```
The signature of this method becomes:
```rust
pub fn recording_devices(&self) -> impl IntoIterator<Item = AudioDeviceInfo>
```

Usage:
```rust
let audio = PlatformAudio::new()?;
for device in audio.recording_devices() {
    println!("{} (GUID: {})", device.name, device.guid);
}

// Alternatively, collect into a Vec
let device_list: Vec<_> = audio.recording_devices().collect();
```

This same pattern would apply to playout devices.

livekit/src/platform_audio/mod.rs
```diff
@@ -0,0 +1,1110 @@
...
+    /// ```rust,ignore
+    /// let audio = PlatformAudio::new()?;
+    /// // Save the GUID of the preferred microphone
+    /// let preferred_guid = audio.recording_device_guid(0);
+    /// // Later, select it by GUID (works even if indices changed)
+    /// audio.set_recording_device_by_guid(&preferred_guid)?;
+    /// ```
+    ///
+    /// [`recording_device_guid`]: Self::recording_device_guid
+    pub fn set_recording_device_by_guid(&self, guid: &str) -> AudioResult<()> {
```
ladvoc: suggestion: This is a good application for the new [type pattern](https://doc.rust-lang.org/rust-by-example/generics/new_types.html). The query API would provide a new type wrapper (e.g. `RecordingDeviceGuid`) instead of a string for device GUID, and this method would accept it. This method still has to be fallible since a device might no longer be available, but applying this pattern adds a level of type safety that enforces correct usage (e.g., providing an arbitrary string that is not a valid guid is not possible). Also applicable to playout devices.

livekit/src/platform_audio/mod.rs
```diff
@@ -0,0 +1,1110 @@
...
+    ///
+    /// ```rust,ignore
+    /// // During an active call, switch to a different microphone
+    /// audio.switch_recording_device(1)?;
+    /// ```
+    ///
+    /// [`set_recording_device`]: Self::set_recording_device
+    pub fn switch_recording_device(&self, index: u16) -> AudioResult<()> {
+        let count = self.recording_devices();
+        if index >= count as u16 {
```
ladvoc: question: What happens if the device index is invalidated between this check and the call to `runtime.set_recording_device(index)`?

livekit/src/platform_audio/mod.rs
```diff
@@ -0,0 +1,1110 @@
...
+
+        // Initialize recording if not already initialized
+        if !runtime.recording_is_initialized() {
+            if !runtime.init_recording() {
+                return Err(AudioError::OperationFailed("init_recording failed".to_string()));
+            }
+        }
+
+        if runtime.start_recording() {
+            log::info!("PlatformAudio: started recording");
```
ladvoc, line 774: nitpick: Maybe this should be at debug level? Also applies to the log in `stop_recording`.

livekit/src/platform_audio/mod.rs
```diff
@@ -0,0 +1,1110 @@
...
+
+    /// Enables or disables echo cancellation.
+    ///
+    /// This is a convenience method equivalent to calling `configure_audio_processing`
+    /// with only the `echo_cancellation` field changed.
+    ///
+    /// # Arguments
+    ///
+    /// * `enable` - `true` to enable AEC, `false` to disable
+    /// * `prefer_hardware` - `true` to prefer hardware AEC on supported devices
```
ladvoc, line 1041: suggestion (non-blocking): If setting `prefer_hardware` to false is not common, it might make sense to break it out into a separate method (e.g. `set_echo_cancellation_prefers_hardware`). Also applies for the other processing settings.

livekit/src/platform_audio/error.rs
```diff
@@ -0,0 +1,132 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Error types for platform audio operations.
+
+use std::fmt;
+
+/// Errors that can occur during audio operations.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AudioError {
```
ladvoc, line 21: suggestion: Derive `thiserror::Error` for consistency with other parts of the codebase:
```rust
#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
pub enum AudioError {
    /// Platform ADM could not be initialized.
    ///
    /// This can happen if:
    /// - No audio devices are available
    /// - Audio permissions are not granted
    /// - Platform audio subsystem is unavailable
    #[error("Failed to initialize platform audio")]
    PlatformInitFailed,

    // ...
}
```
This also gives you implementations of `fmt::Display` and `error::Error` for free, so you can delete those.

Also, it might be a good idea to call this type `PlatformAudioError` to be explicit.

examples/basic_room/Cargo.toml
```diff
@@ -10,3 +10,4 @@ env_logger = { workspace = true }
 livekit = { workspace = true, features = ["rustls-tls-native-roots"] }
 livekit-api = { workspace = true, features = ["rustls-tls-native-roots"] }
 log = { workspace = true }
+hound = "3.5"
```
ladvoc: suggestion: This should be made a workspace dependency since it is also used by `livekit-wakeword`, the `basic_room` example, and `soxr-sys` (as a dev dependency). This will ensure we only pull down one version.

livekit/src/rtc_engine/peer_transport.rs
```diff
@@ -353,6 +353,13 @@ impl PeerTransport {
         let mut offer = self.peer_connection.create_offer(options).await?;
         let mut sdp = offer.to_string();
 
+        // Log audio m-lines to debug sample rate issues
+        for line in sdp.lines() {
+            if line.starts_with("m=audio") || line.contains("opus") || line.contains("a=rtpmap") {
+                log::info!("SDP audio: {}", line);
```
ladvoc: issue: This should be removed (or made debug level) before merging.

livekit/tests/platform_audio_test.rs
```diff
@@ -0,0 +1,1597 @@
...
+mod common;
+
+use std::time::Duration;
+
+use anyhow::{anyhow, Result};
+#[cfg(feature = "__lk-e2e-test")]
+use livekit::options::TrackPublishOptions;
+use livekit::{prelude::*, AudioError, AudioResult, PlatformAudio, RtcAudioSource};
+use serial_test::serial;
+use tokio::time::timeout;
```
ladvoc: nitpick: There are a few unused imports here.

livekit-ffi/src/server/requests.rs
```diff
@@ -1367,7 +1367,377 @@ pub fn handle_request(
...
+        Request::StopRecording(req) => on_stop_recording(server, req)?.into(),
     });
 
     Ok(res)
 }
+
+// ==================== Platform Audio ====================
+
+/// FFI wrapper for PlatformAudio handle.
+pub struct FfiPlatformAudio {
```
ladvoc: issue: To follow the established pattern, all request handlers and FFI wrappers for platform audio should be moved into their own module.

livekit/src/rtc_engine/lk_runtime.rs
```diff
@@ -51,6 +54,259 @@ impl LkRuntime {
     pub fn pc_factory(&self) -> &PeerConnectionFactory {
         &self.pc_factory
     }
+
+    // ===== Device Management Methods =====
+    // These methods are primarily for FFI use. Use PlatformAudio for the public API.
```
ladvoc: issue: I do not see these methods being called from `livekit-ffi`. If they are never used directly for FFI, they should be marked `pub(crate)`.

livekit/src/platform_audio/mod.rs
xianshijing-lk: in that case, if the index is invalid, the set_recording_device(index) call will fail and return false
the code will return AudioError::OperationFailed (not InvalidDeviceIndex)

That is expected behavior.

livekit-ffi/src/server/requests.rs
xianshijing-lk: Done with moving FfiPlatformAudio and all platform audio handlers to a new module platform_audio.rs:
    - Created /Users/shijing/dev/rust-sdks/livekit-ffi/src/server/platform_audio.rs
    - Added pub mod platform_audio; to mod.rs
    - Updated requests.rs to import and call functions from the new module
    - Updated audio_source.rs to reference platform_audio::FfiPlatformAudio

livekit/src/rtc_engine/lk_runtime.rs
xianshijing-lk: Done with changing all internal device management methods from pub to pub(crate):
    - Updated comment from "for FFI use" to "internal - used by PlatformAudio"
    - Removed #[doc(hidden)] since pub(crate) makes them private outside the crate
    - Changed ~25 methods to pub(crate) including:
        - Device enumeration (playout_devices, recording_devices, etc.)
      - Device selection (set_playout_device, set_recording_device, etc.)
      - Recording/playout control (stop_recording, init_recording, etc.)
      - Audio processing (builtin_aec_is_available, enable_builtin_aec, etc.)
      - Platform ADM lifecycle (acquire_platform_adm, release_platform_adm, etc.)

livekit/src/rtc_engine/peer_transport.rs
xianshijing-lk: Done with removing it,

examples/basic_room/Cargo.toml
xianshijing-lk: done" 
Added hound = "3.5" to workspace dependencies in root Cargo.toml
  2. Updated all three packages to use hound = { workspace = true }:
    - examples/basic_room/Cargo.toml
    - livekit-wakeword/Cargo.toml
    - soxr-sys/Cargo.toml (was using 3.4, now standardized to 3.5)

livekit/src/platform_audio/mod.rs
xianshijing-lk: done with implementing a iterator-based device enumeration (livekit/src/platform_audio/mod.rs):
    - Created RecordingDeviceInfo and PlayoutDeviceInfo structs with id, name, index fields
    - Changed recording_devices() and playout_devices() to return impl Iterator<Item = ...>

examples/basic_room/src/main.rs
xianshijing-lk: done with using  clap::Parser;

examples/basic_room/src/main.rs
xianshijing-lk: Done

xianshijing-lk:
Hi @ladvoc , I think I addressed your comments, PTAL.

livekit/src/platform_audio/mod.rs
```diff
@@ -355,110 +460,108 @@ impl PlatformAudio {
...
-    /// Returns the name of a recording device by index.
+    /// Returns the number of available recording devices.
     ///
-    /// # Arguments
+    /// Prefer using [`recording_devices()`] iterator instead.
     ///
-    /// * `index` - Device index (0-based, must be < `recording_devices()`)
+    /// [`recording_devices()`]: Self::recording_devices
+    #[doc(hidden)]
+    pub fn recording_device_count(&self) -> i16 {
```
ladvoc: suggestion: cast to and return `usize` to match the Rust convention for representing indices/lengths. Also applies to `playout_device_count`, the index-based accessors, and the info structs.

livekit/src/platform_audio/mod.rs
```diff
@@ -355,110 +460,108 @@ impl PlatformAudio {
...
-    ///
-    /// ```rust,ignore
-    /// let audio = PlatformAudio::new()?;
-    /// for i in 0..audio.playout_devices() as u16 {
-    ///     println!("Speaker {}: {}", i, audio.playout_device_name(i));
-    /// }
-    /// ```
+    /// Prefer using [`playout_devices()`] iterator instead.
+    #[doc(hidden)]
     pub fn playout_device_name(&self, index: u16) -> String {
```
ladvoc: issue: Now that there is an iterator-based API in place, I would strongly consider making these accessors internal to minimize API surface. If looking up a specific device by index is important, this could be turned into `pub fn playout_device_for_index(&self, index: usize) → Option<PlayoutDeviceInfo>` (the iterator based API could call this inside of `.map` too).

livekit/src/platform_audio/mod.rs
ladvoc: issue: It looks like these methods that don't accept new types have been left here for FFI support. A better pattern would be to remove them and add a constructor on the new types to initialize from the raw value (explicitly marked as unchecked, only to be used for FFI):
```rust
impl PlayoutDeviceId {
   #[doc(hidden)]
   pub fn from_unchcked_guid(guid: &str) → Self {
       Self(guid.to_string())
   }
}
```

Then FFI can still construct as needed and use the existing set/switch methods:
```rust
match ffi_audio.audio.set_recording_device(PlayoutDeviceId::from_unchecked_guid(&req.device_id)) {
 // ...
}
```

ladvoc (commented):
This is looking great, just a few more comments.

livekit/src/platform_audio/mod.rs
xianshijing-lk: Good idea, done with addressing the comment.

livekit/src/platform_audio/mod.rs
xianshijing-lk: Done with using usize.

livekit/src/platform_audio/mod.rs
xianshijing-lk: Done with changing the hidden public index/count accessors to internal helpers and kept recording_devices() / playout_devices() as the public enumeration API.

webrtc-sys/include/livekit/peer_connection_factory.h
```diff
@@ -57,15 +57,81 @@ class PeerConnectionFactory {
...
+      rust::String label) const;
+
   RtpCapabilities rtp_sender_capabilities(MediaType type) const;
 
   RtpCapabilities rtp_receiver_capabilities(MediaType type) const;
 
   std::shared_ptr<RtcRuntime> rtc_runtime() const { return rtc_runtime_; }
 
+  // Device enumeration
+  int16_t playout_devices() const;
```
cloudwebrtc: Should we move the following method to the audio_device.rs file? And add a `PeerConnectionFactory::audio_device() -> SharedPtr<AudioDevice>` pointing to (AudioDevice/AdmProxy)?

This isn't a strong requirement, but it will make the webrtc-sys codebase look clearer.

livekit/src/platform_audio/mod.rs
```diff
@@ -0,0 +1,1126 @@
...
+    /// let audio = PlatformAudio::new()?;
+    /// // ... use audio ...
+    /// audio.release();  // ADM disabled if this was the last reference
+    /// ```
+    pub fn release(self) {
+        drop(self);
+    }
+
+    // =========================================================================
+    // Audio Processing (AEC, AGC, NS)
```
ladvoc, line 872: nitpick: Prefer grouping related methods into separate `impl` blocks. `impl` blocks can also have doc comments attached to them, and each block will get it's own section in the generated docs to make it easier for developers to find things. Preview generated docs locally with `cargo doc --no-deps` from within the _livekit_ crate. Example: 
```rust
/// Audio Processing (AEC, AGC, NS)
impl PlatformAudio {
    // ...
}
```

livekit/src/platform_audio/mod.rs
```diff
@@ -0,0 +1,1126 @@
...
+/// ```rust,ignore
+/// use livekit::{PlatformAudio, reset_platform_audio};
+///
+/// let audio = PlatformAudio::new()?;
+/// // ... use audio ...
+///
+/// // Reset handle references
+/// reset_platform_audio();
+/// ```
+pub fn reset_platform_audio() {
```
ladvoc, line 1112: suggestion (non-blocking): Maybe make this an associated function (no `&self`) on `PlatformAudio` to avoid the extra import for the user:
```rust
impl PlatformAudio {
   pub fn reset() {
      // ...
   }
}
```

User can invoke with `PlatformAudio::reset()`.

webrtc-sys/src/adm_proxy.cpp
```diff
@@ -0,0 +1,848 @@
...
+
+AdmProxy::AdmProxy(const webrtc::Environment& env, webrtc::Thread* worker_thread)
+    : env_(env),
+      worker_thread_(worker_thread) {
+  RTC_LOG(LS_INFO) << "AdmProxy::AdmProxy() - Initializing";
+
+  // Create synthetic ADM for synthetic mode.
+  // This pumps the WebRTC audio pipeline without platform audio,
+  // allowing FFI callbacks to receive decoded remote audio.
+  synthetic_adm_ = webrtc::make_ref_counted<AudioDevice>(env_);
```
cloudwebrtc: Should we rename `AudioDevice` to `DummyAudioDevice/dummy_audio_device.cpp`? (maybe SyntheticAudioDevice)

ladvoc (approved):
LGTM ✅

ladvoc (approved):
LGTM ✅

ladvoc:
Last thing I just noticed: `livekit-ffi` is missing from the changeset.

cloudwebrtc (approved):
lgtm!

xianshijing-lk:
> Last thing I just noticed: `livekit-ffi` is missing from the changeset.

Done with adding livekit-ffi to the changeset and also added 
**FFI platform audio support**: Expose platform audio device enumeration and selection through `livekit-ffi`

webrtc-sys/src/adm_proxy.cpp
xianshijing-lk: Good idea, SyntheticAudioDevice is clearer than AudioDevice.

Done with renaming it.

webrtc-sys/include/livekit/peer_connection_factory.h
xianshijing-lk: hmm, I got where this comment comes from. 

I refactor it into a audio_device_controller.cpp and audio_device_controller.r, and move those device handling code to such audio_device_controller so that the peer connection factory will be cleaner. 
It is a larger refactoring than what I initially expected it to be.

The layers now look like this:

  - PlatformAudio
  - LkRuntime
  - libwebrtc::PeerConnectionFactory
  - webrtc-sys::PeerConnectionFactory
  - webrtc-sys::AudioDeviceController
  - AdmProxy

  What each layer does:

  - PlatformAudio
      - public Rust API for device enumeration, selection, switching, and platform audio lifecycle
  - LkRuntime
      - internal LiveKit runtime wrapper
      - forwards platform-audio operations to libwebrtc
  - libwebrtc::PeerConnectionFactory
      - Rust wrapper over the native bridge
      - still owns the main sys factory handle
      - now also caches one sys audio-device controller handle
  - webrtc-sys::PeerConnectionFactory
      - native C++ factory object
      - owns the real shared audio-device state via AdmProxy
      - creates and caches one AudioDeviceController
  - webrtc-sys::AudioDeviceController
      - focused facade for audio-device/ADM operations only
      - no independent audio state of its own
  - AdmProxy
      - the actual implementation behind device enumeration, selection, playout/recording control, and platform ADM activation

  How AdmProxy is shared:

  - webrtc-sys::PeerConnectionFactory constructs AdmProxy once in its constructor
  - it stores that in adm_proxy_
  - it also constructs one AudioDeviceController with the same adm_proxy_
  - both hold references to the same AdmProxy

  So the ownership/reference picture is:

  - PeerConnectionFactory
      - owns adm_proxy_
      - owns cached audio_device_
  - AudioDeviceController
      - holds a scoped_refptr<AdmProxy>

livekit-ffi-node-bindings/proto/audio_frame_pb.d.ts
ladvoc: @lukasIO moved the Node bindings to this repo a month or two ago (I don't remember the exact reasons). But yes, the generated Node bindings get committed. However, they are automatically generated by CI so you shouldn't have to worry about generating/committing manually.

xianshijing-lk:
Landing now

PR #1044, opened 2026-04-23
fix(signal-client): carry access token on validate() fallback

### LINKED ISSUE #1042: signal_client::validate() drops the access_token — fabricates a 401 on any non-404 connection error (e.g., server returns 503)
When `SignalStream::connect` returns any non-404 error on the initial signal WebSocket — for example a `503` from a node whose `num_tracks` limit is reached — the SDK's error-handling path in `SignalInner::connect` calls [`Self::validate(lk_url)`](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/signal_client/mod.rs#L325) to "retrieve more information". That `validate()` function ([`mod.rs:349-370`](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/signal_client/mod.rs#L349-L370)) issues a plain `http_client::get(validate_url)`. On the tokio build, [`http_client::get` is a re-export of `reqwest::get`](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/http_client.rs#L18) — a helper that attaches no headers. The access token is attached only to the WebSocket handshake `Authorization: Bearer` header inside [`signal_stream.rs:105-107`](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/signal_client/signal_stream.rs#L105-L107); it is never written into the URL. So the `/rtc/validate` request arrives at the server with no credentials, the server's claims check returns `401 "no permissions to access the room"`, and the SDK surfaces that 401 as the terminal error — **masking the real 503 and reporting a misleading auth error to the caller**.

The engine then retries up to 3 times. Each retry produces the same 503→401 pair, for a deterministic pattern of **4 × `/rtc` requests + 4 × `/rtc/validate` requests per failed connect attempt**, terminating in:

```
engine: signal failure: client error: 401 Unauthorized - no permissions to access the room
```

## Impact

We hit this in production when a batch of concurrent interviews drove our self-hosted livekit-server pods to `num_tracks=80` saturation. Clients that should have seen "server is at capacity, retry shortly" instead saw "no permissions to access the room" — we spent hours chasing a phantom auth bug in our token-minting code before the repro below pinned the fault to the SDK.

More broadly, any operator whose livekit-server returns a 503 (node limits, maintenance drain, LB rejection, anything other than 404) will see their operators' bots emit 401s instead of 503s, derailing both human and automated diagnostics.

## Environment

| Version | Status |
|---|---|
| `livekit==1.0.25` (Python, shipped 2026-01-30) | reproduces |
| `livekit==1.1.6` (Python, shipped 2026-04-16) | reproduces |
| rust-sdks HEAD (`main` @ commit `3b01caf8b68cdaeaf3dc7ef7ed53597a020736de`, 2026-04-23) | code path unchanged, bug still present |

Tested OS: Ubuntu 24.04, Python 3.12.

## Reproducer

A single file, stdlib only (apart from `livekit` itself), no docker/no real livekit-server. The script starts an HTTP server on an ephemeral port that returns `503` to `/rtc` and `401` to `/rtc/validate`, drives `rtc.Room().connect(...)` against it, captures every HTTP request, and classifies them.

```bash
python3 -m venv .venv && source .venv/bin/activate
pip install 'livekit==1.1.6' 'livekit-api>=1.1.0'
# save the repro.py from the collapsible section below
python repro.py
```

Exit codes: `1` = bug reproduced, `0` = fixed, `2` = inconclusive.

<details><summary>Full <code>repro.py</code> (self-contained, stdlib + <code>livekit</code> only)</summary>

```python
"""Self-contained reproducer for a bug in livekit-rtc's signal client.

Bug
---
When the signal-server returns any non-404 error on the initial WebSocket
upgrade (e.g. 503 when the node is at capacity), the Rust SDK falls back
to calling GET /rtc/validate "to retrieve more information" (mod.rs:325).

That fallback is issued via http_client::get(url), which on Tokio is a
re-export of reqwest::get (http_client.rs:18). reqwest::get attaches no
Authorization header, and the URL constructed by get_validate_url never
contained the access_token (the token is attached to the WS handshake as
an Authorization: Bearer header in signal_stream.rs:107, not to the URL).

So the /rtc/validate call arrives at the server with no credentials, the
server returns 401 "no permissions to access the room", and the SDK
surfaces that 401 as the terminal error — masking the real 503.

Reproducer
----------
This script starts a trivial stdlib HTTP server on localhost that:
  - responds 503 on GET /rtc            (mimics a saturated node)
  - logs every inbound request          (so we can see what the SDK sends)

Then it uses the livekit Python SDK to attempt room.connect() against
that server. The SDK's error cascade makes it hit /rtc/validate; we
print the captured bytes and assert that no auth is present.

Requirements
------------
Python 3.9+, and:

    pip install "livekit==1.1.6" "livekit-api>=1.1.0"

Run
---
    python repro.py

Expected output
---------------
The script terminates with a non-zero exit code and prints:

    [REPRODUCED] 4 x /rtc 503 + 4 x /rtc/validate (no Authorization,
                  no access_token query) -> terminal 401.

If the bug is fixed, /rtc/validate requests should carry the token and
the script will print:

    [NOT REPRODUCED] /rtc/validate requests carry Authorization header.

Tested against
--------------
livekit==1.0.25  (bug present)
livekit==1.1.6   (bug present — shipped 2026-04-16)
main HEAD        (bug present — commit 3b01caf, 2026-04-23)

Source lines
------------
- livekit-api/src/signal_client/mod.rs:319  (first call site)
- livekit-api/src/signal_client/mod.rs:325  (main call site, non-404 errors)
- livekit-api/src/signal_client/mod.rs:349  (validate() body)
- livekit-api/src/signal_client/mod.rs:354  (the auth-less GET)
- livekit-api/src/http_client.rs:18         (reqwest::get re-export)
- livekit-api/src/signal_client/signal_stream.rs:105-107
    (where the token IS attached — to the WS handshake only)
"""

from __future__ import annotations

import asyncio
import datetime
import http.server
import socket
import sys
import threading
from typing import List, Tuple
from urllib.parse import parse_qs

try:
    from livekit import api, rtc
except ImportError as e:
    sys.exit(
        "This reproducer needs `livekit` and `livekit-api`:\n"
        "    pip install 'livekit==1.1.6' 'livekit-api>=1.1.0'\n"
        f"({e})"
    )


# ----------------------------- mock server ---------------------------------


# Each item: (path, query_params_dict, headers_dict, raw_first_line)
captured: List[Tuple[str, dict, dict, str]] = []


class _Handler(http.server.BaseHTTPRequestHandler):
    # HTTP/1.1 keeps the Rust signal_client's parser happy (the default
    # HTTP/1.0 triggers a noisy `WrongHttpVersion` warning in the SDK logs,
    # though the bug still reproduces either way).
    protocol_version = "HTTP/1.1"

    # Silence the default access-log noise so our output is clean.
    def log_message(self, *args, **kwargs):  # noqa: D401
        return

    def _capture(self):
        request_line = self.requestline  # e.g. "GET /rtc?... HTTP/1.1"
        parts = self.path.split("?", 1)
        path = parts[0]
        qs = parse_qs(parts[1]) if len(parts) > 1 else {}
        qs_flat = {k: v[0] if v else "" for k, v in qs.items()}
        headers = {k.lower(): v for k, v in self.headers.items()}
        captured.append((path, qs_flat, headers, request_line))
        return path

    def do_GET(self):  # noqa: N802
        path = self._capture()
        if path == "/rtc":
            # Simulate a saturated node. Matches the exact prod response body.
            body = b"node has exceeded its configured limit"
            self.send_response(503)
            self.send_header("Content-Type", "text/plain")
            self.send_header("Content-Length", str(len(body)))
            self.send_header("Connection", "close")
            self.end_headers()
            self.wfile.write(body)
        elif path == "/rtc/validate":
            # What a real livekit-server returns for claims == nil. We respond
            # 401 with the same error body that the server emits when the token
            # is missing — so the SDK's terminal error message matches prod.
            body = b"no permissions to access the room"
            self.send_response(401)
            self.send_header("Content-Type", "text/plain")
            self.send_header("Content-Length", str(len(body)))
            self.send_header("Connection", "close")
            self.end_headers()
            self.wfile.write(body)
        else:
            self.send_response(404)
            self.send_header("Content-Length", "0")
            self.end_headers()


def _start_server() -> Tuple[str, int, http.server.HTTPServer]:
    """Start the mock server on a free localhost port and return (url, port, server)."""
    # Pick any free port so the repro doesn't clash with other services.
    with socket.socket() as s:
        s.bind(("127.0.0.1", 0))
        port = s.getsockname()[1]
    srv = http.server.HTTPServer(("127.0.0.1", port), _Handler)
    threading.Thread(target=srv.serve_forever, daemon=True).start()
    return f"http://127.0.0.1:{port}", port, srv


# ----------------------------- livekit client ------------------------------


# These are arbitrary. The mock server doesn't verify anything; the SDK
# does signs-its-own-JWT locally and sends it in the WS Authorization header.
_KEY = "APIxxxxxxxxxxxxx"
_SECRET = "secretxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"


def _mint_token(room: str, identity: str) -> str:
    return (
        api.AccessToken(_KEY, _SECRET)
        .with_identity(identity)
        .with_grants(
            api.VideoGrants(
                room_join=True,
                room=room,
                can_publish=True,
                can_subscribe=True,
            )
        )
        .with_ttl(datetime.timedelta(hours=1))
        .to_jwt()
    )


async def _drive_sdk(ws_url: str):
    token = _mint_token("repro-room", "repro-identity")
    room = rtc.Room()
    try:
        await asyncio.wait_for(room.connect(ws_url, token), timeout=30)
        print("[unexpected] room.connect returned without error")
    except Exception as e:
        # Expected — the 503 cascade terminates in a 401 error.
        print(f"[sdk] room.connect raised (expected): {type(e).__name__}: {e}")
    finally:
        try:
            await room.disconnect()
        except Exception:
            pass


# ----------------------------- classifier ----------------------------------


def _classify():
    rtc_requests = [c for c in captured if c[0] == "/rtc"]
    validate_requests = [c for c in captured if c[0] == "/rtc/validate"]

    print()
    print("=" * 70)
    print(f"Captured {len(captured)} HTTP request(s) total")
    print(f"  /rtc          : {len(rtc_requests)}")
    print(f"  /rtc/validate : {len(validate_requests)}")
    print("=" * 70)

    print("\n--- detail ---")
    for i, (path, qs, headers, line) in enumerate(captured, 1):
        print(f"\n#{i}  {line.strip()}")
        has_auth = "authorization" in headers and headers["authorization"].lower().startswith("bearer ")
        has_qs_token = bool(qs.get("access_token"))
        auth_preview = headers.get("authorization", "(none)")
        if has_auth and len(auth_preview) > 48:
            auth_preview = auth_preview[:40] + "... [truncated]"
        print(f"    Authorization header : {auth_preview}")
        print(f"    access_token query   : {'<JWT present>' if has_qs_token else '(none)'}")

    # A /rtc/validate request is "buggy" if the SDK sent no token via any
    # channel — the call cannot succeed against a real livekit server, whose
    # claims check rejects requests with no auth and returns the exact error
    # string "no permissions to access the room".
    buggy = [
        r for r in validate_requests
        if not r[2].get("authorization", "").lower().startswith("bearer ")
        and not r[1].get("access_token")
    ]

    print("\n" + "=" * 70)
    if validate_requests and buggy == validate_requests:
        print(
            f"[REPRODUCED] {len(rtc_requests)} x /rtc 503 + {len(validate_requests)} "
            f"x /rtc/validate (no Authorization, no access_token query) -> terminal 401."
        )
        print("The SDK's Self::validate(lk_url) path in livekit-api/src/signal_client/mod.rs")
        print("issues a plain GET via http_client::get (= reqwest::get on tokio), which")
        print("attaches no auth. Server sees claims==nil and returns 401.")
        return 1
    elif not validate_requests:
        print("[INCONCLUSIVE] SDK did not call /rtc/validate. Did the 503 reach the SDK?")
        return 2
    else:
        print("[NOT REPRODUCED] /rtc/validate requests carry Authorization header or")
        print("access_token query. This version of livekit-rtc appears to have fixed")
        print("the missing-token fallback.")
        return 0


# ----------------------------- entrypoint ----------------------------------


async def main() -> int:
    base_url, port, srv = _start_server()
    print(f"[server] mock livekit listening on {base_url}")
    print(f"[server] will 503 on /rtc and 401 'no permissions' on /rtc/validate\n")

    # Give the HTTP server a tick to bind cleanly.
    await asyncio.sleep(0.1)

    try:
        await _drive_sdk(base_url.replace("http://", "ws://"))
        # Let any in-flight retries settle.
        await asyncio.sleep(1.0)
    finally:
        srv.shutdown()
        srv.server_close()

    return _classify()


if __name__ == "__main__":
    try:
        code = asyncio.run(main())
    except KeyboardInterrupt:
        code = 130
    sys.exit(code)
```
</details>

### What the Rust engine logs

```
livekit_api::signal_client:287 - signal connection failed on v0 path:
    WsError(Http(Response { status: 503, ..., body: "node has exceeded its configured limit" }))
livekit::rtc_engine:434 - failed to connect: Signal(Client(401, "no permissions to access the room")), retrying... (1/3)
livekit_api::signal_client:287 - signal connection failed on v0 path:
    WsError(Http(Response { status: 503, ..., body: "node has exceeded its configured limit" }))
livekit::rtc_engine:434 - failed to connect: Signal(Client(401, "no permissions to access the room")), retrying... (2/3)
livekit_api::signal_client:287 - signal connection failed on v0 path:
    WsError(Http(Response { status: 503, ..., body: "node has exceeded its configured limit" }))
livekit::rtc_engine:434 - failed to connect: Signal(Client(401, "no permissions to access the room")), retrying... (3/3)
livekit_api::signal_client:287 - signal connection failed on v0 path:
    WsError(Http(Response { status: 503, ... }))
livekit_ffi::server::room:287 - error while connecting to a room:
    engine: signal failure: client error: 401 Unauthorized - no permissions to access the room
```

Note the 503 is received on line 287 of `signal_client`, but the very next line reports a 401 as the retry-able reason. The 401 is synthesised by the SDK's own `validate()` call, not sent by the server in response to the `/rtc` upgrade.

### What the mock server sees (from `repro.py`)

```
Captured 8 HTTP request(s) total
  /rtc          : 4
  /rtc/validate : 4

#1  GET /rtc?sdk=python&...&version=1.1.6 HTTP/1.1
    Authorization header : Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...  ← token present
    access_token query   : (none)

#2  GET /rtc/validate?sdk=python&...&version=1.1.6 HTTP/1.1
    Authorization header : (none)                                          ← token MISSING
    access_token query   : (none)

... (pattern repeats 3 more times: #3+#4, #5+#6, #7+#8)
```

The `/rtc` request carries `Authorization: Bearer <jwt>`. The `/rtc/validate` request that immediately follows carries neither the header nor an `access_token` query param.

### Expected behaviour

Either:
1. `Self::validate()` should forward the access token (simplest — propagate the `token: &str` that is already available at the call sites in `SignalInner::connect`, and attach it as `Authorization: Bearer <token>` on the GET), or
2. `Self::validate()` should not be called at all when the original error already carries actionable status information (e.g., `WsError::Http(status)` where the status is itself meaningful — 503 is already a clearer signal than 401 would be), or
3. The SDK should preserve and surface the *original* error (the 503) rather than overwriting it with the validate-call result.

## Root cause

Three source references — all on `main` at `3b01caf`.

### 1. `mod.rs:325` — the fallback call site

https://github.com/livekit/rust-sdks/blob/3b01caf8b68cdaeaf3dc7ef7ed53597a020736de/livekit-api/src/signal_client/mod.rs#L317-L328

```rust
} else {
    // Connection failed, try to retrieve more information
    Self::validate(lk_url).await?;
    return Err(err);
}
```

Fires on every non-404 connection error. The `token: &str` is in scope at this call site but is not passed to `validate`.

### 2. `mod.rs:349-370` — `validate()` itself

https://github.com/livekit/rust-sdks/blob/3b01caf8b68cdaeaf3dc7ef7ed53597a020736de/livekit-api/src/signal_client/mod.rs#L349-L370

```rust
/// Validate the connection by calling rtc/validate
async fn validate(ws_url: url::Url) -> SignalResult<()> {
    let validate_url = get_validate_url(ws_url);

    let validate_fut = async {
        if let Ok(res) = http_client::get(validate_url.as_str()).await {
            //                          ^^^ no auth, no headers at all
            let status = res.status();
            ...
```

### 3. `http_client.rs:18` — the `get` implementation has no auth hook

https://github.com/livekit/rust-sdks/blob/3b01caf8b68cdaeaf3dc7ef7ed53597a020736de/livekit-api/src/http_client.rs#L18

```rust
#[cfg(feature = "signal-client-tokio")]
pub use reqwest::get;
```

`reqwest::get(url)` is documented as "a shortcut for `reqwest::Client::new().get(url).send()`" — it attaches no `Authorization` header by design.

### And where the token does live

https://github.com/livekit/rust-sdks/blob/3b01caf8b68cdaeaf3dc7ef7ed53597a020736de/livekit-api/src/signal_client/signal_stream.rs#L99-L107

```rust
async fn connect_inner(
    url: url::Url,
    token: &str,
) -> SignalResult<...> {
    log::info!("connecting to {}", url);
    let mut request = url.clone().into_client_request()?;
    let auth_header = HeaderValue::from_str(&format!("Bearer {token}"))
        .map_err(|_| SignalError::TokenFormat)?;
    request.headers_mut().insert(AUTHORIZATION, auth_header);
```

The token is inserted as a header on the WebSocket handshake `Request` object. It is not present on the `url::Url` value that `get_validate_url(ws_url)` later receives. `get_validate_url` does nothing to re-attach it.

## Suggested fix

Thread the token through `validate()`. Sketch (untested, intended to convey shape):

```diff
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -316,7 +316,7 @@
                             if let SignalError::TokenFormat = err {
                                 return Err(err);
                             }
-                            Self::validate(lk_url_v0).await?;
+                            Self::validate(lk_url_v0, token).await?;
                             return Err(err);
                         }
                     }
                 } else {
                     // Connection failed, try to retrieve more information
-                    Self::validate(lk_url).await?;
+                    Self::validate(lk_url, token).await?;
                     return Err(err);
                 }
@@ -347,11 +347,11 @@
     }

     /// Validate the connection by calling rtc/validate
-    async fn validate(ws_url: url::Url) -> SignalResult<()> {
+    async fn validate(ws_url: url::Url, token: &str) -> SignalResult<()> {
         let validate_url = get_validate_url(ws_url);

         let validate_fut = async {
-            if let Ok(res) = http_client::get(validate_url.as_str()).await {
+            if let Ok(res) = http_client::get_with_bearer(validate_url.as_str(), token).await {
                 let status = res.status();
```

…together with a small `http_client::get_with_bearer` helper that uses `reqwest::Client::new().get(url).bearer_auth(token).send().await` on tokio and the isahc equivalent on async-std.

Alternative (less invasive): append `access_token=<jwt>` to `validate_url` inside `validate()`. That works because the server accepts the token either as the `Authorization` header *or* as the `access_token` query param.

## Non-goal: this is not a server-side fix

The server behaviour (returning 401 when `claims == nil`) is correct and documented. The bug is that the SDK is generating claims-less requests on a failure path; fixing the server would mask the real issue.

## Notes on the reproducer

- Single file, stdlib HTTP server only — no docker, no real livekit-server.
- Tested against Python 3.12 and livekit 1.0.25 + 1.1.6. Should work on any version that exposes the same signal-client internals.
- Will automatically classify: prints `[REPRODUCED]` and exits 1 if any `/rtc/validate` request lacks both the header and the query param; prints `[NOT REPRODUCED]` and exits 0 otherwise.

Happy to iterate on the fix or land a PR if this is interesting to the maintainers.

abhisheksingh-R41:
### Before you submit your PR

### PR description

Fixes #1042.

When `SignalInner::connect`'s primary WS upgrade fails with any non-404 error, the code calls `Self::validate(lk_url)` "to retrieve more information" ([mod.rs:319](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/signal_client/mod.rs#L319), [:325](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/signal_client/mod.rs#L325)). `validate()` then issues `http_client::get(url)` — which on tokio is a bare `reqwest::get` re-export ([http_client.rs:18](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/http_client.rs#L18)) and on async-std is `isahc::get_async`. Neither attaches auth.

The access token is only ever written to the WS handshake as an `Authorization: Bearer` header inside [`signal_stream.rs:105-107`](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/signal_client/signal_stream.rs#L105-L107) — never into the URL. So the validate GET arrives at the server with no credentials, the server's claims check returns 401 "no permissions to access the room", and `validate()` surfaces that 401 as the terminal error. In effect: any non-404 failure on the primary path — including a 503 from a saturated node — gets rewritten into a fabricated 401 at the caller.

#### What this PR does

1. Thread `token: &str` through `SignalInner::validate` and pass it from both call sites in `connect()`.
2. Add `http_client::get_with_token(url, token)` in both the tokio (reqwest `.bearer_auth()`) and async-std (isahc `Authorization` header) branches of `http_client.rs`.
3. Switch `validate()` from `http_client::get` to `get_with_token`.
4. Remove the now-unused `http_client::get` exports — `grep -rn 'http_client::get\b' livekit-api/src/` returns zero other callers.

Auth is sent as `Authorization: Bearer <token>`, matching the precedent in `signal_stream.rs` (WS handshake), `region::fetch_from_endpoint` ([region.rs:54](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/signal_client/region.rs#L54)), and the Twirp clients in `services/*`. The access token is a JWT; keeping it out of request URLs avoids leaking it into any log that records paths (LB access logs, proxies, `jsonPayload.path` in server error logs, etc.).

### Testing

New regression test `validate_sends_bearer_token` in the existing `signal_client::tests` module. It:

- Binds a `tokio::net::TcpListener` on an ephemeral port (matches the existing `signal_stream_connect_timeout` / `region_fetch_parses_response` test pattern).
- Captures the first request bytes sent to that port via a `oneshot` channel.
- Calls `SignalInner::validate(ws_url, "test-bearer-token").await`.
- Asserts the captured request contains `authorization: bearer test-bearer-token` (case-insensitive per RFC 9110).

The assertion checks the **value** of the token, not just presence of the header — catches both "no auth" and "wrong auth" regressions.

Verified locally:
- Before applying the fix (restored `http_client::get(validate_url)` without the token), the new test **FAILS** with the captured request body showing no `authorization` header.
- With the fix applied, the new test **PASSES** and all existing tests (12/12) continue to pass: `cargo +nightly test --lib --features signal-client-tokio`.
- `cargo +nightly-2023-12-30 fmt -- --check` clean (the toolchain used by the format CI job).
- Clippy surfaces no new warnings on changed code.

A reproducer outside the test harness (stdlib HTTP server + Python `livekit` SDK) is attached to issue #1042 for anyone who wants to see the 503 → fabricated-401 cascade end-to-end against a real client.

### Breaking changes

None. `SignalInner::validate` is a private async method on a private struct — the added `token: &str` parameter is not observable outside the crate. `http_client` is declared `mod http_client;` (not `pub mod`), so the removed `get` exports weren't reachable from outside either.

### MSRV

No MSRV change. `reqwest::RequestBuilder::bearer_auth` has been available since reqwest 0.3; `isahc::Request::get().header().body()` is standard since isahc 1.0.

### DIFF
--- a/.changeset/fix_signal_client_validate_missing_auth.md
+++ b/.changeset/fix_signal_client_validate_missing_auth.md
@@ -1,0 +1,5 @@
+---
+livekit-api: patch
+---
+
+fix(signal-client): carry access token on validate() fallback - #1044 (@abhisheksingh-R41)
--- a/livekit-api/src/http_client.rs
+++ b/livekit-api/src/http_client.rs
@@ -14,11 +14,19 @@
 
 #[cfg(any(feature = "services-tokio", feature = "signal-client-tokio"))]
 mod tokio {
-    #[cfg(feature = "signal-client-tokio")]
-    pub use reqwest::get;
-
     #[cfg(feature = "services-tokio")]
     pub use reqwest::Client;
+
+    /// GET with an `Authorization: Bearer <token>` header attached.
+    ///
+    /// `reqwest::get(url)` (the free function) constructs a fresh default
+    /// `Client` and attaches no auth, which is why `SignalInner::validate` —
+    /// the sole caller — uses this helper: the access token must reach the
+    /// server or the server returns 401 regardless of the underlying error.
+    #[cfg(feature = "signal-client-tokio")]
+    pub async fn get_with_token(url: &str, token: &str) -> reqwest::Result<reqwest::Response> {
+        reqwest::Client::new().get(url).bearer_auth(token).send().await
+    }
 }
 
 #[cfg(any(feature = "services-tokio", feature = "signal-client-tokio"))]
@@ -56,8 +64,18 @@ mod async_std {
             }
         }
 
-        pub async fn get(url: &str) -> io::Result<Response> {
-            let response = isahc::get_async(url).await?;
+        /// GET with an `Authorization: Bearer <token>` header attached.
+        ///
+        /// `isahc::get_async(url)` attaches no auth, which is why
+        /// `SignalInner::validate` — the sole caller — uses this helper: the
+        /// access token must reach the server or the server returns 401
+        /// regardless of the underlying error. Mirrors the tokio variant.
+        pub async fn get_with_token(url: &str, token: &str) -> io::Result<Response> {
+            let request = isahc::Request::get(url)
+                .header("Authorization", format!("Bearer {}", token))
+                .body(())
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+            let response = isahc::send_async(request).await?;
             Ok(Response(response))
         }
     }
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -316,13 +316,13 @@ impl SignalInner {
                                 if let SignalError::TokenFormat = err {
                                     return Err(err);
                                 }
-                                Self::validate(lk_url_v0).await?;
+                                Self::validate(lk_url_v0, token).await?;
                                 return Err(err);
                             }
                         }
                     } else {
                         // Connection failed, try to retrieve more information
-                        Self::validate(lk_url).await?;
+                        Self::validate(lk_url, token).await?;
                         return Err(err);
                     }
                 }
@@ -346,12 +346,21 @@ impl SignalInner {
         Ok((inner, join_response, events))
     }
 
-    /// Validate the connection by calling rtc/validate
-    async fn validate(ws_url: url::Url) -> SignalResult<()> {
+    /// Validate the connection by calling rtc/validate.
+    ///
+    /// This is called from `connect()` when the primary WebSocket upgrade fails
+    /// with a non-404 status, to surface a clearer HTTP-level error than the WS
+    /// upgrade error. The access token is sent as `Authorization: Bearer <token>`
+    /// so the server can actually authenticate the request; without it, the
+    /// server returns 401 "no permissions to access the room" regardless of
+    /// what the original error was, masking the real cause (e.g. a 503 from a
+    /// saturated node becomes a fabricated 401 to the caller). See
+    /// https://github.com/livekit/rust-sdks/issues/1042.
+    async fn validate(ws_url: url::Url, token: &str) -> SignalResult<()> {
         let validate_url = get_validate_url(ws_url);
 
         let validate_fut = async {
-            if let Ok(res) = http_client::get(validate_url.as_str()).await {
+            if let Ok(res) = http_client::get_with_token(validate_url.as_str(), token).await {
                 let status = res.status();
                 let body = res.text().await.ok().unwrap_or_default();
 
@@ -791,6 +800,58 @@ mod tests {
         // Should be /rtc/validate, not /rtc/rtc/validate
         assert_eq!(validate_url.path(), "/rtc/validate");
         assert_eq!(validate_url.scheme(), "https");
+    }
+
+    /// Regression test for https://github.com/livekit/rust-sdks/issues/1042.
+    ///
+    /// Prior to the fix, `SignalInner::validate` issued an auth-less GET to
+    /// `/rtc/validate`, so a server performing the standard auth check (claims
+    /// parsing) would return 401 regardless of the real reason the WebSocket
+    /// upgrade had failed. That 401 then masked the original error (e.g. a
+    /// 503 from a saturated node) at the caller.
+    ///
+    /// This test binds a TCP listener, calls `validate()`, captures the first
+    /// request sent to that listener, and asserts the request carries an
+    /// `Authorization: Bearer <token>` header.
+    #[cfg(feature = "signal-client-tokio")]
+    #[tokio::test]
+    async fn validate_sends_bearer_token() {
+        use tokio::io::{AsyncReadExt, AsyncWriteExt};
+        use tokio::net::TcpListener;
+
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+
+        let (tx, rx) = tokio::sync::oneshot::channel::<Vec<u8>>();
+
+        tokio::spawn(async move {
+            let (mut socket, _) = listener.accept().await.unwrap();
+            let mut buf = vec![0u8; 4096];
+            let n = socket.read(&mut buf).await.unwrap();
+            buf.truncate(n);
+            let _ = tx.send(buf);
+
+            // Respond 200 OK so `validate()` returns Ok() rather than
+            // surfacing a synthetic client/server error. The purpose of this
+            // test is the request side, not the response side.
+            let response = b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\nConnection: close\r\n\r\n";
+            let _ = socket.write_all(response).await;
+        });
+
+        let ws_url = url::Url::parse(&format!("ws://127.0.0.1:{}/rtc", addr.port())).unwrap();
+        let result = SignalInner::validate(ws_url, "test-bearer-token").await;
+        assert!(result.is_ok(), "expected Ok from validate, got: {:?}", result);
+
+        let request = rx.await.expect("server task never received a request");
+        let request = String::from_utf8_lossy(&request);
+
+        // Header names are case-insensitive per RFC 9110; reqwest emits the
+        // canonical `Authorization` but we normalise both for robustness.
+        assert!(
+            request.to_lowercase().contains("authorization: bearer test-bearer-token"),
+            "validate() must attach the access token as a Bearer header; request was:\n{}",
+            request
+        );
     }
 
     #[cfg(feature = "signal-client-tokio")]


reenboog (approved):
Looks ok, but fix this, please:

`No changeset found. Add a changeset or apply the 'internal' label.`

abhisheksingh-R41:
> Looks ok, but fix this, please:
> 
> `No changeset found. Add a changeset or apply the 'internal' label.`

@reenboog added the changeset.

PR #1047, opened 2026-04-23
fix race in download_webrtc to reduce flaky build

hechen-eng:
### Before you submit your PR

### PR description

It seems we have a flaky [build](https://github.com/livekit/rust-sdks/actions/runs/24849766517/job/72746715362#step:11:1003).
`webrtc_sys_build::download_webrtc()` extracts the downloaded zip via:                                                                                                                                     
  
```rust
archive.extract(webrtc_dir.parent().unwrap())?;
```

Zip extraction creates webrtc_dir/ as soon as its first entry is written, but LICENSE.md and other files land in an arbitrary order over the next few hundred milliseconds. During that window, webrtc_dir.exists() returns true but the directory is only partially populated.                                                                                                                                                             

Concurrent build scripts (livekit-ffi/build.rs, webrtc-sys/build.rs) check webrtc_dir.exists() outside the flock as a fast-path, observe the half-extracted state, skip calling the locked downloader, and then fail when they try to read a specific file that hasn't been written yet.

**Fix:**

Make extraction atomic from observers' point of view: extract into a sibling temp directory, then fs::rename it into place as a single step. Same filesystem → rename is atomic. Observers now see webrtc_dir as either non-existent or fully populated, never mid-extraction.

### DIFF
--- a/.changeset/fix_race_in_download_webrtc_to_reduce_flaky_build.md
+++ b/.changeset/fix_race_in_download_webrtc_to_reduce_flaky_build.md
@@ -1,0 +1,9 @@
+---
+libwebrtc: patch
+livekit: patch
+livekit-ffi: patch
+webrtc-sys: patch
+webrtc-sys-build: patch
+---
+
+fix race in download_webrtc to reduce flaky build - #1047 (@hechen-eng)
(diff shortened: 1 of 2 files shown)


ladvoc (approved):
LGTM ✅

PR #1048, opened 2026-04-27
encoded_video_ingest (sderosa)

stephen-derosa:
# Overview
This PR adds encoded video ingest support, allowing callers to publish pre-compressed video frames through a VIDEO_SOURCE_ENCODED source instead of sending raw frames through WebRTC’s normal encoder path.

It introduces the FFI/protobuf surface for creating encoded video sources and pushing encoded access units, wires those sources into WebRTC using a passthrough encoder, and forwards encoder-side feedback such as keyframe requests and target bitrate changes. For H.264/H.265, the native source also caches parameter sets and prepends them to keyframes when needed.

This solves the need to ingest externally encoded video without decoding and re-encoding it inside the SDK. To reproduce the original limitation, attempt to publish an already-encoded H.264/H.265/VPx/AV1 stream through the existing raw video source APIs; the SDK only accepted raw video frames and would route them through normal encoding.

## Breaking changes
None.

## MSRV
No MSRV changes.

## Testing
Added/updated tests for:
- encoded source behavior
- queue handling
- metadata propagation
- ingest configuration
- examples for Manual validation can be performed with the encoded video ingest example by sending an externally encoded stream into the SDK and verifying it publishes without re-encoding.

Async
No changes to the runtime model were made, it uses the existing livekit video track mechanisms.

# API Exampls
There are two producer APIs: the helper TCP ingest API for common “external encoder over TCP” workflows, and the base encoded source API for applications that already own demuxing, frame boundaries, and encoder control.

Base API: NativeEncodedVideoSource
Use the base API when the application already has complete encoded access units and wants to push them directly.
```
// create source and track
let source = NativeEncodedVideoSource::new(
    VideoCodec::H264,
    VideoResolution { width: 640, height: 480 },
);
source.set_observer(Arc::new(MyObserver));
let track = LocalVideoTrack::create_video_track(
    "encoded-h264",
    RtcVideoSource::Encoded(source.clone()),
);
// publish
room.local_participant()
    .publish_track(LocalTrack::Video(track), TrackPublishOptions::default())
    .await?;
// push each complete encoded picture/access unit
let info = EncodedFrameInfo {
    is_keyframe,
    has_sps_pps: false,
    width: 640,
    height: 480,
    capture_time_us: 0,
};
if !source.capture_frame(&encoded_access_unit, &info) {
    // queue was full; frame was dropped
}
```

Using the EncodedTcpIngest handles is a more complete tool for the user.

### DIFF
--- a/examples/encoded_video_ingest/src/sender.rs
+++ b/examples/encoded_video_ingest/src/sender.rs
@@ -1,0 +1,800 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Encoded H.264 / H.265 / VP8 / VP9 / AV1 ingest sender.
+//!
+//! Connects to a gstreamer pipeline as a TCP client and pushes each
+//! decoded access unit / frame straight through
+//! `NativeEncodedVideoSource::capture_frame`. No software encoding
+//! happens on the Rust side — the bytes on the wire are the bytes that
+//! get packetized into RTP.
+//!
+//! Two framings are supported, picked by `--codec`:
+//!
+//! * **H.264 / H.265**: raw Annex-B bytestream. The sender splits on
+//!   AUD NAL boundaries (NAL type 9 for H.264, type 35 for H.265) and
+//!   delivers each access unit.
+//! * **VP8 / VP9 / AV1**: IVF container (gstreamer's `ivfmux` or
+//!   `avmux_ivf`). The sender parses the 32-byte IVF file header once
+//!   (when present), then each 12-byte frame header + payload, and
+//!   delivers each raw VPx frame (for AV1, each IVF record is one
+//!   Temporal Unit — a complete OBU sequence for one frame).
+//!
+//! TCP is used instead of UDP because macOS caps per-datagram UDP
+//! payloads well below 64 KB by default, which is easy to exceed with
+//! keyframes. The matching gstreamer pipelines are documented in
+//! README.md.
+
+use std::{
+    sync::{
+        atomic::{AtomicBool, AtomicU64, Ordering},
+        Arc, Mutex,
+    },
+    time::{Duration, Instant},
+};
+
+use anyhow::{Context, Result};
+use clap::Parser;
+use libwebrtc::video_source::{EncodedFrameInfo, RtcVideoSource, VideoCodec, VideoResolution};
+use livekit::{
+    options::{TrackPublishOptions, VideoCodec as LkVideoCodec, VideoEncoding},
+    prelude::*,
+    webrtc::video_source::native::{EncodedVideoSourceObserver, NativeEncodedVideoSource},
+};
+use livekit_api::access_token;
+use log::{info, warn};
+use tokio::{io::AsyncReadExt, net::TcpStream, time::sleep};
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// LiveKit server URL (or set LIVEKIT_URL env var)
+    #[arg(long, env = "LIVEKIT_URL")]
+    url: String,
+
+    /// LiveKit API key (or set LIVEKIT_API_KEY env var)
+    #[arg(long, env = "LIVEKIT_API_KEY")]
+    api_key: String,
+
+    /// LiveKit API secret (or set LIVEKIT_API_SECRET env var)
+    #[arg(long, env = "LIVEKIT_API_SECRET")]
+    api_secret: String,
+
+    /// Room name to join
+    #[arg(long, default_value = "encoded-video-demo")]
+    room: String,
+
+    /// Participant identity
+    #[arg(long, default_value = "encoded-sender")]
+    identity: String,
+
+    /// Host of the gstreamer `tcpserversink` producing the Annex-B bytestream
+    #[arg(long, default_value = "127.0.0.1")]
+    tcp_host: String,
+
+    /// Port of the gstreamer `tcpserversink` producing the Annex-B bytestream
+    #[arg(long, default_value_t = 5000)]
+    tcp_port: u16,
+
+    /// Declared stream width (px)
+    #[arg(long, default_value_t = 640)]
+    width: u32,
+
+    /// Declared stream height (px)
+    #[arg(long, default_value_t = 480)]
+    height: u32,
+
+    /// RTP sender max bitrate advertised to WebRTC, in kbps
+    #[arg(long, default_value_t = 2_500)]
+    max_bitrate_kbps: u64,
+
+    /// RTP sender max framerate advertised to WebRTC
+    #[arg(long, default_value_t = 30.0)]
+    max_framerate: f64,
+
+    /// Encoded codec on the wire. Must match the gstreamer pipeline.
+    #[arg(long, value_enum, default_value_t = CodecArg::H264)]
+    codec: CodecArg,
+}
+
+/// Codec selector for the CLI. Drives both framing (Annex-B vs. IVF)
+/// and keyframe detection.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, clap::ValueEnum)]
+enum CodecArg {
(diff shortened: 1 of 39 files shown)


examples/encoded_video_ingest/src/simple_sender.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    env_logger::init();
+    let args = Args::parse();
+
+    let url = args
+        .url
+        .or_else(|| env::var("LIVEKIT_URL").ok())
```
ladvoc: suggestion: You can enable the _env_ feature on the _clap_ crate to get this functionality for free. Also can be applied to the other binaries. Example:
```rust
#[arg(long, env = "LIVEKIT_URL")]
url: Option<String>,
```

livekit-ffi/protocol/encoded_tcp_ingest.proto
theomonnom, line 1: I think TCP should be something done at the user's application layer, not the rust-sdks

stephen-derosa: ty, done

xianshijing-lk:
This project involves API design and changes, and might touch lots of code. 
can we start with filling in a new project template https://www.notion.so/livekit/LiveKit-Client-SDK-New-Product-Development-Process-Template-3353c4901a428072912aef1f44c66b7a and follow some of the instructions there ?

xianshijing-lk:
I would think the team will need to align on the API designs and architecture.

webrtc-sys/src/video_encoder_factory.cpp
```diff
@@ -146,13 +149,28 @@ VideoEncoderFactory::CodecSupport VideoEncoderFactory::QueryCodecSupport(
...
+  if (!format.IsCodecInList(internal_factory_->GetSupportedFormats())) {
+    return nullptr;
   }
 
-  return encoder;
+#ifdef LK_PRE_ENCODED_VIDEO
+  // Wrap the real encoder construction in a lazy shim so we can branch
+  // between passthrough and a real encoder based on the first VideoFrame's
+  // id. The builder is called at most once and only for non-passthrough
+  // tracks; passthrough tracks never instantiate the SimulcastEncoderAdapter.
```
theomonnom, line 414: If it's determined at the first VideoFrame, how does libwebrtc knows which codec to offer in the sdp?

webrtc-sys/src/encoded_video_source.cpp
```diff
@@ -0,0 +1,386 @@
...
+  auto dummy_buffer = webrtc::I420Buffer::Create(2, 2);
+  webrtc::I420Buffer::SetBlack(dummy_buffer.get());
+
+  webrtc::VideoFrame frame =
+      webrtc::VideoFrame::Builder()
+          .set_video_frame_buffer(dummy_buffer)
+          .set_rotation(webrtc::kVideoRotation_0)
+          .set_timestamp_us(capture_time_us != 0 ? capture_time_us
+                                                 : webrtc::TimeMicros())
+          .set_id(source_id_)
```
theomonnom, line 307: Is there a less hacky way to detect pre-encoded sources?

theomonnom (commented):
Otherwise lg!

webrtc-sys/src/video_encoder_factory.cpp
stephen-derosa: Ah, good callout. I think this comment is a bit misleading. My understanding is that SDP codec selection happens earlier, from the encoder factory’s advertised capabilities plus the codec preferences we set on the transceiver. For encoded video, the Rust publish path already knows the source codec, overrides the publish option to match it, and then set_codec_preferences narrows the offer to that codec.

The first-frame check here is only deciding which encoder implementation to use for that already-negotiated codec: passthrough for an encoded source, or the normal encoder path for raw frames. The VideoFrame::id() is just how we identify that the frame came from an encoded source.

webrtc-sys/src/encoded_video_source.cpp
stephen-derosa: hm, what do you suggest?

webrtc-sys/include/livekit/encoded_video_source.h
```diff
@@ -0,0 +1,168 @@
...
+ public:
+  class InternalSource : public webrtc::AdaptedVideoTrackSource {
+   public:
+    InternalSource(uint16_t source_id,
+                   EncodedVideoCodecType codec,
+                   uint32_t width,
+                   uint32_t height);
+    ~InternalSource() override;
+
+    bool is_screencast() const override { return false; }
```
ladvoc: question: We expose this for `NativeVideoSource`; should we also expose for `NativeEncodedVideoSource`?

livekit/src/room/participant/local_participant.rs
```diff
@@ -302,6 +302,36 @@ impl LocalParticipant {
         track: LocalTrack,
         options: TrackPublishOptions,
     ) -> RoomResult<LocalTrackPublication> {
+        // Encoded video sources deliver encoded single-layer frames.
+        // Force-disable simulcast and pin the negotiated codec to the
+        // source's codec so WebRTC's encoder factory picks our passthrough
+        // encoder path.
+        #[cfg(all(not(target_arch = "wasm32"), feature = "encoded-video"))]
+        let options = {
```
ladvoc: nitpick: This could be broken out into a private helper (e.g., `publish_options_for_track`).

livekit/Cargo.toml
```diff
@@ -13,6 +13,7 @@ default = ["tokio"]
 async = ["livekit-api/signal-client-async"]
 tokio = ["livekit-api/signal-client-tokio"]
 dispatcher = ["livekit-api/signal-client-dispatcher"]
+encoded-video = ["libwebrtc/encoded-video"]
```
ladvoc: suggestion: Originally this made some sense since this was going to be used to conditionally pull in the additional tokio features below. However, now that that part is going to be stripped out of the core crate, there is no longer an advantage to putting this behind a feature; impact on binary size is negligible and this won't get built and tested in CI since it isn't default.

libwebrtc/src/video_source.rs
```diff
@@ -29,15 +29,60 @@ impl Default for VideoResolution {
...
+/// [`native::NativeEncodedVideoSource`].
+#[cfg(feature = "encoded-video")]
+#[derive(Debug, Copy, Clone)]
+pub struct EncodedFrameInfo {
+    /// True when this frame is an IDR / keyframe.
+    pub is_keyframe: bool,
+    /// True when the `data` buffer already has SPS/PPS (or equivalent)
+    /// prepended. H.264/H.265 only; ignored for other codecs.
+    pub has_sps_pps: bool,
+    pub width: u32,
```
ladvoc: suggestion: Reuse `VideoResolution` here instead of discrete width/heigh fields.

examples/encoded_video_ingest/src/sender.rs
```diff
@@ -0,0 +1,800 @@
...
+            for au in out.drain(..) {
+                encoded_bytes.fetch_add(au.len() as u64, Ordering::Relaxed);
+                let is_keyframe = is_keyframe(args.codec, &au);
+                if is_keyframe {
+                    keyframes.fetch_add(1, Ordering::Relaxed);
+                }
+                let info = EncodedFrameInfo {
+                    is_keyframe,
+                    has_sps_pps: false, // the source scans+prepends SPS/PPS as needed
+                    width: args.width,
```
ladvoc: question: What happens if width or height are zero?

ladvoc (commented):
Looking good so far!

livekit/Cargo.toml
stephen-derosa: ty, agreed, done

webrtc-sys/include/livekit/encoded_video_source.h
stephen-derosa: yup i think we do, this is missing plumbing. ty, updated

livekit/src/room/participant/local_participant.rs
stephen-derosa: nice, done

libwebrtc/src/video_source.rs
stephen-derosa: nice, done

examples/encoded_video_ingest/src/sender.rs
stephen-derosa: If width or height are zero, they’re treated as “no resolution update” in the capture path.

In the normal example path, resolution is created from CLI args and used both to create the source and in EncodedFrameInfo. If either dimension is zero there, the source is also initialized with a zero dimension, so frames may still be accepted, but the queued encoded frames can carry 0 dimensions downstream. This is the same behavior as `NativeVideoSource`

stephen-derosa:
closing since this work has been moved to livekit-capture crate

PR #1051, opened 2026-04-29
feat(connector-service): update for new protos

anunaym14:
also update protocol to latest release

### DIFF
--- a/livekit-api/src/services/connector.rs
+++ b/livekit-api/src/services/connector.rs
@@ -13,12 +13,17 @@
 // limitations under the License.
 
 use livekit_protocol as proto;
-use std::collections::HashMap;
+use pbjson_types::Duration as ProtoDuration;
+use std::{collections::HashMap, time::Duration};
 
 use super::{ServiceBase, ServiceResult, LIVEKIT_PACKAGE};
 use crate::{access_token::VideoGrants, get_env_keys, services::twirp_client::TwirpClient};
 
 const SVC: &str = "Connector";
+
+fn duration_to_proto(d: Option<Duration>) -> Option<ProtoDuration> {
+    d.map(|d| ProtoDuration { seconds: d.as_secs() as i64, nanos: d.subsec_nanos() as i32 })
+}
 
 /// Options for dialing a WhatsApp call
 #[derive(Default, Clone, Debug)]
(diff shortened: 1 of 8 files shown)


livekit-api/src/services/connector.rs
```diff
@@ -149,14 +160,17 @@ impl ConnectorClient {
...
+    /// * `api_key` - The API key of the business disconnecting the call. Required when
+    ///   `disconnect_reason` is `BusinessInitiated`; pass an empty string for `UserInitiated`.
+    /// * `disconnect_reason` - The reason for disconnecting the call
     ///
     /// # Returns
     /// Empty response on success
     pub async fn disconnect_whatsapp_call(
         &self,
         call_id: impl Into<String>,
         api_key: impl Into<String>,
```
anunaym14, line 167: slight breaking change, should be okay for now because connector is still in beta, but not sure how stuff like this should be handled long term

livekit-api/src/services/connector.rs
```diff
@@ -13,13 +13,18 @@
...
+use std::{collections::HashMap, time::Duration};
 
 use super::{ServiceBase, ServiceResult, LIVEKIT_PACKAGE};
 use crate::{access_token::VideoGrants, get_env_keys, services::twirp_client::TwirpClient};
 
 const SVC: &str = "Connector";
 
+fn duration_to_proto(d: Option<Duration>) -> Option<ProtoDuration> {
+    d.map(|d| ProtoDuration { seconds: d.as_secs() as i64, nanos: d.subsec_nanos() as i32 })
+}
```
anunaym14: copied over from sip, can probably reduce redundancy

anunaym14: done

livekit-uniffi/src/access_token.rs
```diff
@@ -69,6 +69,8 @@ pub struct SIPGrants {
 pub struct RoomAgentDispatch {
     pub agent_name: String,
     pub metadata: String,
+    /// cloud only - matches `JobRestartPolicy` (0 = on failure, 1 = never)
+    pub restart_policy: i32,
```
anunaym14, line 73: should we declare a new enum here instead?

livekit-api/src/services/connector.rs
```diff
@@ -60,6 +62,10 @@ pub struct AcceptWhatsAppCallOptions {
     pub participant_attributes: Option<HashMap<String, String>>,
     /// Optional - Country where the call terminates as ISO 3166-1 alpha-2
     pub destination_country: Option<String>,
+    /// Optional - Max time for the callee to answer the call
+    pub ringing_timeout: Option<Duration>,
+    /// Optional - Wait for the answer for the call before returning
+    pub wait_until_answered: bool,
```
lukasIO, line 68: shouldn't this be an `Option<bool>` ?

livekit-api/src/services/connector.rs
lukasIO: if you think there might potentially be more arguments added in the future here it might be worth to replace the reason with a `non_exhaustive` options `struct` that allows extending later on

PR #1052, opened 2026-04-29
fix: derive `simulcasted` from non-deprecated TrackInfo fields

cloudwebrtc:
Fixed a bug where the `is_simulcasted` in the `set_video_quality` directive failed, causing the video layer to be unselectable. This bug was found in the `test_change_video_quality` test of the Python SDK.


lukasIO (approved):

xianshijing-lk (approved):

xianshijing-lk:
landed it.

PR #1053, opened 2026-04-30
Improve WebRTC build scripts and add external_audio_source patch

xianshijing-lk:
- Add external_audio_source.patch for NativeAudioSource support
  - Remove unused use_clang_modules build arg (fixes license generation)
  - Use vpython3 for license generation (Python version compatibility)
  - Make license generation failures non-critical
  
 ---
  Before you submit your PR

  - I have read the https://github.com/livekit/rust-sdks/blob/main/CONTRIBUTING.md and validated that this PR will be accepted.
  - I have read and followed the principles regarding breaking changes, testing, and code quality.

  PR description

  This PR improves the WebRTC build scripts and adds a patch required for external audio source support (NativeAudioSource).

  Changes:

  1. Add external_audio_source.patch - This patch modifies WebRTC to support external audio sources that bypass the Audio Device Module (ADM). It adds an is_external_source() method to both AudioSourceInterface and AudioSource classes, and modifies AudioSendStream to not register with AudioState when using external sources. This prevents mixing of device-captured audio with externally-sourced audio (e.g., from NativeAudioSource).
  2. Remove unused use_clang_modules=false build argument - This argument is no longer recognized by WebRTC m144 and causes GN to output warnings. These warnings break the license generation script because gn desc --format=json outputs warnings before the JSON, causing json.loads() to fail.
  3. Use vpython3 instead of python3 for license generation - vpython3 from depot_tools uses a controlled Python environment (Python 3.11) that is compatible with WebRTC's tooling, rather than relying on the system Python which may be incompatible (e.g., Python 3.14).
  4. Make license generation failures non-critical - Added fallback error handling so that license generation failures don't break the build. The libwebrtc.a static library is the critical output; LICENSE.md is nice-to-have.

  Breaking changes

  None. These are build script improvements only and do not affect the Rust API.

  MSRV

  No changes to MSRV.

  Testing

  The changes were tested by running:
  ./build_macos.sh --arch arm64 --profile release

  The build completes successfully and LICENSE.md is now generated without errors.

  The external_audio_source.patch was verified to apply cleanly to a fresh WebRTC m144_release checkout:
  cd webrtc-sys/libwebrtc/src
  git checkout -- .
  git apply ../patches/external_audio_source.patch -v

  Async

  N/A - This PR only modifies shell build scripts, no Rust code changes.

### DIFF
--- a/webrtc-sys/libwebrtc/build_android.sh
+++ b/webrtc-sys/libwebrtc/build_android.sh
@@ -71,6 +71,7 @@ cd src
 git apply "$COMMAND_DIR/patches/ssl_verify_callback_with_native_handle.patch" -v --ignore-space-change --ignore-whitespace --whitespace=nowarn
 git apply "$COMMAND_DIR/patches/add_deps.patch" -v --ignore-space-change --ignore-whitespace --whitespace=nowarn
 git apply "$COMMAND_DIR/patches/android_use_libunwind.patch" -v --ignore-space-change --ignore-whitespace --whitespace=nowarn
+git apply "$COMMAND_DIR/patches/external_audio_source.patch" -v --ignore-space-change --ignore-whitespace --whitespace=nowarn
 # livekit prefixed jni
 git apply "$COMMAND_DIR/patches/jni_prefix.patch" -v --ignore-space-change --ignore-whitespace --whitespace=nowarn
 
@@ -105,7 +106,6 @@ args="is_debug=$debug \
   enable_iterator_debugging=false \
   android_package_prefix=\"livekit\" \
   use_custom_libcxx=false \
-  use_clang_modules=false \
   use_rtti=true"
 
 if [ "$debug" = "true" ]; then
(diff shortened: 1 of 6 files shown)


webrtc-sys/libwebrtc/build_android.sh
```diff
@@ -105,7 +106,6 @@ args="is_debug=$debug \
   enable_iterator_debugging=false \
   android_package_prefix=\"livekit\" \
   use_custom_libcxx=false \
-  use_clang_modules=false \
```
xianshijing-lk, line 108: use_clang_modules=false seem to make the licence generate script fail, @cloudwebrtc , any concern of removing this flag ?

webrtc-sys/libwebrtc/patches/external_audio_source.patch
```diff
@@ -0,0 +1,99 @@
...
++++ b/media/engine/webrtc_voice_engine.cc
+@@ -1017,6 +1017,14 @@ class WebRtcVoiceSendChannel::WebRtcAudioSendStream : public AudioSource::Sink {
+       RTC_DCHECK(source_ == source);
+       return;
+     }
++
++    // Check if this is an external audio source (delivers audio via AddSink).
++    // If so, mark the config so AudioState doesn't send device audio to this
++    // stream. This must be done before UpdateSendState() calls Start().
++    if (source->is_external_source() && !config_.external_source) {
```
cloudwebrtc, line 93: QQ: This method (`AudioSource::is_external_source() const`) is looks like not overridden in any inherited class.

In another Rust PR, [is_external_source ](https://github.com/livekit/rust-sdks/pull/1037/changes#diff-bd8e1e3f4ae86c785f61891f4e7d57f0ece039edf22a21569a6180c1361e2428R131)is overridden to `webrtc::AudioSourceInterface`, but it doesn't seem to be associated with `webrtc::AudioSource`. I suspect that `source->is_external_source()` here will always return `false`.

Perhaps we're missing a copy of the `is_external_source` state from `webrtc::AudioSourceInterface` to `webrtc::AudioSrouce`?  The class associated with `webrtc::AudioSource` is [LocalAudioSinkAdapter](https://github.com/webrtc-sdk/webrtc/blob/m144_release/pc/rtp_sender.h#L304-L305).

webrtc-sys/libwebrtc/build_android.sh
cloudwebrtc: Yeah, `use_clang_modules=false` was introduced in the m144 upgrade; I'll double-check if it's a required parameter.

webrtc-sys/libwebrtc/patches/external_audio_source.patch
xianshijing-lk: Good point, I think the problem with the current approach is that it will send both the external source file and mic source to one track. 
which will result in duplicating the mic data with some drift there.

I try patching it, can you check it again ? 

It might worth coming up with some integration tests, let me work on it

cloudwebrtc (approved):
lgtm!

PR #1056, opened 2026-05-04
Add AGENTS.md

ladvoc:
Resolves CLT-2486

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,40 @@
+# AGENTS.md
+
+## API changes
+
+- Breaking public API changes are to be avoided unless necessary to complete your task
+  - Be explicit when you plan to make breaking API changes
+- When introducing new API surface, always default to private or `pub(crate)` unless there is a specific reason to expose publicly
+- Introduce new public APIs sparingly
+- New public types that will be extensively used should be added to the prelude
+
+## Dependencies
+
+- When using APIs from third-party crates, never assume you already know the API
+  - **ALWAYS** reference [_docs.rs_](https://docs.rs) for docs of the specific version being used
+- Pull in new dependencies as a last resort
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,40 @@
+# AGENTS.md
+
+## API changes
+
+- Breaking public API changes are to be avoided unless necessary to complete your task
+  - Be explicit when you plan to make breaking API changes
+- When introducing new API surface, always default to private or `pub(crate)` unless there is a specific reason to expose publicly
+- Introduce new public APIs sparingly
+- New public types that will be extensively used should be added to the prelude
+
```
xianshijing-lk, line 16: [question](https://pullpo.io/cc?l=question) 

Is it worth adding a section for Ownership & lifetimes ? 
I think your mention of avoiding singleton was a good callout when reviewing my PR.

AGENTS.md
```diff
@@ -0,0 +1,40 @@
+# AGENTS.md
+
+## API changes
+
+- Breaking public API changes are to be avoided unless necessary to complete your task
+  - Be explicit when you plan to make breaking API changes
+- When introducing new API surface, always default to private or `pub(crate)` unless there is a specific reason to expose publicly
+- Introduce new public APIs sparingly
+- New public types that will be extensively used should be added to the prelude
```
xianshijing-lk: regarding prelude, will this lead to prelude bloat and namespace pollution  ? 
do we want to tighten the guideline regarding prelude ?

xianshijing-lk (approved):
two comments, lgtm iff you address them

AGENTS.md
ladvoc: I think so, added.

AGENTS.md
ladvoc: I've removed this line for now.

PR #1061, opened 2026-05-06
webrtc-sys-build: disable zip default features

charliermarsh:
## Summary

My understanding is that `webrtc-sys-build` needs the `zip` crate to extract the prebuilt WebRTC archives, which always use `deflate`. Trimming the default features removes a significant portion of the dependency tree.

### DIFF
--- a/.changeset/disable_default_features_for_zip_dependency.md
+++ b/.changeset/disable_default_features_for_zip_dependency.md
@@ -1,0 +1,6 @@
+---
+webrtc-sys-build: patch
+---
+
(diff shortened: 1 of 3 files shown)


ladvoc:
@charliermarsh, thank you for your contribution. Please add the following changeset at `.changeset/disable_default_features_for_zip_dependency.md`:
```yml
---
webrtc-sys-build: patch
---

# Disable default features for zip dependency

```

charliermarsh:
Done, thank you.

ladvoc (approved):

PR #1065, opened 2026-05-08
Fix LocalTrackPublished handle leak

MaxHeimbrock:
## Summary

Found while counting handles during Unity sessions via https://github.com/livekit/rust-sdks/pull/1079

### Background

The `forward_event` handler for `RoomEvent::LocalTrackPublished` was allocating a fresh `FfiPublication`, calling `FfiServer::store_handle`, and then sending the proto event without the handle id. Because `proto::LocalTrackPublished` only carries `track_sid`, the foreign-language client never received the id and had no way to call `drop_handle` on it. Each local `publish_track` call therefore leaked one publication entry in the FFI handle table for the lifetime of the process.

The publication is already wrapped and handed to the client via the `PublishTrackCallback` path. The `LocalTrackPublished` event is a notification that fires after that callback to let the client run any post-publish bookkeeping; it does not need its own handle. Drop the redundant allocation entirely.

Surfaced while investigating untracked memory growth in the Unity SDK, where the FFI handle table held two publication entries per local track across a session and only one of them could be dropped on disconnect.

### Test plan
- [x] Verified with an FFI handle table diagnostic that local publish + disconnect drops to zero residual publication handles.


MaxHeimbrock:
Not sure if still needed due to https://github.com/livekit/rust-sdks/pull/1072/changes#diff-6dde26c71e292a49e4f1b870a674ca134b8a083136ccecb0de4f18a498e2e8b9 also making changes, storing the handle by:

```
server.store_handle(handle_id, ffi_publication);
```

stephen-derosa (approved):
lgtm -- any tests we can put in place to validate this fix?

xianshijing-lk (approved):

xianshijing-lk:
@alan-george-lk , can we build some tests on C++ and enable the LeakSanitizer with -fsanitize=address on clang / GCC ?

MaxHeimbrock:
> lgtm -- any tests we can put in place to validate this fix?

To my knowledge we need an actual client using the FFI and a local server. And in the Rust SDK we don't have CI setup to test E2E, that is done in https://github.com/livekit/e2e.

I have tested this manually again today and will merge, but I want to also catch up on how to make testing easier.

PR #1067, opened 2026-05-08
local video example enhancements

chenosaurus:
* add `--test-pattern` flag to publisher example to generate SMPTE color bars.
* improve latency metrics overlay
* add `--display-video` on publisher example to render video being published.

### DIFF
--- a/examples/local_video/README.md
+++ b/examples/local_video/README.md
@@ -68,6 +68,13 @@ Publisher usage:
    --room-name demo \
    --identity cam-1 \
    --e2ee-key my-secret-key
+
+ # publish and display the outgoing video locally
+ cargo run -p local_video -F desktop --bin publisher -- \
+   --camera-index 0 \
+   --room-name demo \
+   --identity cam-1 \
+   --display-video
 ```
 
 List devices usage:
@@ -87,6 +94,7 @@ Publisher flags (in addition to the common connection flags above):
 - `--attach-timestamp`: Attach the current wall-clock time (microseconds since UNIX epoch) as the user timestamp on each published frame. The subscriber can display this to measure end-to-end latency.
 - `--burn-timestamp`: Burn the attached timestamp into the video frame as a visible overlay. Has no effect unless `--attach-timestamp` is also set.
 - `--attach-frame-id`: Attach a monotonically increasing frame ID to each published frame via the packet trailer. The subscriber displays this in the timestamp overlay when `--display-timestamp` is used.
+- `--display-video`: Open a window that displays the video frames being published.
 - `--e2ee-key <key>`: Enable end-to-end encryption with the given shared key. The subscriber must use the same key to decrypt.
 
 Subscriber usage:
--- a/examples/local_video/README.md
+++ b/examples/local_video/README.md
@@ -56,6 +56,12 @@ Publisher usage:
    --room-name demo \
    --identity cam-1
 
+ # publish a static SMPTE color-bar test pattern (no camera required)
+ cargo run -p local_video -F desktop --bin publisher -- \
+   --test-pattern \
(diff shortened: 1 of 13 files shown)


ladvoc:
_local_video_ should not be in the changeset since it is not a versioned package configured in _knope.toml_.

examples/local_video/src/publisher.rs
```diff
@@ -228,6 +309,30 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
...
+            Err(err) if is_twirp_not_found(&err) => {
+                debug!("Room '{}' did not exist before recreation", args.room_name);
+            }
+            Err(err) => return Err(err.into()),
+        }
+        room_client
+            .create_room_with_playout_delay(
+                &args.room_name,
+                CreateRoomOptions::default(),
+                args.min_playout_delay.unwrap_or_default(),
```
stephen-derosa, line 330: in playout delay, are delay packets buffered, or is this basically "dont use data until timeout"

alan-george-lk (approved):
LGTM: my review is more of a rubber stamp since there's a lot of context I'm missing, but was a good Rust exercise for me

examples/local_video/README.md
```diff
@@ -71,6 +85,7 @@ List devices usage:
 
 Publisher flags (in addition to the common connection flags above):
 - `--camera-index <n>`: Camera index to use (default: `0`). Use `--list-cameras` to see available indices.
+- `--test-pattern`: Generate a standard SMPTE 75% color-bar test pattern instead of capturing from a camera. `--camera-index` is ignored when this is set; `--width`, `--height`, and `--fps` still control the output resolution and frame rate.
```
stephen-derosa, line 88: nice!

stephen-derosa (commented):
Nice, i appreciate the new additions, they will definitely come in handy!

examples/local_video/src/publisher.rs
```diff
@@ -228,6 +309,30 @@ async fn run(args: Args, ctrl_c_received: Arc<AtomicBool>) -> Result<()> {
         .or_else(|| env::var("LIVEKIT_API_SECRET").ok())
         .expect("LIVEKIT_API_SECRET must be provided via --api-secret or env");
 
+    if args.min_playout_delay.is_some() || args.max_playout_delay.is_some() {
```
ladvoc, line 312: suggestion: This can be applied from the token (no need for server API requests); see [`with_room_config`](https://docs.rs/livekit-api/0.4.21/livekit_api/access_token/struct.AccessToken.html#method.with_room_config).

examples/local_video/src/publisher.rs
chenosaurus: this is a hint to the subscriber to set target jitter buffer duration.  By setting this to a low value, it will cause the subscriber to worry less about buffering frames for a smooth playback and try to render the latest frame.

examples/local_video/src/publisher.rs
chenosaurus: ooh I didn't know that.  Do you know if this must be set in the token that triggers room creation?  In the publisher example, I explicitly destroy the room then recreate it w/ the options to ensure they are set since it's a room level config.

ladvoc: I can't seem to find that documented anywhere. For other properties that can be set on the token (such as agent dispatches), it doesn't have to be the token that creates the room. However, this might be different for something like playout delay (i.e., might not get updated after the room is created).

chenosaurus: I opted for creating the room via server API before the publisher joins to ensure it has the right room config.

PR #1068, opened 2026-05-08
FFI room events: only sent after receiving ReadyForRoomEventRequest after initial connection

ladvoc:
Resolves BOT-344


livekit-ffi/protocol/ffi.proto
```diff
@@ -15,18 +15,19 @@
 syntax = "proto2";
 
 package livekit.proto;
-option csharp_namespace = "LiveKit.Proto";
 
+import "audio_frame.proto";
```
stephen-derosa: im not sure why these got re-ordered, i dont have a formatter on for protos...

livekit-ffi/protocol/ffi.proto
```diff
@@ -167,7 +168,10 @@ message FfiRequest {
     // Reconnection / chaos testing
     SimulateScenarioRequest simulate_scenario = 76;
 
-    // NEXT_ID: 77
+    // Room event flush handshake
+    FlushEventsRequest flush_events = 77;
```
ladvoc: suggestion (non-blocking): Maybe we should prefix these message names with `Room` so it is clear this only applies to room events.

livekit-ffi/protocol/room.proto
```diff
@@ -136,7 +144,7 @@ message PublishDataRequest {
   required uint64 data_ptr = 2;
   required uint64 data_len = 3;
   required bool reliable = 4;
-  repeated string destination_sids = 5 [deprecated=true];
```
ladvoc, line 139: nitpick: There are some unrelated formatting changes in this file.

livekit-ffi/src/server/requests.rs
```diff
@@ -114,6 +114,18 @@ fn on_simulate_scenario(
     Ok(proto::SimulateScenarioResponse { async_id })
 }
 
+/// Release the connect task's wait point so room event forwarding can begin.
+/// The FFI client should send this once it has installed its event listener
+/// and is ready to receive room events.
+fn on_flush_events(
```
stephen-derosa: @ladvoc any thoughts on renaming flush to `ReadyForRoomEventsRequest` or `ParticipantReadySignal` ?

livekit-ffi/src/server/room.rs
```diff
@@ -225,7 +243,27 @@ impl FfiRoom {
...
+                    // send a FlushEventsRequest before forwarding any room
+                    // events. This avoids a race where events emitted between
+                    // the ConnectCallback and the listener registration are
+                    // dropped.
+                    if wait_for_room_event_flush(&ffi_room.flush_notify).await.is_err() {
+                        let msg = format!(
+                            "timed out waiting for FlushEventsRequest after ConnectCallback \
+                             (room_handle={handle_id})"
+                        );
+                        log::error!("{msg}");
```
ladvoc: nitpick: Log macros support format specifiers just like format does.

livekit-ffi/src/server/room.rs
```diff
@@ -1596,3 +1641,23 @@ fn build_initial_states(
         remote_infos,
     )
 }
+
+#[cfg(test)]
```
stephen-derosa: I also tested this locally by adding a 2 second sleep in the cpp sdk before sending the flush request and seeing the error log `Timed out waiting for FlushEventsRequest after ConnectCallback`. This test adds new deps to tokio:
```
["macros", "rt", "test-util", "time"]
```

is the test worth the deps?

livekit-ffi/src/server/room.rs
```diff
@@ -225,7 +243,27 @@ impl FfiRoom {
                         .into(),
                     );
 
-                    // Update Room SID on promise resolve
+                    // Wait for the FFI client to install its event listener and
+                    // send a FlushEventsRequest before forwarding any room
+                    // events. This avoids a race where events emitted between
+                    // the ConnectCallback and the listener registration are
+                    // dropped.
+                    if wait_for_room_event_flush(&ffi_room.flush_notify).await.is_err() {
```
ladvoc: suggestion: I am open to just inlining the definition of `wait_for_room_event_flush` here and doing away with the unit tests—they  are essentially just testing `tokio::time::timeout`.

livekit-ffi/Cargo.toml
```diff
@@ -50,6 +50,7 @@ webrtc-sys-build = { workspace = true }
 
 [dev-dependencies]
 livekit-api = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt", "test-util", "time"] }
```
ladvoc: This is just a dev dependency so I wouldn't be concerned about the additional features. However, if you decide to remove the test (see other comment), this should be removed too.

ladvoc (commented):
LGTM. Can't approve since it's my PR, but here's a few minor comments.

livekit-ffi/src/server/requests.rs
ladvoc: Good callout, I didn't see this comment until now. Yes, I think we should have room in the title just to be explicit that this only applies to room events.

stephen-derosa: going with `ReadyForRoomEventsRequest`

livekit-ffi/protocol/room.proto
stephen-derosa: updated

livekit-ffi/src/server/requests.rs
stephen-derosa: done

1egoman (approved):
Makes sense to me.

question: Is it possible to add some sort of test case for this? I know that `livekit-ffi` historically hasn't had the best test coverage but if feasible without crazy amounts of mocking I think it would be a fairly straightforward test case.

stephen-derosa:
> Makes sense to me.
> 
> question: Is it possible to add some sort of test case for this? I know that `livekit-ffi` historically hasn't had the best test coverage but if feasible without crazy amounts of mocking I think it would be a fairly straightforward test case.

@1egoman so i did have a test in here but it was ultimately testing the tokio timeout functionality. I have however run this from the cpp sleep with a hard coded sleep and saw that the error does in fact get thrown

cloudwebrtc (approved):

PR #1069, opened 2026-05-09
support SimulateScenario through FFI to improve testing

davidzhao:
also fixed a few edge cases with connection resumption. added tests


ladvoc (approved):

livekit/src/room/mod.rs
```diff
@@ -1498,8 +1498,10 @@ impl RoomSession {
     }
 
     fn handle_resumed(self: &Arc<Self>, tx: oneshot::Sender<()>) {
+        // Resume preserves the PeerConnection and existing local publications,
+        // so we intentionally do NOT dispatch RoomEvent::Reconnected here.
+        // Applications observe the recovery via ConnectionStateChanged.
         self.update_connection_state(ConnectionState::Connected);
-        self.dispatcher.dispatch(&RoomEvent::Reconnected);
```
ladvoc, line 1502: I think the removal of this line is causing the E2E signaling tests to fail.

ladvoc (approved):

livekit/src/room/mod.rs
davidzhao: looking at JS, we actually do emit this event.. so I'm going to revert this.. 

also found a few other discrepancies in behavior.. will address them too

theomonnom (approved):

livekit/src/room/mod.rs
theomonnom: oh interesting, so we've always been firing the Reconnected event even for resume?

davidzhao: yeah looks that way.. though the behavior is the same where these are notifications. the client still takes care of all of the republishing tasks internally.

livekit-api/src/signal_client/mod.rs
xianshijing-lk, line 618: This is not an issue with this PR, but simply a question.

When a ping timeout happened, it emits SignalEvent::Close, exits the loop, then closes/removes the stream.

Is this an expected or documented behavior ? or should we try reconnecting instead ? 

@davidzhao @ladvoc

livekit-ffi/src/server/requests.rs
```diff
@@ -82,6 +82,39 @@ fn on_disconnect(
...
+        proto::SimulateScenarioKind::SimulateServerLeave => SimulateScenario::ServerLeave,
+        proto::SimulateScenarioKind::SimulateMigration => SimulateScenario::Migration,
+        proto::SimulateScenarioKind::SimulateForceTcp => SimulateScenario::ForceTcp,
+        proto::SimulateScenarioKind::SimulateForceTls => SimulateScenario::ForceTls,
+        proto::SimulateScenarioKind::SimulateFullReconnect => SimulateScenario::FullReconnect,
+    };
+
+    let handle = server.async_runtime.spawn(async move {
+        let ffi_room =
+            server.retrieve_handle::<room::FfiRoom>(request.room_handle).unwrap().clone();
```
xianshijing-lk: nit, there is chance that the unwrap() in async task can panic if handle is disposed concurrently ? 

maybe it is worth being a bit more careful :  
```
let ffi_room = match server.retrieve_handle::<room::FfiRoom>(disconnect.room_handle) {
    Ok(room) => room.clone(),
    Err(err) => {
        let _ = server.send_event(proto::DisconnectCallback { async_id }.into());
        return;
    }
};
```

xianshijing-lk (approved):
great work. it looks like a solid improvement, thanks for fixing the problems there. 

one question and one nit.

livekit-api/src/signal_client/mod.rs
davidzhao: yup, this will exit the loop and the engine layer would trigger a resume.

PR #1072, opened 2026-05-10
chore: add LocalTrackRepublished event for FFI clients

davidzhao:
during a reconnect, all published tracks will be republished and receive new TrackID. there is currently no way for FFI to learn about the new TrackID.


cloudwebrtc (approved):
lgtm!

PR #1078, opened 2026-05-11
UniFFI: Publish Swift packages

pblazej:
Leverages https://github.com/livekit/publish-xcframework-action that seems to be working https://github.com/pblazej/test-xcframework-hosting/releases

### DIFF
--- a/.github/workflows/uniffi-packages.yml
+++ b/.github/workflows/uniffi-packages.yml
@@ -1,0 +1,145 @@
+name: UniFFI packages
+
+# Builds per-language wrapper packages produced by `cargo make <lang>-package`
+# (Swift xcframework today; Kotlin/Python can be added as matrix entries) and
+# publishes them to their respective hosting repos.
+#
+# Trigger pattern mirrors ffi-builds.yml: every push to main checks whether
+# knope has produced a draft release for the livekit-uniffi package; if so the
+# matching package job runs. Manual runs go through workflow_dispatch.
+#
+# NOTE: livekit-uniffi must be added to knope.toml with `assets = "marker"`
+# for the auto-trigger path to work. Until then, use workflow_dispatch.
+
+on:
+  push:
+    branches: ["main"]
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: "Release tag (e.g., livekit-uniffi/v0.0.7). Required if no draft release exists."
+        required: false
+        type: string
+      dry_run:
+        description: "Build artifacts but skip release upload/PR."
+        type: boolean
+        default: false
+
+env:
+  CARGO_TERM_COLOR: always
+  TAG_PREFIX: livekit-uniffi/v
+
+jobs:
+  resolve-tag:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    outputs:
+      tag_name: ${{ steps.get-tag.outputs.tag_name }}
+      version: ${{ steps.get-tag.outputs.version }}
+    steps:
+      - uses: actions/checkout@v6
+      - id: get-tag
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          MANUAL_TAG="${{ inputs.tag_name }}"
+          if [ -n "$MANUAL_TAG" ]; then
+            TAG="$MANUAL_TAG"
+            echo "Using manually provided tag: '$TAG'"
+          else
+            TAG=$(gh release list --repo "${{ github.repository }}" \
+              --json 'isDraft,tagName' \
+              --jq "[.[] | select(.isDraft and (.tagName | startswith(\"${TAG_PREFIX}\")))] | first | .tagName // empty")
+            echo "Resolved livekit-uniffi draft release tag: '${TAG:-<none>}'"
+          fi
+          echo "tag_name=${TAG}" >> "$GITHUB_OUTPUT"
+          # Strip the "livekit-uniffi/v" prefix to get the hosting-repo tag (e.g., 0.0.7)
+          VERSION="${TAG#${TAG_PREFIX}}"
+          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
+
+  build:
+    name: Build ${{ matrix.package }} package
+    needs: resolve-tag
+    if: needs.resolve-tag.outputs.tag_name != ''
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - package: swift
+            runner: macos-26
+            hosting-repo: livekit/livekit-uniffi-xcframework
+            # NOTE: spm-name must match SPM_NAME in livekit-uniffi/Makefile.toml.
+            spm-name: LiveKitUniFFI
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: write
+    env:
+      OUTPUT_DIR: livekit-uniffi/packages/${{ matrix.package }}/${{ matrix.spm-name }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - uses: maxim-lobanov/setup-xcode@v1
+        if: matrix.package == 'swift'
+        with:
+          xcode-version: latest-stable
+
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Install Apple Rust targets
+        if: matrix.package == 'swift'
+        run: |
+          rustup target add \
+            aarch64-apple-ios \
+            aarch64-apple-ios-sim \
+            x86_64-apple-ios \
+            aarch64-apple-ios-macabi \
+            x86_64-apple-ios-macabi \
+            aarch64-apple-darwin \
+            x86_64-apple-darwin \
+            aarch64-apple-tvos \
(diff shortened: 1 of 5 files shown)


.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+    permissions:
+      contents: write
+    env:
+      OUTPUT_DIR: livekit-uniffi/packages/${{ matrix.package }}/${{ matrix.spm-name }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - uses: maxim-lobanov/setup-xcode@v1
```
1egoman: issue: Just FYI, before merging this, you should be sure to pin all third party action uses to specific hashes to prevent some flavors of supply chain attacks.

.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+
+on:
+  push:
+    branches: ["main"]
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: "Release tag (e.g., livekit-uniffi/v0.0.7). Required if no draft release exists."
+        required: false
+        type: string
```
1egoman, line 20: question: I'm a bit confused why this is set up to work with draft releases - would it be better to use a similar kick off mechanism as the [release.yml](https://github.com/livekit/rust-sdks/blob/main/.github/workflows/release.yml) and tie into knope PR merges somehow?

.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+    if: needs.resolve-tag.outputs.tag_name != ''
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - package: swift
+            runner: macos-26
+            hosting-repo: livekit/livekit-uniffi-xcframework
+            # NOTE: spm-name must match SPM_NAME in livekit-uniffi/Makefile.toml.
+            spm-name: LiveKitUniFFI
```
1egoman: thought: Realistically, I'm not sure that it's worth trying to parameterize this job for all platforms. In the ai-coustics plugin project, I found it worked a lot better to treat each platform as its own independent pipeline which handles its own builds / packaging concerns / etc. So I think my vote would be to rename this `swift-build` and get rid of the matrix here.

.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+
+# Builds per-language wrapper packages produced by `cargo make <lang>-package`
+# (Swift xcframework today; Kotlin/Python can be added as matrix entries) and
+# publishes them to their respective hosting repos.
+#
+# Trigger pattern mirrors ffi-builds.yml: every push to main checks whether
+# knope has produced a draft release for the livekit-uniffi package; if so the
+# matching package job runs. Manual runs go through workflow_dispatch.
+#
+# NOTE: livekit-uniffi must be added to knope.toml with `assets = "marker"`
```
ladvoc: question: I don't see that added here. Did you want to add in a separate PR?

.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+  build:
+    name: Build ${{ matrix.package }} package
+    needs: resolve-tag
+    if: needs.resolve-tag.outputs.tag_name != ''
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - package: swift
+            runner: macos-26
```
ladvoc: suggestion: How about using a large runner here to speed things up (e.g. `macos-26-large`)?

.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+            aarch64-apple-ios-sim \
+            x86_64-apple-ios \
+            aarch64-apple-ios-macabi \
+            x86_64-apple-ios-macabi \
+            aarch64-apple-darwin \
+            x86_64-apple-darwin \
+            aarch64-apple-tvos \
+            aarch64-apple-tvos-sim
+
+      - name: Install nightly + rust-src (visionOS tier-3 target)
```
ladvoc: comment: [visionOS](https://doc.rust-lang.org/beta/rustc/platform-support/apple-visionos.html) has been upgrade to tier II now.

.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+        if: matrix.package == 'swift'
+        run: |
+          rustup toolchain install nightly
+          rustup component add rust-src --toolchain nightly
+
+      - uses: cargo-bins/cargo-binstall@v1
+      - run: cargo binstall cargo-make --no-confirm
+
+      - name: Install Protoc
```
ladvoc: question: Did you get an error without `protoc` installed? It shouldn't be needed here.

.github/workflows/uniffi-packages.yml
```diff
@@ -0,0 +1,145 @@
...
+        run: cargo make --profile release ${{ matrix.package }}-package
+
+      - name: Publish Swift package to hosting repo
+        if: matrix.package == 'swift' && !inputs.dry_run
+        uses: livekit/publish-xcframework-action@v1
+        with:
+          xcframework-zip: ${{ env.OUTPUT_DIR }}/Rust${{ matrix.spm-name }}.xcframework.zip
+          version: ${{ needs.resolve-tag.outputs.version }}
+          hosting-repo: ${{ matrix.hosting-repo }}
+          files: |
```
ladvoc: question: Does this fail if any of the specified files are missing? To avoid similar incomplete release issues to what we've seen with _libwebrtc_ in this repo, it probably should.

ladvoc (approved):

.github/workflows/uniffi-packages.yml
pblazej: good point, I think it should be applied elsewhere (sdk et al)

.github/workflows/uniffi-packages.yml
pblazej: `-large` is x86, `-xlarge` is arm (and paid), I don't have any exact numbers yet, may be worth trying.

pblazej:
Thanks for constructive fb, I'll revisit that in the meantime.

@ladvoc are we still fine with creating some "draft" releases of the crate for e2e test?

.github/workflows/uniffi-packages.yml
1egoman: Most livekit repos (at least in the web sphere) are doing this already, but yes if there are repos which are not pinning actions, definitely update those!

.github/workflows/uniffi-packages.yml
pblazej: Done: 753b26d6

.github/workflows/uniffi-packages.yml
pblazej: Done: a37c961f

.github/workflows/uniffi-packages.yml
pblazej: you were right, no error without protoc. Done: a37c961f

.github/workflows/uniffi-packages.yml
pblazej: addressed in livekit/publish-xcframework-action#3 — explicit `test -f` per source file + `set -euo pipefail`, no more silent partials. workflow pinned to that commit. Done: a37c961f

.github/workflows/uniffi-packages.yml
pblazej: Done: a37c961f. Also pinned `livekit/publish-xcframework-action` to the fix commit from livekit/publish-xcframework-action#3 — will repin to the main-merged SHA once that lands.

.github/workflows/uniffi-packages.yml
pblazej: same pattern as ffi-builds.yml — knope creates the draft on the release-PR merge, this workflow watches main pushes and picks it up. wiring it into release.yml directly would mean baking swift-specific build/publish logic into the release orchestrator, which feels out of place.

.github/workflows/uniffi-packages.yml
pblazej: looked into it: GH Actions can't pick `uses:` based on `matrix.*` dynamically, so "separate workflow + matrix" collapses to N hardcoded `uses:` jobs (basically your suggestion). matrix-of-one stays simpler for now; if kotlin/python end up with very divergent steps we revisit and split.

pblazej: Reconsidered — extracted to a reusable workflow (`uniffi-swift.yml`) called from a thin parent that handles tag resolution. No matrix; each future lang would be a hardcoded `uses:` job. Done: ac8941a5

.github/workflows/uniffi-packages.yml
pblazej: Correction: tvOS is still tier-3 and `cargo-swift` falls back to nightly+`-Zbuild-std` for any tier-3 arch, so I dropped tier-3 entries from the stable `rustup target add` list and put the nightly + rust-src install back. Verified end-to-end locally. Done: 6ff5f799

PR #1081, opened 2026-05-11
Proper setting of client SDK

stephen-derosa:
### PR description

Report the corect client fclient_info_sdk_for_name

### Breaking changes

I dont believe there are breaking changes

### Testing

Unit tests inline

### DIFF
--- a/.changeset/draft_proper_floating_up_of_version.md
+++ b/.changeset/draft_proper_floating_up_of_version.md
@@ -1,0 +1,7 @@
+---
+livekit-api: patch
+livekit: patch
+livekit-ffi: patch
+---
+
+Draft: proper floating up of version - #1081 (@stephen-derosa)
--- a/.changeset/draft_proper_floating_up_of_version.md
+++ b/.changeset/draft_proper_floating_up_of_version.md
@@ -1,7 +1,5 @@
 ---
 livekit-api: patch
-livekit: patch
-livekit-ffi: patch
 ---
 
-Draft: proper floating up of version - #1081 (@stephen-derosa)
+Proper client SDK - #1081 (@stephen-derosa)
--- a/livekit-api/src/signal_client/mod.rs
+++ b/livekit-api/src/signal_client/mod.rs
@@ -641,6 +641,26 @@ fn is_pass_through(signal: &proto::signal_request::Message) -> bool {
     )
 }
 
+fn client_info_sdk_for_name(sdk: &str) -> proto::client_info::Sdk {
+    match sdk.trim().to_ascii_lowercase().as_str() {
+        "js" => proto::client_info::Sdk::Js,
+        "swift" => proto::client_info::Sdk::Swift,
+        "android" => proto::client_info::Sdk::Android,
+        "flutter" => proto::client_info::Sdk::Flutter,
+        "go" => proto::client_info::Sdk::Go,
+        "unity" => proto::client_info::Sdk::Unity,
+        "react_native" | "react-native" | "reactnative" => proto::client_info::Sdk::ReactNative,
+        "rust" => proto::client_info::Sdk::Rust,
+        "python" => proto::client_info::Sdk::Python,
+        "cpp" | "c++" => proto::client_info::Sdk::Cpp,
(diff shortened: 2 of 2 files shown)


livekit-api/src/signal_client/mod.rs
```diff
@@ -641,6 +641,26 @@ fn is_pass_through(signal: &proto::signal_request::Message) -> bool {
...
+        "unity" => proto::client_info::Sdk::Unity,
+        "react_native" | "react-native" | "reactnative" => proto::client_info::Sdk::ReactNative,
+        "rust" => proto::client_info::Sdk::Rust,
+        "python" => proto::client_info::Sdk::Python,
+        "cpp" | "c++" => proto::client_info::Sdk::Cpp,
+        "unity_web" | "unity-web" | "unityweb" => proto::client_info::Sdk::UnityWeb,
+        "node" => proto::client_info::Sdk::Node,
+        "unreal" => proto::client_info::Sdk::Unreal,
+        "esp32" => proto::client_info::Sdk::Esp32,
+        _ => proto::client_info::Sdk::Unknown,
```
ladvoc: suggestion: It might be a good idea to log a warning on `_ =>` for visibility.

livekit-api/src/signal_client/mod.rs
```diff
@@ -641,6 +641,26 @@ fn is_pass_through(signal: &proto::signal_request::Message) -> bool {
...
 
+fn client_info_sdk_for_name(sdk: &str) -> proto::client_info::Sdk {
+    match sdk.trim().to_ascii_lowercase().as_str() {
+        "js" => proto::client_info::Sdk::Js,
+        "swift" => proto::client_info::Sdk::Swift,
+        "android" => proto::client_info::Sdk::Android,
+        "flutter" => proto::client_info::Sdk::Flutter,
+        "go" => proto::client_info::Sdk::Go,
+        "unity" => proto::client_info::Sdk::Unity,
+        "react_native" | "react-native" | "reactnative" => proto::client_info::Sdk::ReactNative,
```
ladvoc: suggestion: I would just match against the ["canonical" string](https://github.com/livekit/livekit/blob/ab7fdeab7c1a95f6e9a92469ee73771201b1e4d0/pkg/service/utils.go#L153) the SFU checks for when this is provided via URL parameter (for non-single peer connection).

ladvoc (approved):

ladvoc (approved):

.changeset/draft_proper_floating_up_of_version.md
```diff
@@ -0,0 +1,7 @@
+---
+livekit-api: patch
+livekit: patch
+livekit-ffi: patch
```
xianshijing-lk: I only see changes touching the livekit-api, but not other two crates

livekit-api/src/signal_client/mod.rs
xianshijing-lk: how about C# ? do we have something like proto::client_info::Sdk::CSharp ?

livekit-api/src/signal_client/mod.rs
```diff
@@ -862,6 +882,28 @@ async fn get_reconnect_response(
...
     use super::*;
 
+    fn signal_options_for_cpp(version: &str) -> SignalOptions {
+        let mut options = SignalOptions::default();
+        options.sdk_options.sdk = "cpp".to_string();
+        options.sdk_options.sdk_version = Some(version.to_string());
+        options
+    }
+
+    fn decode_join_request_param(param: &str) -> proto::JoinRequest {
```
xianshijing-lk: is this function used by test only ? any way to be more explicit if that is the case ?

xianshijing-lk (approved):
some comments, lgtm after addressing them

livekit-api/src/signal_client/mod.rs
ladvoc: Currently that is covered by `proto::client_info::Sdk::Unity`.

livekit-api/src/signal_client/mod.rs
stephen-derosa: updated the name to `decode_join_request_param_for_test`

PR #1089, opened 2026-05-13
Add env var to set preferred encoder

chenosaurus:
On system that have both a Nvidia GPU & an Intel or AMD processor with hardware video encoders, sometimes it's preferred to use the VAAPI hw encoder for video.  Currently, the SDK prefers NVENC if available and will bypass the VAAPI encoders.

This PR adds a check for a `PREFERRED_HW_ENCODER` env var that can be set to either `nvenc` or `vaapi` to explicitly set which encoder should take priority if it's available.

### DIFF
--- a/webrtc-sys/src/video_encoder_factory.cpp
+++ b/webrtc-sys/src/video_encoder_factory.cpp
@@ -16,6 +16,9 @@
 
 #include "livekit/video_encoder_factory.h"
 
+#include <cstdlib>
+#include <string_view>
+
 #include "api/environment/environment_factory.h"
 #include "api/video_codecs/sdp_video_format.h"
 #include "api/video_codecs/video_encoder.h"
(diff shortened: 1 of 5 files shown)


webrtc-sys/src/video_encoder_factory.cpp
```diff
@@ -47,6 +50,86 @@
 
 namespace livekit_ffi {
 
+namespace {
+
+constexpr char kPreferredHwEncoderEnv[] = "PREFERRED_HW_ENCODER";
```
ladvoc: Maybe this should be prefixed with `LK_` or `LIVEKIT_`?

ladvoc (approved):
Looks good ✅. Do you think we should expose this in the SDK rather than just as an environment variable? If this ever needs to be configured dynamically by the application (which I am not sure you would ever need to do), it might be worth exposing. If we stick with the environment variable, it should be documented somewhere that it exists (maybe the README is the best place for now).

cloudwebrtc (approved):

PR #1092, opened 2026-05-14
Reword audio filter logs to be less confusing

1egoman:
This log can occur in potentially multiple cases, and it's not super clear exactly all of them. So, increase the log verbosity of the `debug!` log to `warn!`, and make the main message a little less definitively communicate "connecting to livekit cloud is the problem".


ladvoc (approved):
LGTM ✅

PR #1093, opened 2026-05-14
Return EOS event from data track stream read request

ladvoc:
Resolves BOT-347


stephen-derosa (approved):
nice, LGTM!

CPP Side MR: https://github.com/livekit/client-sdk-cpp/pull/130

1egoman (approved):

PR #1096, opened 2026-05-15
Always emit Disconnected on engine close

MaxHeimbrock:
## Summary

`RtcEngineInner::close` only emitted `EngineEvent::Disconnected` from inside the `if let Some(engine_task)` block. Each failed `try_restart_connection` calls `running_handle.engine_task.take()` and does not restore it on error, so after the reconnect loop exhausts all attempts and calls `close(UnknownReason)`, `engine_task` is already `None` and the disconnect event is never sent. The room task never runs `handle_disconnected`, the connection state stays at Reconnecting, and `RoomEvent::Disconnected` is never dispatched — consumers see the room appear permanently stuck in reconnect after a network outage.

## Repro
Use Unity SDK Meet sample and lock the screen. Android will remove network access to the app and it will try 10 reconnects, until it fails eventually. In browser I can observe the Android participant being removed.
Unlock Android phone again and without the fix, I did not receive a Room.OnDisconnected event, the state is still `Reconnecting` and not disconnected.

## Changes

Move the emit outside the if-let so the event always fires on close. Duplicates are absorbed by `update_connection_state` on the room side.

### DIFF
--- a/livekit/src/rtc_engine/mod.rs
+++ b/livekit/src/rtc_engine/mod.rs
@@ -664,8 +664,13 @@ impl EngineInner {
             session.close(reason).await;
             let _ = close_tx.send(());
             let _ = engine_task.await;
-            let _ = self.engine_tx.send(EngineEvent::Disconnected { reason });
         }
+
+        // Always emit Disconnected, even when the engine_task was already taken by a
+        // prior failed `try_restart_connection`. Without this, a reconnect cycle that
+        // exhausts all attempts leaves the room stuck in Reconnecting forever because
+        // the room's task never sees the event that drives `handle_disconnected`.
+        let _ = self.engine_tx.send(EngineEvent::Disconnected { reason });
     }
 
     /// When waiting for reconnection, it ensures we're always using the latest session.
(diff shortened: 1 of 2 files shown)


livekit/src/rtc_engine/mod.rs
```diff
@@ -680,8 +664,13 @@ impl EngineInner {
...
             let _ = close_tx.send(());
             let _ = engine_task.await;
-            let _ = self.engine_tx.send(EngineEvent::Disconnected { reason });
         }
+
+        // Always emit Disconnected, even when the engine_task was already taken by a
+        // prior failed `try_restart_connection`. Without this, a reconnect cycle that
+        // exhausts all attempts leaves the room stuck in Reconnecting forever because
+        // the room's task never sees the event that drives `handle_disconnected`.
+        let _ = self.engine_tx.send(EngineEvent::Disconnected { reason });
```
xianshijing-lk, line 673: any chance that there could be duplicated disconnected events ? 

what will happen if close() is called multiple times ?

ladvoc (approved):
LGTM ✅

xianshijing-lk (approved):

xianshijing-lk:
I'm OK with this quick fix for now, but we should review the connection state machine more carefully.

MaxHeimbrock:
> I'm OK with this quick fix for now, but we should review the connection state machine more carefully.

Lets do that, saw your post on Slack and I also think this is the right way if we can spend the time.

PR #1097, opened 2026-05-15
change the sample rate and num of channel optional in the proto

xianshijing-lk:
### Before you submit your PR

### PR description

change the sample rate and num of channel optional in the proto
Those fileds are not needed by the PlatformAudio source

### Breaking changes

No
### MSRV

N/A

### Testing

CI

### Async

N/A

### DIFF
--- a/.changeset/optional_audio_source_params.md
+++ b/.changeset/optional_audio_source_params.md
@@ -1,5 +1,5 @@
 ---
-livekit-ffi: patch
+livekit-ffi: minor
 ---
 
 Make `sample_rate` and `num_channels` optional in `NewAudioSourceRequest`.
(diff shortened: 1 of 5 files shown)


.changeset/optional_audio_source_params.md
```diff
@@ -0,0 +1,7 @@
+---
+livekit-ffi: patch
```
ladvoc: nitpick: We haven't been following semver consistently for _livekit-ffi_, but we should probably make this a minor bump just so it is clear to FFI client maintainers this is a breaking change.

ladvoc (approved):
LGTM ✅

.changeset/optional_audio_source_params.md
xianshijing-lk: changing the proto field from required to optional is not a breaking change. 

and I think livekit-ffi is mostly consumed internally ?

xianshijing-lk: Done with changing it from patch to minor

PR #1098, opened 2026-05-17
webrtc-sys: don't panic when C++ hands us a malformed RtcError string

c-tonneslan:
Closes #944.

The unsafe `RtcError::from(&str)` wrapper was `.expect("malformed serialized RtcError")` on the inner `parse()` Option, which panicked the caller's task whenever C++ handed across a string that didn't match the fixed-width hex header. Per #944 that happens on participant disconnect / call end (empty `what()`).

Fall back to a generic `RtcErrorType::None` and put the raw input in `message` so the error still surfaces something useful instead of taking down the task.

Added a quick test covering the empty-string and non-hex cases that used to panic.

### DIFF
--- a/.changeset/fix_malformed_rtc_error_handling.md
+++ b/.changeset/fix_malformed_rtc_error_handling.md
@@ -1,0 +1,8 @@
+---
+webrtc-sys: patch
+libwebrtc: patch
+livekit: patch
+livekit-ffi: patch
+---
+
+# Fix malformed RTC error handling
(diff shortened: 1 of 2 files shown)


ladvoc:
Hi @c-tonneslan, this looks good to me! Please add a changeset at `.changeset/fix_malformed_rtc_error_handling.md`:
```yaml
---
‎webrtc-sys: patch
libwebrtc: patch
livekit: patch
livekit-ffi: patch
---

# Fix malformed RTC error handling
```
Then I will merge this.

ladvoc (approved):

PR #1099, opened 2026-05-17
fix: don't fire local_track_subscribed during reconnect

davidzhao:
During a full reconnect, the engine auto-resubscribes the republished local tracks under their new server-issued SIDs. Forwarding this event would race with the sid rekey driven by `LocalTrackRepublished` (different task, same dispatcher channel) and surface as a KeyError / already-resolved future on the binding side.

### DIFF
--- a/.changeset/fix_don_t_fire_local_track_subscribed_during_reconnect.md
+++ b/.changeset/fix_don_t_fire_local_track_subscribed_during_reconnect.md
@@ -1,0 +1,5 @@
+---
+livekit-ffi: patch
+---
+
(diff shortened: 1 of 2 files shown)


ladvoc (approved):
LGTM ✅

theomonnom:
For this fix, I'm wondering if it should be inside the rtc_engine vs inside the ffi?

davidzhao:
> For this fix, I'm wondering if it should be inside the rtc_engine vs inside the ffi?

what is this event used for by local clients? I know that FFI clients def don't need it.. because the handle is abstracted away locally.

PR #1100, opened 2026-05-18
FFI logging improvements

ladvoc:
Summary of changes:
- Make some logs debug level
- Log warning when attempting to drop an unknown handle
- Disable capture logs after teardown completes

### DIFF
--- a/livekit-ffi/src/server/mod.rs
+++ b/livekit-ffi/src/server/mod.rs
@@ -151,7 +151,7 @@ impl FfiServer {
         *self.config.lock() = Some(config.clone());
         self.logger.set_capture_logs(config.capture_logs);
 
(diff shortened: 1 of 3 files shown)


alan-george-lk (approved):

livekit-ffi/src/server/mod.rs
```diff
@@ -161,7 +161,7 @@ impl FfiServer {
 
     pub async fn dispose(&'static self) {
         self.logger.set_capture_logs(false);
-        log::info!("disposing ffi server");
+        log::debug!("disposing ffi server");
```
alan-george-lk, line 163: Consider dropping the capture logs after this line, that way this one doesn't creep through always

alan-george-lk (approved):

PR #1101, opened 2026-05-18
Improved data tracks depacketizer to support mutliple in flight packets

ladvoc:
Equivalent to the implementation in: https://github.com/livekit/client-sdk-js/pull/1923

Closes BOT-342


livekit-datatrack/src/remote/depacketizer.rs
```diff
@@ -14,104 +14,150 @@
...
+        // is surfaced in `drop_error`; the rest are silently dropped to keep the public
+        // `DepacketizerPushResult` shape unchanged.
+        while self.partials.len() >= options.max_partial_frames {
+            let Some(evicted) = self.evict_oldest(frame_number) else {
+                // Partials map is empty - nothing more to evict.
+                break;
+            };
+            if result.drop_error.is_none() {
+                result.drop_error = Some(evicted);
             }
```
1egoman, line 130: thought: It it worth logging the `drop_error`s which aren't being returned here so they are at least surfaced to the user somehow?

livekit-datatrack/src/remote/manager.rs
```diff
@@ -516,8 +536,12 @@ mod tests {
         let sid = info.sid();
         let publisher_identity: Arc<str> = Faker.fake::<String>().into();
 
-        let pipeline_opts =
-            PipelineOptions { info: info.clone(), publisher_identity, decryption_provider: None };
+        let pipeline_opts = PipelineOptions {
+            info: info.clone(),
+            publisher_identity,
+            decryption_provider: None,
+            max_partial_frames: Arc::new(AtomicUsize::new(1)),
```
1egoman: nitpick: It might be worth making this `max_partial_frames` value default of `1` a constant - it is used enough (at least a few spots) where I think it would be beneficial.

1egoman (approved):

livekit-datatrack/src/remote/depacketizer.rs
ladvoc: All drop errors are currently logged at debug level [here](https://github.com/livekit/rust-sdks/blob/df2787d066566c09fd2dd3025f31bc81dae02e7c/livekit-datatrack/src/remote/pipeline.rs#L69).

livekit-datatrack/src/remote/manager.rs
ladvoc: The source of truth for this is `RemoteDataTrackPipelineOptions::default()`; updated to use that here too.

PR #1104, opened 2026-05-20
fix Android PlatformAudio

xianshijing-lk:
### Before you submit your PR

### PR description

⏺ Android PlatformAudio: Problems & Solutions

  Problem 1: ADM Creation Failed

  Symptom:
  CreateAudioDeviceModule returned nullptr
  Failed to initialize PlatformAudio

  Root Cause:
  WebRTC's CreateAndroidAudioDeviceModule internally calls GetAppContext() to get the Android application context. But GetAppContext() returns null unless ContextUtils.initialize(context) was called first.

  CreateAndroidAudioDeviceModule()
      └── GetAppContext()
              └── Returns null ❌ (ContextUtils not initialized)

  Solution:
  Pass the Android Context from Unity to Rust, then call ContextUtils.initialize(context) before any WebRTC audio initialization.

  Unity                          Rust/C++
  ─────                          ────────
  GetApplicationContext() ──────► livekit_ffi_initialize_android_context()
                                      │
                                      ├── do_android_init(vm)      // JVM init
                                      └── ContextUtils.initialize(context)  // Now GetAppContext() works!

  Files changed:
  - livekit-ffi/src/cabi.rs - Added livekit_ffi_initialize_android_context()
  - webrtc-sys/src/android.cpp - Added init_android_context()
  - client-sdk-unity/.../FFIClient.cs - Calls the init with context
  - client-sdk-unity/.../NativeMethods.cs - P/Invoke declaration

  ---
  Problem 2: Wrong ADM Factory Function

  Symptom:
  AdmProxy: All Android audio layers failed!

  Root Cause:
  The code was using the generic CreateAudioDeviceModule() which doesn't work properly on Android. Android requires CreateAndroidAudioDeviceModule() from the Android-specific header.

  Solution:
  Changed adm_proxy.cpp to use the correct function:

  // Before (wrong):
  #include "modules/audio_device/include/audio_device.h"
  platform_adm_ = webrtc::AudioDeviceModule::Create(...);

  // After (correct):
  #include "sdk/android/native_api/audio_device_module/audio_device_android.h"
  platform_adm_ = webrtc::CreateAndroidAudioDeviceModule(...);

  File changed:
  - webrtc-sys/src/adm_proxy.cpp

  ---
  Problem 3: Empty GUID Validation Failed

  Symptom:
  Recording device at index 0 has no GUID

  Root Cause:
  Unity SDK threw an exception when device GUID was empty. But on Android, devices don't have GUIDs - WebRTC returns empty strings.

  Solution:
  Allow empty GUIDs and pass them to native code, which falls back to index 0:

  // Before:
  if (string.IsNullOrEmpty(deviceId))
      throw new InvalidOperationException("has no GUID");

  // After:
  SetRecordingDevice(deviceId ?? "");  // Empty string triggers fallback

  File changed:
  - client-sdk-unity/.../PlatformAudio.cs

  ---
  Problem 4: Device Not Found with Empty GUID

  Symptom:
  Failed to set recording device: Device not found

  Root Cause:
  Native code searched for a device matching the empty GUID string, found nothing.

  Solution:
  Added fallback in C++ - if GUID is empty and devices exist, use index 0:

  bool AudioDeviceController::set_recording_device_by_guid(rust::String guid) const {
      int16_t count = adm_proxy_->RecordingDevices();

      // Android fallback: empty GUID → use index 0
      if (guid.empty() && count > 0) {
          return adm_proxy_->SetRecordingDevice(0) == 0;
      }
      // ... normal GUID lookup for desktop
  }

  File changed:
  - webrtc-sys/src/audio_device_controller.cpp

### Breaking changes

N/A
### MSRV

N/A
### Testing

Tested with Unity

### Async

N/A

### DIFF
--- a/.changeset/android_platform_audio_context_init.md
+++ b/.changeset/android_platform_audio_context_init.md
@@ -1,0 +1,19 @@
+---
+livekit-ffi: minor
+livekit: patch
+libwebrtc: patch
+webrtc-sys: patch
+---
+
+feat: add Android application context initialization for PlatformAudio support.
+
+Android requires `ContextUtils.initialize(applicationContext)` before WebRTC audio components can be created. This change:
+
+- Adds `livekit_ffi_initialize_android_context()` C FFI function for Unity and other FFI consumers
+- Uses `CreateAndroidAudioDeviceModule()` instead of generic `CreateAudioDeviceModule()` on Android
+- Handles empty device GUIDs on Android (falls back to index 0)
+- Documents Android-specific limitations: single default device, no app-level device selection
+
+Platform notes:
+- Android device enumeration returns only one "default" device with empty name/GUID
+- Audio routing (speaker/earpiece/Bluetooth) is controlled by Android's AudioManager, not WebRTC
--- a/examples/mobile/src/lib.rs
+++ b/examples/mobile/src/lib.rs
@@ -441,6 +441,8 @@ pub mod android {
         );
 
         log::info!("JNI_OnLoad, initializing LiveKit");
+        // Early JVM init for video/data. PlatformAudio requires additional context init
+        // via initializeContextNative() - see Java_io_livekit_rustexample_App_initializeContextNative.
         livekit::webrtc::android::initialize_android(&vm);
         JNI_VERSION_1_6
     }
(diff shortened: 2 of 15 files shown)


cloudwebrtc (approved):

xianshijing-lk:
Going to land it now and make a ffi release. 

In case you have more comments, I will address them in follow up PRs

PR #1105, opened 2026-05-20
Add support for multi video track subscriber in local_video example

chenosaurus:
- In the `local_video` example, when there are multiple video tracks published by the target participant, render each video in its own window.


ladvoc (approved):
LGTM ✅. Do you think at some point it makes sense to add missing features to the *wgpu_room* example to supersede this?

PR #1106, opened 2026-05-20
Build docs workflow

ladvoc:
Validates docs build without any warnings (e.g., invalid links).

Closes CLT-3120

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,10 +35,12 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Setup Rust Toolchain
         uses: actions-rust-lang/setup-rust-toolchain@46268bd060767258de96ed93c1251119784f2ab6 # v1.16.1
+        with:
+          rustflags: ""
       - name: Cargo Doc
         env:
           RUSTDOCFLAGS: "-D warnings" # Fail on warnings
-        run: cargo doc --no-deps
+        run: cargo doc --no-deps --workspace
       - name: Upload
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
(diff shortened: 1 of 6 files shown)


alan-george-lk (approved):

.github/workflows/docs.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Setup Rust Toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@46268bd060767258de96ed93c1251119784f2ab6 # v1.16.1
+        with:
+          rustflags: ""
+      - name: Cargo Doc
+        env:
+          RUSTDOCFLAGS: "-D warnings" # Fail on warnings
+        run: cargo doc --no-deps --workspace
+      - name: Upload
```
xianshijing-lk, line 44: where does it upload the doc to ? 

I wonder if you should upload it to S3 so that docs.livekit.io can reference the docs there ?

ladvoc: All crates published to crates.io automatically get their docs built and published to docs.rs, so you wouldn't typically self-host. I upload as an artifact here because it's sometimes useful to see the built docs for a specific branch.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,50 @@
...
+
+name: Docs
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+    paths: ["**/*.rs"]
+  push:
+    branches: [main]
+    paths: ["**/*.rs"]
```
xianshijing-lk, line 23: nit, does cargo doc get broken due to cargo config changes ? should we consider things like 
paths:
  - "**/*.rs"
  - "**/Cargo.toml"
  - "Cargo.lock"

.github/workflows/docs.yml
xianshijing-lk: Got it, thanks for clarification

xianshijing-lk (approved):
one nit.

PR #1109, opened 2026-05-20
Refactor data track E2E test

ladvoc:
Test delivery behavior, not throughput.

### DIFF
--- a/livekit/tests/data_track_test.rs
+++ b/livekit/tests/data_track_test.rs
@@ -46,15 +46,16 @@ async fn test_data_track(payload_len: usize) {
     let remote_track = wait_for_remote_track(&mut sub_room_event_rx).await.unwrap();
     log::info!("Got remote track: {}", remote_track.info().sid());
 
+    const PAYLOAD_VALUE: u8 = 0xFA;
+
     let publish = async move {
         assert!(local_track.is_published());
         assert!(!local_track.info().uses_e2ee());
         assert_eq!(local_track.info().name(), "my_track");
 
-        let mut index: u8 = 0;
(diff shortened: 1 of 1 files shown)


stephen-derosa (approved):

livekit/tests/data_track_test.rs
```diff
@@ -30,84 +30,65 @@ use {
...
         assert_eq!(local_track.info().name(), "my_track");
 
-        let sleep_duration = Duration::from_secs_f64(1.0 / publish_fps as f64);
-        for index in 0..frame_count {
-            local_track.try_push(vec![index as u8; payload_len].into()).unwrap();
-            time::sleep(sleep_duration).await;
+        let mut index: u8 = 0;
+        loop {
+            local_track.try_push(vec![index; payload_len].into()).unwrap();
+            index = index.wrapping_add(1);
```
alan-george-lk: Maybe not visible in this diff, is index actually used for anything with this change?

alan-george-lk (approved):

livekit/tests/data_track_test.rs
alan-george-lk: Edit: nevermind it's used in the line above (facepalm)

ladvoc: No, you're right. Since this test ends after the first frame is received, there's no reason to increment the index and use that as the repeated byte to check for. Changed in [87912b5](https://github.com/livekit/rust-sdks/pull/1109/commits/87912b597f9090fcb61899ac7d63ca5c4562d40b) to use a fixed byte.

PR #1112, opened 2026-05-21
Filter internal data streams out of livekit-ffi interface

1egoman:
Without this, logs like the below would show up in the node sdk while rpc requests are occurring with rpc v2 capable participants:

> ignoring incoming text stream due to no handler for topic: lk.rpc_request

So, adds a new mechanism to specify "internal topics" which should not be exposed into the room's public api interface. Also adds a new e2e test to exercise this case and ensure that these events aren't inadvertently exposed (which if they aren't exposed in `livekit`, they won't be exposed in `livekit-ffi`, which means they won't be exposed in end sdks).

### DIFF
--- a/livekit/src/room/data_stream/incoming.rs
+++ b/livekit/src/room/data_stream/incoming.rs
@@ -298,7 +298,7 @@ impl IncomingStreamManager {
     /// (e.g. `lk.rpc_request`). Used to suppress `RoomEvent::Stream*Received`
     /// dispatches for traffic the SDK handles itself.
     pub fn is_internal(&self, stream_id: &str) -> bool {
-        self.inner.lock().open_streams.get(stream_id).map(|d| d.is_internal).unwrap_or(false)
+        self.inner.lock().open_streams.get(stream_id).is_some_and(|d| d.is_internal)
     }
 
     /// Handles an incoming chunk packet.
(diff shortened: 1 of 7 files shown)


livekit/src/room/data_stream/incoming.rs
```diff
@@ -287,10 +289,18 @@ impl IncomingStreamManager {
...
+            is_internal,
         };
         inner.open_streams.insert(id, descriptor);
     }
 
+    /// Returns whether the given open stream belongs to an internal topic
+    /// (e.g. `lk.rpc_request`). Used to suppress `RoomEvent::Stream*Received`
+    /// dispatches for traffic the SDK handles itself.
+    pub fn is_internal(&self, stream_id: &str) -> bool {
+        self.inner.lock().open_streams.get(stream_id).map(|d| d.is_internal).unwrap_or(false)
```
xianshijing-lk: nit, will it be slightly cleaner if doing is_some_and(|d| d.is_internal) ? like
pub fn is_internal(&self, stream_id: &str) -> bool {
    self.inner.lock().open_streams.get(stream_id).is_some_and(|d| d.is_internal)
}

xianshijing-lk (approved):
one nit. lgtm in general

livekit/src/room/data_stream/incoming.rs
1egoman: Done in 99f4f90

PR #1113, opened 2026-05-21
Fix depacketizer test compilation error

xianshijing-lk:
Use correct variable for assertion in test_interrupted test. The new_frame_number field is destructured into `reported`, so we need to compare against the actual new frame number value.

### Before you submit your PR

- [ x] I have read the [contributing guidelines](https://github.com/livekit/rust-sdks/blob/main/CONTRIBUTING.md) and validated that this PR will be accepted.
- [x ] I have read and followed the principles regarding breaking changes, testing, and code quality.

### PR description

Fix the CI.

### DIFF
--- a/livekit-datatrack/src/remote/depacketizer.rs
+++ b/livekit-datatrack/src/remote/depacketizer.rs
@@ -359,7 +359,8 @@ mod tests {
         assert!(result.frame.is_none() && result.drop_error.is_none());
 
         let first_frame_number = packet.header.frame_number;
-        packet.header.frame_number = packet.header.frame_number.wrapping_add(1); // Next frame
+        let new_frame_number = packet.header.frame_number.wrapping_add(1);
+        packet.header.frame_number = new_frame_number; // Next frame
 
         let result = depacketizer.push(packet, Default::default());
         assert!(result.frame.is_none());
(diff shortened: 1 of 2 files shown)


livekit-datatrack/src/remote/depacketizer.rs
```diff
@@ -369,7 +369,7 @@ mod tests {
         let DepacketizerDropReason::Interrupted { new_frame_number: reported } = drop.reason else {
             panic!("Expected Interrupted, got {:?}", drop.reason);
         };
-        assert_eq!(reported, new_frame_number);
+        assert_eq!(reported, first_frame_number.wrapping_add(1));
```
1egoman: In case its useful, here's the corresponding test in the js implementation: https://github.com/livekit/client-sdk-js/blob/283048fb5414b594f64969c59687281cdfd73719/src/room/data-track/depacketizer.test.ts#L80-L112

The web tests aren't asserting the frame number on the error it seems, though.

1egoman (commented):
I think this looks good to me but I'd like @ladvoc to give the final ✅ since he ported these tests.

1egoman (approved):
Approving to unblock the build, @ladvoc please do a post merge review on this!

ladvoc (approved):
LGTM ✅, I will follow up to find out how CI checks passed when this was merged originally.

PR #1115, opened 2026-05-24
Fix AV1 packet trailers

chenosaurus:
## What this fixes

Fixes AV1 packet trailer delivery when packet trailers are enabled.

Previously, packet trailers were appended as raw bytes at the end of the encoded frame. That works for codecs like VP8, but breaks AV1 because the subscriber-side decoder sees the appended bytes as invalid AV1 bitstream data. This caused AV1 subscribers to fail/loop on recovery behavior when packet trailers were enabled.

## What changed

- Encodes LiveKit packet trailer metadata for AV1 as an AV1 Metadata OBU instead of appending raw trailer bytes.
- Inserts the metadata OBU after leading AV1 sequence header OBUs when present.
- Extracts and strips the AV1 metadata OBU on receive before the frame reaches the decoder.
- Keeps the existing raw trailing trailer behavior for non-AV1 codecs.
- Adds AV1 coverage to the packet trailer tests.

## Validation

- `cargo fmt`
- `cargo check -p webrtc-sys`
- `cargo test -p livekit --features default,__lk-e2e-test --test packet_trailer_test -- --nocapture --test-threads=1`
- Verified with local SFU using publisher `--test-pattern --codec av1 --attach-timestamp --attach-frame-id`; subscriber decoded AV1 and reported packet trailer features successfully.

### DIFF
--- a/.changeset/av1-packet-trailers.md
+++ b/.changeset/av1-packet-trailers.md
@@ -1,0 +1,8 @@
+---
+webrtc-sys: patch
+libwebrtc: patch
+livekit: patch
+livekit-ffi: patch
+---
+
+Fix AV1 subscriber decode when packet trailers are enabled.
--- a/.changeset/disable-nvdec-env.md
+++ b/.changeset/disable-nvdec-env.md
@@ -1,0 +1,8 @@
+---
+webrtc-sys: patch
+libwebrtc: patch
+livekit: patch
+livekit-ffi: patch
+---
+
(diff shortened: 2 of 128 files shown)


ladvoc:
@chenosaurus, is this ready for review?

chenosaurus:
> @chenosaurus, is this ready for review?

yea, after I resolve all the conflicts.

ladvoc (approved):

PR #1118, opened 2026-05-26
local video example publisher/subscriber pipeline timing

chenosaurus:
- add detailed timing metrics to `examples/local_video` publisher & subscriber apps.
- add timing hooks in packet_trailer transformer
- add a --codec flag to `examples/local_video/src/publisher.rs`
image

### DIFF
--- a/livekit/src/room/e2ee/manager.rs
+++ b/livekit/src/room/e2ee/manager.rs
@@ -39,6 +39,41 @@ use std::fmt::Debug;
 
 type StateChangedHandler = Box<dyn Fn(ParticipantIdentity, EncryptionState) + Send>;
 
+fn has_packet_trailer_features(features: &[i32]) -> bool {
+    features.iter().any(|f| {
+        *f == PacketTrailerFeature::PtfUserTimestamp as i32
+            || *f == PacketTrailerFeature::PtfFrameId as i32
+    })
+}
+
+fn needs_video_receiver_packet_trailer_handler(features: &[i32]) -> bool {
+    has_packet_trailer_features(features)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn receiver_packet_trailer_handler_is_skipped_without_features() {
+        assert!(!needs_video_receiver_packet_trailer_handler(&[]));
+    }
+
+    #[test]
+    fn receiver_packet_trailer_handler_is_needed_for_user_timestamp() {
+        let features = [PacketTrailerFeature::PtfUserTimestamp as i32];
+
+        assert!(needs_video_receiver_packet_trailer_handler(&features));
+    }
+
+    #[test]
+    fn receiver_packet_trailer_handler_is_needed_for_frame_id() {
+        let features = [PacketTrailerFeature::PtfFrameId as i32];
+
+        assert!(needs_video_receiver_packet_trailer_handler(&features));
+    }
+}
+
 struct ManagerInner {
     options: Option<E2eeOptions>, // If Some, it means the e2ee was initialized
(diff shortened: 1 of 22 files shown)


livekit/src/room/e2ee/manager.rs
```diff
@@ -39,6 +39,41 @@ use std::fmt::Debug;
...
+            || *f == PacketTrailerFeature::PtfFrameId as i32
+    })
+}
+
+fn needs_video_receiver_packet_trailer_handler(features: &[i32]) -> bool {
+    has_packet_trailer_features(features)
+}
+
+#[cfg(test)]
+mod tests {
```
ladvoc: nitpick: Usually unit test module goes at the bottom of the file.

livekit/src/room/participant/local_participant.rs
```diff
@@ -57,6 +57,52 @@ const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
...
 
+fn needs_video_sender_transformer(
+    options: &TrackPublishOptions,
+    has_publish_timing_observer: bool,
+) -> bool {
+    !options.packet_trailer_features.is_empty() || has_publish_timing_observer
+}
+
+#[cfg(test)]
+mod tests {
```
ladvoc: nitpick: Usually unit test module goes at the bottom of the file.

livekit/src/room/track/local_video_track.rs
```diff
@@ -26,6 +34,57 @@ pub struct LocalVideoTrack {
...
+}
+
+type PublishTimingObserverFn = dyn Fn(PublishTimingEvent) + Send + Sync + 'static;
+
+/// Callback invoked for native local video publish timing events.
+pub type PublishTimingObserver = Box<dyn Fn(PublishTimingEvent) + Send + Sync + 'static>;
+
+/// Stage reached by a native local video frame in the publish pipeline.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PublishTimingStage {
```
ladvoc: suggestion: These types are pretty webrtc specific, so it might make more sense to re-export them instead of redefining:
```rust
pub use libwebrtc::{native::packet_trailer::{PublishTimingEvent, ...}};
```

livekit/src/room/track/local_video_track.rs
```diff
@@ -126,18 +186,72 @@ impl LocalVideoTrack {
         self.source.clone()
     }
 
+    /// Set a callback for native local video publish pipeline timing events.
+    ///
+    /// The observer is invoked from WebRTC worker threads and should avoid
+    /// blocking. Pass `None` to clear a previously registered observer.
+    pub fn set_publish_timing_observer(&self, observer: Option<PublishTimingObserver>) {
```
ladvoc: suggestion (non-blocking): API-wise, it would be cleaner to expose this as an async stream and hide the raw callback. Usage would look like this:
```rust
let mut events = local_video_track.publish_timing_events();
while let Some(event) = events.next().await {
   log::info!("Got timing event: {:?}", event);
}
```

This can be implemented using a bounded `tokio::sync::broadcast` channel, where the underlying callback sends onto the channel (also means it's not possible for the user to accidentally do work on the WebRTC worker thread). `publish_timing_events` would return `PublishTimingEventStream`, which implements `Stream` and yields `PublishTimingEvent`. See [`DataTrackStream`](https://github.com/livekit/rust-sdks/blob/1972db7cdc394e3033772a96cb3452d69faba8ca/livekit-datatrack/src/remote/mod.rs#L135-L134) which is implemented in the same way.

ladvoc (approved):

PR #1119, opened 2026-05-26
add support for Apple VideoToolbox AV1 decoder

chenosaurus:
- On M3+ macs, use VideoToolbox hardware AV1 decoder instead of Dav1d for decoding AV1 video tracks.


ladvoc:
You can test with this [E2E test](https://github.com/livekit/rust-sdks/blob/1848e77a32ca2f302a58323a7a21858165d81aad/livekit/tests/video_test.rs#L42) locally.

PR #1121, opened 2026-05-27
bump protocol to v1.46.4

lukasIO:
### PR description

- bumps the protocol to v1.46.4 and adds the `livekit_analytics.proto` messages to the generated stubs.
- exposes participant `client_protocol` via FFI
- Adds a `remove_participant_with_options` method that allows to pass new remove options to the API without a previous `remove_participant` call breaking. 

### Breaking changes

for the `livekit-api` crate I've added a `derive(default)` and an additional (option) field to the EgressListOptions struct. I don't see a good way around this though. Open for suggestions.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Generated from the LiveKit protocol definitions; do not edit by hand.
+livekit-protocol/src/livekit.rs linguist-generated=true
+livekit-protocol/src/livekit.serde.rs linguist-generated=true
--- a/.changeset/bump_protocol_to_v1_46_4.md
+++ b/.changeset/bump_protocol_to_v1_46_4.md
@@ -1,0 +1,10 @@
+---
+livekit: patch
+livekit-api: patch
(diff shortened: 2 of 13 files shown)


.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Generated from the LiveKit protocol definitions; do not edit by hand.
+livekit-protocol/src/livekit.rs linguist-generated=true
+livekit-protocol/src/livekit.serde.rs linguist-generated=true
```
ladvoc, line 3: suggestion: Consider also listing generated files from _livekit-ffi-node-bindings/proto_.

ladvoc (approved):
LGTM ✅

ladvoc:
> for the livekit-api crate I've added a derive(default) and an additional (option) field to the EgressListOptions struct. I don't see a good way around this though. Open for suggestions.

Without a larger redesign, I think this is the best you can do.

PR #1036, opened 2026-04-22
Add a pr template and CONTRIBUTING.md.

reenboog:
Describe expectations for external/internal contributors regarding PR structure, testing and code quality.

PR #1122, opened 2026-05-27
Pin Rust toolchain

ladvoc:
Summary of changes:
- Pin to specific toolchain version to ensure consistent local and CI builds
- Enable Dependabot to keep this up-to-date with new releases

In a follow up PR, this will be leveraged to improve CI caching.

Closes CLT-2445

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: rust-toolchain
+    directory: /
+    schedule:
+      interval: weekly
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "1.95.0"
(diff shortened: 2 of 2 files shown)


NWPlayer123:
Yesterday was part of the six-week release cadence so 1.96.0 is now the most recent stable release, worth updating so that Dependabot doesn't immediately open a PR once this gets merged.

Worth also figuring out what y'alls MSRV is, I see #1036 from a month ago mentions this in your PR template. Your biggest dependent crate on crates.io is an optional dependency from Foxglove which has an MSRV of 1.85 since they wanted 2024 Edition; that would probably be the most logical MSRV, but you'd have to actually verify that, your best bet would just to be to add a CI check for `cargo check --workspace` using `dtolnay/rust-toolchain@1.85` (or just use cargo-msrv's `verify`).

1egoman (approved):

ladvoc:
@NWPlayer123, agree we need to document MSRV—I will address this in a follow-up.

PR #1125, opened 2026-05-28
add clock with millisecond display to local_video example for timing tests

chenosaurus:
- add a new binary to the local_video examples that renders a on-screen clock with milliseconds for measuring timing.

image


ladvoc (approved):
LGTM ✅. Later we could consider moving this to a separate repo and do standalone builds + binary distributions—seems like a generally useful utility even when working outside the Rust repo.

PR #1126, opened 2026-05-28
adding support to setting preferred encoder on video track publication

chenosaurus:
- Add field to `TrackPublishOptions` that allows the user to set a preferred video encoder when there are multiple options.
- Options can be `auto`, `software`, `hardware`, `nvenc`, `vaapi`, `videotoolbox`.

### DIFF
--- a/examples/local_video/src/publisher.rs
+++ b/examples/local_video/src/publisher.rs
@@ -67,7 +67,7 @@ enum PublisherEncoder {
     Hardware,
     Nvenc,
     Vaapi,
-    #[value(alias = "videotoolbox")]
+    #[value(name = "videotoolbox")]
     VideoToolbox,
 }
 
@@ -79,7 +79,7 @@ impl PublisherEncoder {
             PublisherEncoder::Hardware => "hardware",
             PublisherEncoder::Nvenc => "nvenc",
             PublisherEncoder::Vaapi => "vaapi",
-            PublisherEncoder::VideoToolbox => "video-toolbox",
+            PublisherEncoder::VideoToolbox => "videotoolbox",
         }
     }
 }
@@ -104,7 +104,7 @@ fn video_encoder_backend_name(backend: VideoEncoderBackend) -> &'static str {
         VideoEncoderBackend::Hardware => "hardware",
         VideoEncoderBackend::Nvenc => "nvenc",
         VideoEncoderBackend::Vaapi => "vaapi",
-        VideoEncoderBackend::VideoToolbox => "video-toolbox",
+        VideoEncoderBackend::VideoToolbox => "videotoolbox",
         _ => "unknown",
     }
 }
--- a/examples/local_video/src/publisher.rs
+++ b/examples/local_video/src/publisher.rs
@@ -3,7 +3,7 @@ use clap::{Parser, ValueEnum};
 use livekit::e2ee::{key_provider::*, E2eeOptions, EncryptionType};
 use livekit::options::{
     self, video as video_presets, PacketTrailerFeatures, TrackPublishOptions, VideoCodec,
-    VideoEncoding, VideoPreset,
+    VideoEncoderBackend, VideoEncoding, VideoPreset,
 };
 use livekit::prelude::*;
 use livekit::webrtc::video_frame::{FrameMetadata, I420Buffer, VideoFrame, VideoRotation};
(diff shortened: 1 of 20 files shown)


webrtc-sys/src/rtp_sender.cpp
```diff
@@ -90,4 +192,21 @@ void RtpSender::set_parameters(RtpParameters params) const {
     throw std::runtime_error(serialize_error(to_error(error)));
 }
 
+void RtpSender::set_video_encoder_backend(VideoEncoderBackend backend) const {
```
cloudwebrtc, line 195: Do we need a `bool video_encoder_backend_is_avaliable(VideoEncoderBackend backend);` or change this method to `bool set_video_encoder_backend(...);` to return whether the backend settings were successful?

cloudwebrtc: Alternatively, you can add `Vec< VideoEncoderBackend > video_encoder_backend_list();` to return the currently registered factories.

chenosaurus: I added video_encoder_backend_list() so user can get available list before requesting.  The request will still fallback to sw if the requested encoder is not available.

livekit/src/room/options.rs
```diff
@@ -407,6 +417,16 @@ pub fn video_layers_from_encodings(
 
 const VIDEO_RIDS: &[char] = &['q', 'h', 'f'];
 
+#[cfg(test)]
+mod tests {
```
ladvoc: nitpick: Test module usually goes at the end of the file.

libwebrtc/src/rtp_sender.rs
```diff
@@ -19,6 +19,41 @@ use crate::{
...
+/// capability checks.
+///
+/// ```
+/// use libwebrtc::rtp_sender::{video_encoder_backend_list, VideoEncoderBackend};
+///
+/// let backends = video_encoder_backend_list();
+/// assert!(backends.contains(&VideoEncoderBackend::Auto));
+/// assert!(backends.contains(&VideoEncoderBackend::Software));
+/// ```
+pub fn video_encoder_backend_list() -> Vec<VideoEncoderBackend> {
```
ladvoc: suggestion: To avoid an extra import, maybe this should be defined as an associated function on `VideoEncoderBackend`:
```rust
impl VideoEncoderBackend {
    pub fn list_available() -> Vec<Self> { 
        imp_rs::video_encoder_backend_list()
    }
}
```

Also, since this will often be iterated and not stored, I would recommend changing the signature to return a `impl Iterator<Item = VideoEncoderBackend>` rather than `Vec<VideoEncoderBackend>`. This will also make the API consistent with the query APIs for platform audio that were recently added which are iterator based.

examples/local_video/src/publisher.rs
```diff
@@ -60,19 +60,72 @@ impl From<PublisherCodec> for VideoCodec {
...
 }
 
+#[derive(Copy, Clone, Debug, ValueEnum)]
+enum PublisherEncoder {
+    Auto,
+    Software,
+    Hardware,
+    Nvenc,
+    Vaapi,
+    #[value(alias = "videotoolbox")]
```
stephen-derosa: FMU why is this alias needed if this is new functionality?

stephen-derosa (approved):
looks good to me!

ladvoc (approved):
LGTM ✅, a few small suggestions

libwebrtc/src/rtp_sender.rs
ladvoc: Example of this pattern: https://github.com/livekit/rust-sdks/blob/2434b45c83ae7a9e4324cb57c249a0ca62cb6379/livekit/src/platform_audio/mod.rs#L595

ladvoc:
One more thing: should this PR remove `LIVEKIT_PREFERRED_HW_ENCODER`?

examples/local_video/src/publisher.rs
chenosaurus: since the enum is VideoToolbox, the converts the str form to  kebab-case `video-toolbox`, thus needed an alias to just have `videotoolbox` be valid.

chenosaurus:
> One more thing: should this PR remove `LIVEKIT_PREFERRED_HW_ENCODER`?

good call, removed the use of LIVEKIT_PREFERRED_HW_ENCODER

cloudwebrtc (approved):
lgtm

PR #1127, opened 2026-06-02
Add max round trip latency

1egoman:
Adds a new `max_round_trip_latency` field to `PerformRpcData`, adding similar behavior as was added to the corresponding [swift](https://github.com/livekit/client-sdk-swift/pull/1023) and [android](https://github.com/livekit/client-sdk-android/pull/953) pull requests.

Note that this is a **breaking api change**. Per @ladvoc's guidance, this was deemed to be ok because there isn't really a great way to add this new field to the `PerformRpcData` struct in a backwards compatible way.

Given this is a breaking change, I opted to break it in a way where if we have to add _more_ fields in here in the future, it could be done in a non breaking way by making `PerformRpcData` use a builder pattern. That results in something that looks like:
```rust
let output = caller_room.local_participant().perform_rpc(
    PerformRpcData::new("identity", "method") // <-- New builder
        .with_payload("payload")
        .with_response_timeout(Duration::from_secs(5))
).await?;
```

### DIFF
--- a/.changeset/add_max_round_trip_latency.md
+++ b/.changeset/add_max_round_trip_latency.md
@@ -1,0 +1,6 @@
+---
+livekit: patch
+livekit-ffi: patch
+---
+
+Add rpc max_round_trip_latency and move to builder pattern - #1127 (@1egoman)
--- a/.changeset/add_max_round_trip_latency.md
+++ b/.changeset/add_max_round_trip_latency.md
@@ -1,5 +1,5 @@
 ---
-livekit: patch
+livekit: minor
 livekit-ffi: patch
 ---
 
--- a/examples/rpc/src/main.rs
+++ b/examples/rpc/src/main.rs
@@ -153,11 +153,10 @@ async fn register_receiver_methods(greeters_room: Arc<Room>, math_genius_room: A
 
                 match math_genius_room
                     .local_participant()
-                    .perform_rpc(PerformRpcData::new(
-                        data.caller_identity.clone(),
-                        "provide-intermediate",
-                        json!({"original": number}).to_string(),
-                    ))
+                    .perform_rpc(
+                        PerformRpcData::new(data.caller_identity.clone(), "provide-intermediate")
+                            .with_payload(json!({"original": number}).to_string()),
+                    )
                     .await
                 {
                     Ok(intermediate_response) => {
@@ -188,7 +187,7 @@ async fn perform_greeting(room: &Arc<Room>) -> Result<(), Box<dyn std::error::Er
     println!("[{}] Letting the greeter know that I've arrived", elapsed_time());
     match room
         .local_participant()
-        .perform_rpc(PerformRpcData::new("greeter", "arrival", "Hello"))
+        .perform_rpc(PerformRpcData::new("greeter", "arrival").with_payload("Hello"))
         .await
     {
         Ok(response) => {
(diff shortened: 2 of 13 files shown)


examples/rpc/src/main.rs
```diff
@@ -232,12 +225,11 @@ async fn perform_quantum_hypergeometric_series(
...
-            destination_identity: "math-genius".to_string(),
-            method: "quantum-hypergeometric-series".to_string(),
-            payload: json!({"number": 42}).to_string(),
-            ..Default::default()
-        })
+        .perform_rpc(PerformRpcData::new(
+            "math-genius",
+            "quantum-hypergeometric-series",
+            json!({"number": 42}).to_string(),
+        ))
```
1egoman: I think generally the only thing I am not that convinced of here is this `PerformRpcData::new` signature, given it takes positional params which makes it somewhat of a downsgrade from how it worked before.

However, I don't really like `PerformRpcData::default().with_destination_identity(...).with_method(...).with_payload(...)` better because it would be really easy to forget one method and end up with a `PerformRpcData::default()` which contains invalid data for something like `destination_identity` (this is how it works today, unfortunately, so I can't "fix this" and make it no longer implement `Default`).

livekit/src/room/rpc/mod.rs
```diff
@@ -107,10 +107,58 @@ impl RpcTransport for SessionTransport {
...
+    payload: String,
+    response_timeout: Duration,
+    max_round_trip_latency: Duration,
+}
+
+impl PerformRpcData {
+    pub fn new(
+        destination_identity: impl Into<ParticipantIdentity>,
+        method: impl Into<String>,
+        payload: impl Into<String>,
```
ladvoc: suggestion: It is reasonable that some RPCs don't accept any arguments, so I think we could drop the payload argument and add `with_payload`. I generally don't love constructors with more than one argument, but at the same time I don't have a better suggestion (`destination_identity` and `method` should always be set).

ladvoc (approved):

examples/rpc/src/main.rs
lukasIO: the examples so far show a `..Default()` pattern which, if we keep the struct as it was before, would not break for users, I think?

It would still be a breaking change on paper, but less likely to cause any actual problems when updating?

.changeset/add_max_round_trip_latency.md
```diff
@@ -0,0 +1,6 @@
+---
+livekit: patch
```
lukasIO:
```suggestion
livekit: minor
```

1egoman: Oops, yes, good catch

examples/rpc/src/main.rs
1egoman: I made the struct fields no longer public, which if I were to undo that change, then yes I think if users were to use that exact pattern it would be "backwards compatible". But just narrowly for that one particular call pattern. Maybe that's better than nothing, though.

1egoman: Yea, I think this makes sense, fixed in bb44a823

livekit/src/room/rpc/mod.rs
1egoman: I haven't come across a RPC without a payload in the wild. but I could see it being useful for purely a "signal" type message. And `String::default()` isn't a "broken" payload for a rpc method, so sounds good, did this in 9d456fee and e5f88baf.

examples/rpc/src/main.rs
lukasIO: cool, but you'd still opt to keep the `PerformRpcData::new` signature regardless?

1egoman: > cool, but you'd still opt to keep the PerformRpcData::new signature regardless?

Yep, the plan would be to update the docs to use the builder pattern, but keep the fields public so `PerformRpcData` with `..Default::default()` still works.

lukasIO (approved):

1egoman:
@ladvoc Before merging, any preference on `with_payload(payload: impl Into<String>)` versus `with_payload(payload: impl ToString)`?

The benefit of the latter is something like a serde value just works (ie, `with_payload(json!({ hello: "world" }))`), but the con is there's an extra copy involved.

ladvoc:
> The benefit of the latter is something like a serde value just works (ie, `with_payload(json!({ hello: "world" }))`), but the con is there's an extra copy involved.

I would go with `impl Into<String>` to avoid the extra copy and to maintain consistency with other APIs.

PR #1128, opened 2026-06-02
#[allow(dead_code)] for dead functions

stephen-derosa:
### PR description
`#[allow(dead_code)]` for `room/mod.rs` dead functions

### Testing
This enables passing the c++ sdk windows builds

### DIFF
--- a/.changeset/allow_dead_code_for_passing_windows_build.md
+++ b/.changeset/allow_dead_code_for_passing_windows_build.md
@@ -1,0 +1,6 @@
+---
+livekit: patch
+livekit-ffi: patch
+---
+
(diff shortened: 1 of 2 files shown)


stephen-derosa:
@ladvoc  we started seeing failing windows builds when trying to update to the latest rust sdk because the functions in this MR are dead code and didnt have #[allow(dead_code)] . It looks like theres president on using this marker in the rust sdk, but if this is the wrong approach lmk -- there should be some stuff we can do cpp to get by this

ladvoc (approved):
Good to add this anyway, but I don't think it is the reason for the build failure you are seeing.

PR #1133, opened 2026-06-03
Improve render latency on local video example

chenosaurus:
- Improve local_video example subscriber render performance by enabling zero-copy path from hw video decoder to GPU when available.

### DIFF
--- a/examples/local_video/src/publisher.rs
+++ b/examples/local_video/src/publisher.rs
@@ -3,7 +3,7 @@ use clap::{Parser, ValueEnum};
 use livekit::e2ee::{key_provider::*, E2eeOptions, EncryptionType};
 use livekit::options::{
     self, video as video_presets, PacketTrailerFeatures, TrackPublishOptions, VideoCodec,
-    VideoEncoding, VideoPreset,
+    VideoEncoderBackend, VideoEncoding, VideoPreset,
 };
 use livekit::prelude::*;
 use livekit::webrtc::video_frame::{FrameMetadata, I420Buffer, VideoFrame, VideoRotation};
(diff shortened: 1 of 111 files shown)


examples/local_video/src/publisher.rs
```diff
@@ -213,6 +213,72 @@ fn unix_time_us_now() -> u64 {
     SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_micros() as u64
 }
 
+const MAX_BACKEND_CAPTURE_TIMESTAMP_AGE_US: u64 = 5_000_000;
```
ladvoc, line 216: question: Where does this value come from? Does it need to be set differently depending on which backend is being used?

examples/local_video/src/viewport_aspect.rs
```diff
@@ -1,48 +1,98 @@
...
+use eframe::{egui, wgpu, Renderer};
+use egui_wgpu as egui_wgpu_backend;
+use std::time::Duration;
 
 const DEFAULT_ASPECT: f32 = 16.0 / 9.0;
-const INITIAL_LONG_EDGE: f32 = 960.0;
+const DEFAULT_INITIAL_LONG_EDGE: f32 = 960.0;
 const MIN_LONG_EDGE: f32 = 320.0;
+/// Repaint cadence used while a local video window is visible.
+pub(crate) const VIDEO_REPAINT_INTERVAL: Duration = Duration::from_millis(8);
```
ladvoc, line 9: question: This was previously 16ms; was changing to 8 intentional?

ladvoc (approved):

examples/local_video/src/viewport_aspect.rs
ladvoc: I think you can just do `ctx.request_repaint()` to repaint at the monitor's native refresh rate.

examples/local_video/src/publisher.rs
chenosaurus: it's used in a sanity check to ensure the sensor provided capture timestamp is close to system time.  It's a very generous buffer to catch timestamps that are way out of line.

examples/local_video/src/viewport_aspect.rs
chenosaurus: This is a cap to ensure we don't try to paint faster than 125fps since vsync is disabled.  Previously it was at 16ms, which didn't max out the refresh rate of macbook displays.

PR #1136, opened 2026-06-05
adding ability to control local video track encode parameters

chenosaurus:
- Add hooks into local video track to set encoding limits
- Add to local_video subscriber example to demonstrate controlling encoding parameters on publisher


examples/local_video/src/subscriber.rs
```diff
@@ -409,6 +419,181 @@ struct Args {
     e2ee_key: Option<String>,
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+struct SetEncodingLimitsRequest {
```
ladvoc: suggestion: Consider putting common structs in a module and share between the two binaries.

ladvoc (approved):

cloudwebrtc (approved):

libwebrtc/src/native/rtp_sender.rs
```diff
@@ -76,8 +76,26 @@ impl RtpSender {
...
+            native_encoding.scale_resolution_down_by =
+                encoding.scale_resolution_down_by.unwrap_or_default();
+            native_encoding.has_scalability_mode = encoding.scalability_mode.is_some();
+            native_encoding.scalability_mode = encoding.scalability_mode.unwrap_or_default();
+        }
+
         self.sys_handle
-            .set_parameters(parameters.into())
+            .set_parameters(native_parameters)
             .map_err(|e| unsafe { sys_err::ffi::RtcError::from(e.what()).into() })
```
xianshijing-lk, line 110: what do these params mean for simulcast, adaptive streaming, and dynacast ?

examples/local_video/src/publisher.rs
```diff
@@ -214,6 +215,24 @@ fn unix_time_us_now() -> u64 {
 }
 
 const MAX_BACKEND_CAPTURE_TIMESTAMP_AGE_US: u64 = 5_000_000;
+const SET_VIDEO_ENCODING_LIMITS_METHOD: &str = "set-video-encoding-limits";
+
+#[derive(Debug, Deserialize, Serialize)]
+struct SetEncodingLimitsRequest {
```
xianshijing-lk: is this designed to cap this track overall ? or we should consider per encoding / layer limits due to simulcast ?

xianshijing-lk (commented):
two questions

examples/local_video/src/publisher.rs
chenosaurus: This still need a bit of work to support simulcast as currently it would apply the same limits to each layer.

libwebrtc/src/native/rtp_sender.rs
chenosaurus: added support for setting encoding config for individual simulcast layers.  This will not have any effect on adaptive streaming and dynacast as they only dynamically turn on/off individual layers to be streamed based on subscriber demand.

examples/local_video/src/publisher.rs
chenosaurus: @xianshijing-lk after thinking about simulcast more, it only makes sense for the encoding limits to control the high layer and for the publisher to automatically generate the medium and low layers.

PR #1137, opened 2026-06-05
Reject oversized data messages before they break the data channel

cnderrauber:
Sending a data message larger than the negotiated SCTP max message size made libwebrtc abruptly close the publisher data channel. Because DataChannelInterface::Send returns true regardless, publish_data() returned Ok while the channel was silently torn down; every subsequent publish then failed with "could not establish publisher connection: timeout" (15s each) since the channel is never recreated once has_published is set.

- Parse the negotiated max message size from the publisher answer SDP (RFC 8841 a=max-message-size), defaulting to 65535 when absent.
- Reject packets whose encoded size exceeds it in publish_data(), returning an error to the caller instead of silently aborting the channel. A value of 0 means "no limit".
- Log an info when a publisher data channel closes while the session and peer connection are still up (the signature of this failure), without attempting renegotiation.

### DIFF
--- a/livekit/src/rtc_engine/rtc_session.rs
+++ b/livekit/src/rtc_engine/rtc_session.rs
@@ -72,6 +72,10 @@ pub const RELIABLE_RECEIVED_STATE_TTL: Duration = Duration::from_secs(30);
 pub const PUBLISHER_NEGOTIATION_FREQUENCY: Duration = Duration::from_millis(150);
 pub const INITIAL_BUFFERED_AMOUNT_LOW_THRESHOLD: u64 = 2 * 1024 * 1024;
 
+/// Default data-channel max message size (bytes), used when the remote SDP
+/// answer does not advertise an `a=max-message-size` attribute (RFC 8841).
+pub const DEFAULT_MAX_MESSAGE_SIZE: u64 = 65535;
+
 /// Buffered-amount low threshold for the `_data_track` DC.
 ///
 /// Kept small (vs. the 2 MiB default for reliable/lossy) so we hand at most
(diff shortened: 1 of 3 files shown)


livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -612,6 +622,28 @@ impl RtcSession {
...
+            let weak_inner = Arc::downgrade(&inner);
+            dc.on_state_change(Some(Box::new(move |state| {
+                if state != DataChannelState::Closed {
+                    return;
+                }
+                let Some(inner) = weak_inner.upgrade() else {
+                    return;
+                };
+                if !inner.closed.load(Ordering::Acquire) && inner.publisher_pc.is_connected() {
+                    log::info!(
```
ladvoc: suggestion: This should probably be logged at error level.

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -612,6 +622,28 @@ impl RtcSession {
             pc_state_notify: Notify::new(),
         });
 
+        // Log when a publisher data channel closes without the engine or peer
+        // connection tearing it down 
+        for (dc, label) in
+            [(&inner.reliable_dc, RELIABLE_DC_LABEL), (&inner.lossy_dc, LOSSY_DC_LABEL)]
```
ladvoc: suggestion: For completeness, this should probably also track the data track DC.

xianshijing-lk:
@1egoman , could you please review this PR ?

1egoman (approved):
Makes sense to me!

PR #1144, opened 2026-06-08
Remove default link-time dependency on GLib

### LINKED ISSUE #1129: webrtc-sys: pkg_config::probe_library for glib/gobject/gio adds unused -l flags, leaks as DT_NEEDED in downstream binaries
## Summary

`webrtc-sys/build.rs` calls `pkg_config::probe_library` for `glib-2.0`, `gobject-2.0`, and `gio-2.0` on Linux. This ultimately results in `-lglib-2.0 -lgobject-2.0 -lgio-2.0` in the final link command. But glib is merely a compile-time header dependency here.

We noticed this issue with a cdylib that we built against webrtc-sys 0.3.32 on aarch64-unknown-linux-gnu (with a github ubuntu-22.04-arm runner). On that runner, the linker doesn't apply `--as-needed` by default, and so we end up with [leaked DT_NEEDED entries in our cdylib](https://github.com/foxglove/foxglove-sdk/issues/1291), even though the there are no undefined g_* symbols. This isn't a problem on x86_64, because the github runner there apparently does apply `--as-needed` by default.

## Proposed Fix

https://github.com/livekit/rust-sdks/blob/fe566a24f5ec363f55a3cc5f586ee7e9242d65a5/webrtc-sys/build.rs#L177-L179

Switch the glib pkg-config calls from probe_library to a header-only probe:

```rust
for lib_name in ["glib-2.0", "gobject-2.0", "gio-2.0"] {
    let lib = pkg_config::Config::new()
        .cargo_metadata(false)
        .probe(lib_name)
        .unwrap();
    for path in lib.include_paths {
        builder.include(path);
    }
}
```

gasmith:
### Before you submit your PR

### PR Description

The webrtc-sys build script used pkg_config::probe_library for glib-2.0/gobject-2.0/gio-2.0, which emitted cargo:rustc-link-lib directives even though only the headers are required at build time. Switch to pkg_config::Config::new().cargo_metadata(false) so the include paths are still discovered but no link metadata is emitted.

Drop glib-main-loop from libwebrtc's default features and re-expose it as an opt-in passthrough on the top-level livekit crate. The screensharing example enables it explicitly. Note that the original change (https://github.com/livekit/rust-sdks/commit/ca86d4c5d7f05c3fbd402e112f466575489dbe6f) was somewhat broken on arrival, since it did not provide any way for a `livekit` consumer to opt out of the internal GLib main loop.

Fixes: https://github.com/livekit/rust-sdks/issues/1129

### Breaking changes

Consumers of `livekit` or `libwebrtc` that rely on Wayland screen sharing must now enable the `glib-main-loop` feature (if they do not have their own GLib main loop).

### MSRV

No change.

### Testing

Validated that a trivial crate that depends on `livekit` no longer pulls in `glib` as a dependency.

Built a cdylib on Linux aarch64 in a build environment where the linker does not use `--as-needed` (or it simply does not work as expected), and validated that the resulting binary does not include library dependencies on GLib. See the [build artifacts](https://github.com/foxglove/foxglove-sdk/actions/runs/27042873647) from https://github.com/foxglove/foxglove-sdk/pull/1292 (`lib/libfoxglove.so`).

### DIFF
--- a/.changeset/remove-default-glib-dependency.md
+++ b/.changeset/remove-default-glib-dependency.md
@@ -1,0 +1,11 @@
+---
+webrtc-sys: patch
+libwebrtc: minor
+livekit: minor
+---
+
+# Make GLib an opt-in dependency
+
+`webrtc-sys` no longer links against `glib-2.0`/`gobject-2.0`/`gio-2.0` by default.
+
+Breaking: Wayland screen sharing now requires the `glib-main-loop` feature on `livekit` (or `libwebrtc`).
--- a/examples/screensharing/Cargo.toml
+++ b/examples/screensharing/Cargo.toml
@@ -7,7 +7,7 @@ publish = false
 [dependencies]
 tokio = { workspace = true, features = ["full"] }
 env_logger = { workspace = true }
-livekit = { workspace = true, features = ["rustls-tls-native-roots"] }
+livekit = { workspace = true, features = ["glib-main-loop", "rustls-tls-native-roots"] }
 livekit-api = { workspace = true }
 log = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
--- a/libwebrtc/Cargo.toml
+++ b/libwebrtc/Cargo.toml
@@ -8,11 +8,12 @@ description = "Livekit safe bindings to libwebrtc"
 repository.workspace = true
 
 [features]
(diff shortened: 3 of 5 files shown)


cloudwebrtc (approved):
lgtm!

PR #1148, opened 2026-06-09
harden reconnect behaviour

lukasIO:
### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async


livekit/src/rtc_engine/mod.rs
```diff
@@ -53,6 +57,37 @@ pub(crate) type EngineResult<T> = Result<T, EngineError>;
 pub const RECONNECT_ATTEMPTS: u32 = 10;
 pub const RECONNECT_INTERVAL: Duration = Duration::from_secs(5);
 
+/// Exponential-backoff-with-full-jitter parameters for spacing reconnect
```
ladvoc: suggestion(non-blocking): Consider moving backoff related constants, functions, and unit tests into a separate module.

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -1848,6 +1790,21 @@ impl SessionInner {
             SimulateScenario::SignalReconnect => {
                 self.signal_client.close().await;
             }
+            SimulateScenario::DisconnectSignalOnResume => {
```
ladvoc, line 1911: suggestion: Simulate scenario is now exposed through _livekit-ffi_; this new case should also be exposed for completeness.

livekit/src/rtc_engine/mod.rs
```diff
@@ -868,7 +946,16 @@ impl EngineInner {
                 }
             }
 
-            self.reconnecting_interval.lock().await.tick().await;
+            // Exponential backoff with full jitter between attempts (DELTA 3).
+            // A server-requested reconnect signals retry_now_notify to collapse
+            // this wait so the next attempt fires immediately.
+            let backoff = reconnect_backoff_delay(i);
+            tokio::select! {
```
ladvoc, line 1015: issue: If the room is explicitly closed by the user, this task will not be properly cancelled. The addition of this `select!` brought my attention here, but this is a pre-existing issue. 

You will find there is a `close_notifier` referenced on L805 with a comment mentioning it needs to be used as a signal to cancel this task—however, this was wired up. Solution is to wire this up and use it as a branch in your `select!` in order to be able to break out of the loop early when the room is closed.

livekit/src/rtc_engine/rtc_session.rs
lukasIO: makes sense, added!

livekit/src/rtc_engine/mod.rs
lukasIO: good idea!

livekit/src/rtc_engine/mod.rs
lukasIO: great catch!

addressed in [e3320d3](https://github.com/livekit/rust-sdks/pull/1148/commits/e3320d3af251eb607bbcea1d4fec914cbd8bd5a4)

livekit/specs/signalling-reconnection.allium
lukasIO, line 1: wanted to call this out specifically. It's "just" a spec generated by https://github.com/juxt/allium, but in case someone is opposed to adding this in, let me know

stephen-derosa (commented):
is there any world where we'd want to apply this reconnet behavior to initial connection? Im thinking of an application facing the same sort of network instability that may cause a reconnect on initial connection.

livekit/specs/signalling-reconnection.allium
ladvoc: My only concern is whether any parts of the spec could be misleading to an agent reading it in the future. In those cases, relying on the code as the source of truth might help avoid confusion. That said, I don’t feel strongly about it.

ladvoc (approved):
LGTM ✅

lukasIO:
> is there any world where we'd want to apply this reconnet behavior to initial connection? Im thinking of an application facing the same sort of network instability that may cause a reconnect on initial connection.


we currently differentiate between initial connection and reconnection, but we still allow for re-trying on the initial connection (which would effectively be a "reconnect" just for a connection that had never been established in the first place), see https://github.com/livekit/rust-sdks/blob/129e5039d9578d5c5053a831e9c8ffc822303724/livekit/src/rtc_engine/mod.rs#L502-L503

livekit/specs/signalling-reconnection.allium
lukasIO: that's a fair concern. I think it's minimised as much as possible by having a special (allium specific) extension and allium itself regularly compares spec against code.

PR #1149, opened 2026-06-09
error early when not authenticated

lukasIO:
### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async


livekit/src/rtc_engine/mod.rs
```diff
@@ -1091,6 +1115,28 @@ fn leave_disconnect_reason(err: &EngineError) -> Option<DisconnectReason> {
...
+/// knows are rejected.
+///
+/// We key on `SignalError::Client(401|403)`, which is produced by the server's
+/// `rtc/validate` probe (see [`super`]'s `SignalInner::validate`) — an
+/// authoritative classification. We deliberately do NOT key on the raw
+/// `WsError::Http` upgrade status, because that can be a fabricated 401 masking a
+/// transient server error (e.g. a 503 from a saturated node), which IS
+/// retryable. A resume that hits a raw 401 simply escalates to a full reconnect,
+/// whose connect path runs `validate()` and surfaces the authoritative status.
+fn auth_failure_reason(err: &EngineError) -> Option<DisconnectReason> {
```
ladvoc, line 1197: suggestion(non-blocking): Alternatively could be defined as a method on the enum:
```rust
impl EngineError {
    fn auth_failure_reason(&self) -> Option<DisconnectReason> {
        // ...
    }
}
```

ladvoc (approved):

PR #1150, opened 2026-06-09
region cache

lukasIO:
### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async


livekit-api/src/signal_client/region.rs
```diff
@@ -12,13 +12,60 @@
...
+fn region_cache() -> &'static Mutex<HashMap<String, CachedRegions>> {
+    static CACHE: OnceLock<Mutex<HashMap<String, CachedRegions>>> = OnceLock::new();
+    CACHE.get_or_init(|| Mutex::new(HashMap::new()))
+}
+
+/// Cached region URLs for `host` if the entry is still within
+/// [`REGION_CACHE_TTL`], else `None` (the caller should re-fetch).
+fn cached_region_urls(host: &str) -> Option<Vec<String>> {
+    let cache = region_cache().lock();
+    cache.get(host).filter(|e| e.fetched_at.elapsed() < REGION_CACHE_TTL).map(|e| e.urls.clone())
```
ladvoc: question: Should this have a mechanism to delete expired entries? In practice, this probably won't accumulate too many entries since you typically would be using the same cloud host, but it might still be a good idea to add for good hygiene. See _livekit/src/utils/ttl_map.rs_ as a reference.

livekit-api/src/signal_client/region.rs
```diff
@@ -12,13 +12,60 @@
...
 
 use super::{SignalError, SignalResult, REGION_FETCH_TIMEOUT};
 
+/// How long a fetched region list is reused before being re-fetched. Matches
+/// client-sdk-js's `DEFAULT_MAX_AGE_MS`. (The server's `Cache-Control: max-age`
+/// is not yet honoured — a fixed TTL keeps this backend-agnostic; honouring the
+/// header is a possible refinement.)
+const REGION_CACHE_TTL: Duration = Duration::from_secs(5);
+
+struct CachedRegions {
```
ladvoc, line 31: suggestion(non-blocking): Consider encapsulating in a struct:
```rust

struct RegionCache {
    entries: Mutex<HashMap<String, CachedRegions>>,
    ttl: Duration,
}
impl RegionCache {
    const TTL: Duration = Duration::from_secs(5);

    fn shared() -> &'static RegionCache {
        static CACHE: OnceLock<Self> = OnceLock::new();
        CACHE.get_or_init(|| Self::new(Self::TTL))
    }

    fn new(ttl: Duration) -> Self {
        Self { entries: Mutex::new(HashMap::new()), ttl }
    }

    fn get(&self, host: &str) -> Option<Vec<String>> {
       // ...
    }

    fn insert(&self, host: &str, urls: Vec<String>) {
       // ...
    }
}
```

ladvoc (approved):

livekit-api/src/signal_client/region.rs
lukasIO: good point. I realised the region cache didn't allow for any stale usage of the most recent region available so I added it in. It will still refetch, but try with the latest available cached entry immediately.
This however means that expired entries can still provide value for a quick connection path.

I added an explicit `::clear()` method to clear the cache, however this isn't called today as it's tricky to decide _when_ to do it. If you have any suggestions let me know, otherwise I keep it exposed so that users have the option to clear it themselves if they have a need for that.

ladvoc: Makes sense, I think we can just keep the stub for now until there's a compelling reason to expose this in public API somehow. I would add `#[allow(dead_code)]` to the `::clear()` method to make it clear it isn't being used yet (and to silence the warning).

PR #1151, opened 2026-06-09
Fix transceiver reusing problem in single-pc mode 

cnderrauber:
Sfu could send empty transceiver to subscriber in single pc
mode when publisher unbpulishes track before answer sdp is
created. The empty transceiver can be reused in later subscription
that changes msid in renegotiation, libwebrtc will not fire
onTrack event in the msid changed case cause subscrption failure.
Chrome detects the msid change and synthsize onTrack event to handle
this case. This PR do the same thing as browser.

### DIFF
--- a/livekit/src/rtc_engine/rtc_session.rs
+++ b/livekit/src/rtc_engine/rtc_session.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     convert::TryInto,
     fmt::Debug,
     sync::{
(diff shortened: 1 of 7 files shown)


livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -943,6 +945,43 @@ impl RtcSession {
...
+            for stream in receiver.streams() {
+                let stream_id = stream.id();
+                current
+                    .entry((mid.clone(), stream_id))
+                    .or_insert_with(|| (stream, track.clone(), transceiver.clone()));
+            }
+        }
+
+        let mut dispatched = self.dispatched_streams.lock();
+        dispatched.retain(|key| current.contains_key(key));
```
xianshijing-lk, line 971: any chance that (mid, stream_id) disappears for one SDP update but later reappears again ? 

if this will happen, the current code will dispatch MediaTrack again. Is this intentional behavior ?

xianshijing-lk (commented):
one question

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -1487,11 +1530,17 @@ impl SessionInner {
...
                 if !streams.is_empty() {
-                    let _ = self.emitter.send(SessionEvent::MediaTrack {
-                        stream: streams.remove(0),
-                        track,
-                        transceiver,
-                    });
+                    let stream = streams.remove(0);
+                    let mid = transceiver.mid().unwrap_or_default();
+                    let already_dispatched = !mid.is_empty()
+                        && !self.dispatched_streams.lock().insert((mid, stream.id()));
```
xianshijing-lk, line 1536: just want to double check, at any condition should we clear this dispatched_streams so that we will dipspatch the tracks again ?

livekit/src/rtc_engine/rtc_session.rs
cnderrauber: It is intentional, in case like:
1. client subscribe, sfu sdp: mid:1, streamid_id: stream_A
2. client unsubscribe, sfu sdp: mid:1, no stream
3. client resubscribe after awhile, sfu reuse mid:1 to send track and client dispatch the track again.

cnderrauber: > just want to double check, at any condition should we clear this dispatched_streams so that we will dipspatch the tracks again ?

`dispatched.retain(|key| current.contains_key(key));` here will also remove all keys don't exist in current streams, so we can dispatch it again when the track has been resubscribed.

xianshijing-lk (approved):

PR #1155, opened 2026-06-10
return DeviceNotFound when device is not there for set_recording_devi…

xianshijing-lk:
…ce() / set_playout_device()

### Before you submit your PR

### PR description

Return  return DeviceNotFound when device is not there for set_recording_device() and set_playout_device()

Otherwise, the API will behave differently on different platforms.

### Breaking changes

### MSRV

N/A

### Testing

CI

### Async

N/A

### DIFF
--- a/livekit/src/platform_audio/mod.rs
+++ b/livekit/src/platform_audio/mod.rs
@@ -657,6 +657,11 @@ impl PlatformAudio {
     ///
     /// * `id` - Device identifier from [`RecordingDeviceInfo::id`]
     ///
+    /// # Errors
+    ///
+    /// Returns [`AudioError::DeviceNotFound`] if the device ID does not exist
+    /// in the current list of available recording devices.
+    ///
     /// # Example
     ///
     /// ```rust,ignore
@@ -668,6 +673,13 @@ impl PlatformAudio {
     /// }
     /// ```
     pub fn set_recording_device(&self, id: &RecordingDeviceId) -> AudioResult<()> {
+        // Validate that the device ID exists in the current device list
+        // Skip validation on mobile where device selection is a no-op
+        #[cfg(not(any(target_os = "ios", target_os = "android")))]
+        if !self.recording_devices().any(|d| &d.id == id) {
+            return Err(AudioError::DeviceNotFound);
+        }
+
         if self.handle.runtime.set_recording_device_by_guid(id.as_str()) {
             Ok(())
         } else {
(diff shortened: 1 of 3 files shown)


livekit/src/platform_audio/mod.rs
```diff
@@ -704,6 +721,13 @@ impl PlatformAudio {
     /// }
     /// ```
     pub fn set_playout_device(&self, id: &PlayoutDeviceId) -> AudioResult<()> {
+        // Validate that the device ID exists in the current device list
+        // Skip validation on mobile where device selection is a no-op
+        #[cfg(not(any(target_os = "ios", target_os = "android")))]
+        if !self.playout_devices().any(|d| &d.id == id) {
+            return Err(AudioError::DeviceNotFound);
+        }
```
jhugman: This looks fine, from a Rust POV. I might suggest a helper method that shares the logic with the above, but this is minor.

Does similar logic need to exist in the `switch_playout_device()` function too?

livekit/src/platform_audio/mod.rs
```diff
@@ -668,6 +673,13 @@ impl PlatformAudio {
     /// }
     /// ```
     pub fn set_recording_device(&self, id: &RecordingDeviceId) -> AudioResult<()> {
+        // Validate that the device ID exists in the current device list
+        // Skip validation on mobile where device selection is a no-op
+        #[cfg(not(any(target_os = "ios", target_os = "android")))]
+        if !self.recording_devices().any(|d| &d.id == id) {
+            return Err(AudioError::DeviceNotFound);
+        }
```
jhugman: Does this need to be called from anywhere else, e.g. `switch_recording_device()`?

jhugman (approved):
✅ with a couple of questions for my understanding.

livekit/src/platform_audio/mod.rs
xianshijing-lk: I intentionally removed those calls from witch_recording_device() to avoid causing it to fail unnecessarily. witch_recording_device() is invoked when devices are unplugged, and in those cases it will silently fall back to the default device.

livekit/src/platform_audio/mod.rs
xianshijing-lk: Good idea, I can make a helper function. 

No, this logic is not needed by switch_playout_device() function, Claude added it there, and I intentionally removed it since I want switch_playout_device() to silently fall back to the default device.

PR #1159, opened 2026-06-10
Data track schema metadata

ladvoc:
Adds support for associating schema metadata with a published data track and storing/retrieving schema definitions.

Schema storage is built on-top of data blobs, a general purpose mechanism for storing large (in the order of KBs), arbitrary data blobs in a room:
- Protocol: https://github.com/livekit/protocol/pull/1595
- SFU implementation: https://github.com/livekit/livekit/tree/raja_async_attributes

Protocol additions for schema metadata:
- https://github.com/livekit/protocol/pull/1553

Closes BOT-368


ladvoc:
For local testing:
- Use this SFU branch: https://github.com/livekit/livekit/tree/ladvoc/schema-metadata
- Build and run with data blobs support enabled: `LIVEKIT_CONFIG="enable_participant_data_blob: true" ./bin/livekit-server --dev`

livekit/src/room/participant/local_participant.rs
```diff
@@ -917,6 +918,144 @@ impl LocalParticipant {
...
+            RoomError::Internal(format!("schema definition is not valid UTF-8: {err}"))
+        })?;
+        Ok(definition)
+    }
+
+    // TODO: unify request/response logic, timeout behavior across SDK.
+    const DATA_BLOB_REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+
+    /// Stores an arbitrary blob of data on the server, keyed by `key`.
+    async fn store_data_blob(&self, key: proto::DataBlobKey, contents: Bytes) -> EngineResult<()> {
```
1egoman: question: Should this be `RoomResult<()>`, not `EngineResult<()>`? There's a few other instances of this in the file as well.

livekit/src/room/participant/local_participant.rs
```diff
@@ -917,6 +918,144 @@ impl LocalParticipant {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `id` - Identifies the schema; the same ID is provided when publishing a
+    ///   data track that uses it.
+    /// * `definition` - The schema definition, stored as-is. It is neither parsed
+    ///   nor validated against its [encoding](DataTrackSchemaId::encoding), so
+    ///   the caller is responsible for ensuring it is well-formed.
+    ///
+    pub async fn define_schema(&self, id: DataTrackSchemaId, definition: String) -> RoomResult<()> {
```
1egoman, line 1052: thought: Are you confident that schemas will / should always be text? Or should `definition: String` be some sort of bytes object, like `Vec<u8>` or `bytes::Bytes` (or maybe `impl Into<bytes::Bytes>`)?

livekit/tests/data_track_test.rs
```diff
@@ -160,6 +160,35 @@ async fn test_publish_duplicate_name() -> Result<()> {
...
+#[cfg(feature = "__lk-e2e-test")]
+#[test_log::test(tokio::test)]
+async fn test_publish_with_schema_and_frame_encoding() -> Result<()> {
+    use livekit::data_track::{DataTrackFrameEncoding, DataTrackSchemaEncoding, DataTrackSchemaId};
+
+    let mut rooms = test_rooms(2).await?;
+    let (pub_room, _) = rooms.pop().unwrap();
+    let (_, mut sub_room_event_rx) = rooms.pop().unwrap();
+
+    let schema_id = DataTrackSchemaId::new("my_schema", DataTrackSchemaEncoding::JsonSchema);
```
1egoman: question: Is this `DataTrackSchemaId::new(...)` expected to be a typical pattern, maybe in the case that you know that a schema already exists on the SFU end? Or is the idea that in practice, a user would always call `LocalParticipant::deine_schema` and this is just being done this way for testing?

1egoman (commented):
Generally makes sense to me!

I think `DataTrackSchemaId` being a non primative value is key here to future interface evolution (like adding a generic type with a default which when set explictly can be used to assert the type of payloads later on).

livekit-ffi/protocol/data_track.proto
```diff
@@ -246,3 +254,72 @@ message DataTrackStreamEOS {
   // Absent if the stream ended normally (i.e., due to the track being unpublished).
   optional SubscribeDataTrackError error = 1;
 }
+
+// MARK: - Schemas
```
stephen-derosa, line 259: FMU, what is this `// MARK:` notation?

livekit-datatrack/src/local/proto.rs
```diff
@@ -169,9 +206,36 @@ mod tests {
...
+        assert_eq!(
+            info.schema,
+            Some(DataTrackSchemaId::new("schema", DataTrackSchemaEncoding::JsonSchema))
+        );
+        assert_eq!(info.frame_encoding, Some(DataTrackFrameEncoding::Json));
         assert!(info.uses_e2ee);
     }
 
+    #[test]
+    fn test_frame_encoding_mapping() {
```
stephen-derosa, line 214: `test_empty_frame_encoding` ?

livekit/src/room/participant/local_participant.rs
```diff
@@ -917,6 +918,144 @@ impl LocalParticipant {
     pub fn update_data_encryption_status(&self, _is_encrypted: bool) {
         // Local participants don't receive data messages, so this is a no-op
     }
+
+    /// Stores the definition of a data track schema.
+    ///
+    /// Called by a publisher to make a schema available to subscribers, who can
+    /// later look up its definition via [`get_schema`](Self::get_schema). Define a
+    /// schema before publishing any data track that references it, so that
```
stephen-derosa, line 1038: what is the behavior if a data track with a schema is published before the schema itself is published?

stephen-derosa (commented):
in general looks good, need to give it a deeper dive

livekit-ffi/protocol/data_track.proto
ladvoc: This is recognized by most editors (see the minimap and navigator in Cursor/Vscode).

livekit/src/room/participant/local_participant.rs
ladvoc: The track will be published normally and carry the schema ID. If you care about retrieving it on the subscriber end, it is recommended to store the schema before publishing the track to ensure it is available right away, but if you don't need the actual definition for you application, you can simply use schema ID as an identifier.

livekit/src/room/participant/local_participant.rs
ladvoc: Yes, while the wire format (frame content) will often be binary, I can't think of any case where the schema definition language would be binary. @stephen-derosa, wdyt?

.github/workflows/tests.yml
```diff
@@ -151,7 +151,9 @@ jobs:
         uses: livekit/dev-server-action@61e2b4dcb170dd3591e0c9b0db3c3fe5db93b500
         with:
           github-token: ${{ github.token }}
-
+          config: |
+            # TODO: Remove once this is enabled by default.
+            enable_participant_data_blob: true
```
alan-george-lk, line 156: For my understanding, when do we decide to flip this on by default? Is convention to ship things off by default then wait a few versions to make sure it's stable?

livekit-datatrack/src/local/proto.rs
```diff
@@ -33,7 +33,15 @@ impl From<SfuPublishRequest> for proto::PublishDataTrackRequest {
     fn from(event: SfuPublishRequest) -> Self {
         use proto::encryption::Type;
         let encryption = if event.uses_e2ee { Type::Gcm } else { Type::None }.into();
-        Self { pub_handle: event.handle.into(), name: event.name, encryption }
+        let schema = event.schema.map(|schema| schema.into());
+        let frame_encoding = event.frame_encoding.map(Into::into);
```
alan-george-lk: For my Rust learning, what's the practical need for these lines to be different? I.e. why the `|schema| schema.into()` vs `Into::into`

livekit-datatrack/src/schema.rs
```diff
@@ -0,0 +1,281 @@
...
+    ///
+    /// [`Json`]: DataTrackFrameEncoding::Json
+    JsonSchema,
+
+    /// Another well-known encoding not known to this client version.
+    Other,
+    /// An application-specific encoding identified by the contained string.
+    ///
+    /// Prefer using one of the well-known encodings unless the format is not enumerated.
+    /// The identifier must be non-empty and no longer than 25 characters.
```
alan-george-lk: Could be my rust knowledge, but I am not seeing the 25 limit enforced anywhere?

alan-george-lk (approved):

livekit-datatrack/src/schema.rs
ladvoc: It is enforced by the SFU. The reason I opted not to do client side validation is because the limit can be changed through the SFU config (which might happen for OSS deployments).

.github/workflows/tests.yml
ladvoc: Yes, this is what we did for data tracks as well; OSS didn't initially enable it by default.

livekit-datatrack/src/local/proto.rs
ladvoc: `Into::into` is the function-item form of calling `.into()`, so here it is equivalent to `|a| a.into()`. It works because `Option::map` provides the input type, and the surrounding assignment/context lets Rust infer the same output type. Addressed in [b2cde31](https://github.com/livekit/rust-sdks/pull/1159/commits/b2cde3138e9b6a05931e8e86dfae0550f8500eb8)

livekit-datatrack/src/local/manager.rs
```diff
@@ -108,6 +108,14 @@ impl Manager {
     }
 
     async fn on_publish_request(&mut self, event: PublishRequest) {
+        if let Err(error) = crate::schema::validate_schema(
+            event.options.frame_encoding.as_ref(),
+            event.options.schema.as_ref().map(|id| id.encoding()),
+        ) {
+            _ = event.result_tx.send(Err(PublishError::InvalidSchema(error)));
```
alan-george-lk, line 115: Maybe this happens elsewhere, but some sort of incompatibility error log would be nice

livekit-datatrack/src/schema.rs
alan-george-lk: Might be misleading to hard-comment the 25 char limit in the doc comment, might be worth adjusting to clarify the configurability

alan-george-lk (approved):

livekit-datatrack/src/local/manager.rs
ladvoc: This is surfaced at the call site when publishing a data track (also for FFI).

livekit-datatrack/src/schema.rs
ladvoc: Good call, addressed in [f0e4663](https://github.com/livekit/rust-sdks/pull/1159/commits/f0e466376c5309d0709e047c0105b31547fbf392).

PR #1161, opened 2026-06-11
Add support for Nvidia Jetson V4L2 M2M video encoder

chenosaurus:
- Add support for Nvidia Jetson V4L2 M2M video encoder
- Support H264, H265, AV1 encoding
- Update the local_video example with --source flag that can be `uvc` or `argus`

### DIFF
--- a/examples/local_video/Cargo.toml
+++ b/examples/local_video/Cargo.toml
@@ -62,4 +62,4 @@ nokhwa = { git = "https://github.com/l1npengtul/nokhwa", rev = "4923ecab7cf26f9d
 nokhwa = { git = "https://github.com/l1npengtul/nokhwa", rev = "4923ecab7cf26f9dba83867a15a9d8662d021296", default-features = false, features = ["input-msmf"] }
 
 [build-dependencies]
-cc = "1"
+cc = { workspace = true }
--- a/examples/local_video/Cargo.toml
+++ b/examples/local_video/Cargo.toml
@@ -22,12 +22,15 @@ name = "list_devices"
 path = "src/list_devices.rs"
 required-features = ["desktop"]
 
+[[bin]]
+name = "clock"
+path = "src/clock.rs"
+required-features = ["desktop"]
+
 [dependencies]
 tokio = { workspace = true, features = ["full", "parking_lot"] }
 tokio-stream = { workspace = true }
 livekit = { workspace = true, features = ["rustls-tls-native-roots"] }
-webrtc-sys = { workspace = true }
-libwebrtc = { workspace = true }
 livekit-api = { workspace = true }
 yuv-sys = { workspace = true, features = ["jpeg"] }
 futures = { workspace = true }
--- a/examples/local_video/Cargo.toml
+++ b/examples/local_video/Cargo.toml
@@ -24,6 +24,7 @@ required-features = ["desktop"]
 
 [dependencies]
 tokio = { workspace = true, features = ["full", "parking_lot"] }
+tokio-stream = { workspace = true }
 livekit = { workspace = true, features = ["rustls-tls-native-roots"] }
 webrtc-sys = { workspace = true }
 libwebrtc = { workspace = true }
--- a/examples/local_video/Cargo.toml
+++ b/examples/local_video/Cargo.toml
@@ -57,3 +57,6 @@ nokhwa = { git = "https://github.com/l1npengtul/nokhwa", rev = "4923ecab7cf26f9d
 
 [target.'cfg(target_os = "windows")'.dependencies]
 nokhwa = { git = "https://github.com/l1npengtul/nokhwa", rev = "4923ecab7cf26f9dba83867a15a9d8662d021296", default-features = false, features = ["input-msmf"] }
+
+[build-dependencies]
+cc = "1"
--- a/examples/local_video/src/argus.rs
+++ b/examples/local_video/src/argus.rs
@@ -1,0 +1,132 @@
+//! Thin FFI wrapper around NVIDIA Argus/libargus for MIPI CSI camera capture.
+//!
+//! This module provides DMA-buffer frame acquisition from MIPI cameras on Jetson
+//! platforms. Frames are blitted from Argus' EGLStream frame into NvBufSurface
+//! DMA file descriptors that can be passed to the hardware encoder without
+//! CPU-side pixel copies.
+//!
+//! The Argus API is C++, so we use a small C shim (linked via build.rs on
+//! Jetson) to expose the capture session lifecycle.
+
+use std::ffi::c_int;
+use std::io;
+
+/// Opaque handle to an Argus capture session.
+pub struct ArgusCaptureSession {
+    handle: *mut std::ffi::c_void,
+    width: u32,
+    height: u32,
+}
+
+/// A captured Argus frame backed by a DMA buffer.
+pub struct ArgusFrame {
+    /// DMA buffer file descriptor containing an NV12 frame.
+    pub dmabuf_fd: i32,
+    /// Argus sensor start timestamp in nanoseconds, when available.
+    pub sensor_timestamp_ns: Option<u64>,
+    /// Time spent waiting for `FrameConsumer::acquireFrame` to return.
+    pub acquire_wait_ns: u64,
+    /// Time spent copying the acquired EGLStream frame into the DMA buffer.
+    pub blit_ns: u64,
+}
+
+// The C++ session is single-threaded but we move it across the tokio runtime.
+unsafe impl Send for ArgusCaptureSession {}
+
+extern "C" {
+    fn lk_argus_create_session(
+        sensor_index: c_int,
+        width: c_int,
+        height: c_int,
+        fps: c_int,
+    ) -> *mut std::ffi::c_void;
+
+    fn lk_argus_destroy_session(session: *mut std::ffi::c_void);
+
+    /// Acquire the next frame and optionally return the Argus sensor timestamp.
+    /// Returns the NvBufSurface DMA fd, or -1 on error.
+    /// The fd is valid until the next acquire call or `lk_argus_release_frame`.
+    fn lk_argus_acquire_frame_with_metadata(
+        session: *mut std::ffi::c_void,
+        sensor_timestamp_ns: *mut u64,
+        acquire_wait_ns: *mut u64,
+        blit_ns: *mut u64,
+    ) -> c_int;
+
+    /// Release the most recently acquired frame back to the Argus buffer pool.
+    fn lk_argus_release_frame(session: *mut std::ffi::c_void);
+}
+
+impl ArgusCaptureSession {
+    /// Open an Argus capture session on the given MIPI CSI sensor.
+    ///
+    /// `sensor_index` selects the camera (0 for the first CSI camera).
+    /// The session negotiates the given resolution and framerate with the ISP.
+    pub fn new(sensor_index: u32, width: u32, height: u32, fps: u32) -> io::Result<Self> {
+        let handle = unsafe {
+            lk_argus_create_session(
+                sensor_index as c_int,
+                width as c_int,
+                height as c_int,
+                fps as c_int,
+            )
+        };
+        if handle.is_null() {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                "Failed to create Argus capture session",
+            ));
+        }
+        Ok(Self { handle, width, height })
+    }
+
+    /// Acquire the next captured frame as a DMA buffer.
+    ///
+    /// The returned fd refers to an NvBufSurface in NV12 format. It remains
+    /// valid until [`release_frame`](Self::release_frame) is called or the
+    /// next `acquire_frame` implicitly releases the previous one.
+    pub fn acquire_frame(&mut self) -> io::Result<ArgusFrame> {
+        let mut sensor_timestamp_ns = 0;
+        let mut acquire_wait_ns = 0;
+        let mut blit_ns = 0;
+        let fd = unsafe {
+            lk_argus_acquire_frame_with_metadata(
+                self.handle,
+                &mut sensor_timestamp_ns,
+                &mut acquire_wait_ns,
+                &mut blit_ns,
+            )
+        };
+        if fd < 0 {
+            return Err(io::Error::new(io::ErrorKind::Other, "Argus frame acquisition failed"));
+        }
+        Ok(ArgusFrame {
+            dmabuf_fd: fd,
+            sensor_timestamp_ns: (sensor_timestamp_ns > 0).then_some(sensor_timestamp_ns),
+            acquire_wait_ns,
+            blit_ns,
+        })
+    }
+
+    /// Release the most recently acquired frame back to the buffer pool.
+    pub fn release_frame(&mut self) {
+        unsafe { lk_argus_release_frame(self.handle) };
+    }
+
+    pub fn width(&self) -> u32 {
+        self.width
+    }
+
+    pub fn height(&self) -> u32 {
+        self.height
+    }
+}
+
+impl Drop for ArgusCaptureSession {
+    fn drop(&mut self) {
+        if !self.handle.is_null() {
+            unsafe { lk_argus_destroy_session(self.handle) };
+            self.handle = std::ptr::null_mut();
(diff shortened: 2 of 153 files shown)


examples/local_video/Cargo.toml
```diff
@@ -60,3 +60,6 @@ nokhwa = { git = "https://github.com/l1npengtul/nokhwa", rev = "4923ecab7cf26f9d
 
 [target.'cfg(target_os = "windows")'.dependencies]
 nokhwa = { git = "https://github.com/l1npengtul/nokhwa", rev = "4923ecab7cf26f9dba83867a15a9d8662d021296", default-features = false, features = ["input-msmf"] }
+
+[build-dependencies]
+cc = "1"
```
ladvoc: Use the workspace dependency here (see top-level Cargo.toml). The parallel feature is enabled for better performance

examples/local_video/src/publisher.rs
```diff
@@ -349,6 +411,77 @@ fn find_video_outbound_encoder(stats: &[livekit::webrtc::stats::RtcStats]) -> Op
     fallback
 }
 
+fn find_video_outbound_stats(
```
ladvoc, line 414: note: Outside the scope of this PR, but we might want to add some helpers in the core SDK to handle this.

examples/local_video/src/publisher.rs
```diff
@@ -666,7 +811,50 @@ fn list_encoders() {
...
+        camera: Camera,
+        is_yuyv: bool,
+    },
+    #[cfg(all(target_os = "linux", target_arch = "aarch64"))]
+    Argus(argus::ArgusCaptureSession),
+}
+
+/// `CLOCK_MONOTONIC` value used to translate Argus sensor timestamps into wall time.
+#[cfg(all(target_os = "linux", target_arch = "aarch64"))]
+fn monotonic_time_ns_now() -> Option<u64> {
```
ladvoc: suggestion (non-blocking): Consider keeping Argus specific helpers in their own module to keep the main module lean.

examples/local_video/src/argus.rs
```diff
@@ -0,0 +1,132 @@
...
+//! CPU-side pixel copies.
+//!
+//! The Argus API is C++, so we use a small C shim (linked via build.rs on
+//! Jetson) to expose the capture session lifecycle.
+
+use std::ffi::c_int;
+use std::io;
+
+/// Opaque handle to an Argus capture session.
+pub struct ArgusCaptureSession {
```
ladvoc, line 15: note: While this is probably fine for an example, C/C++ bindings typically live in their own -sys crates (e.g., webrtc-sys, argus-sys) and then provide a higher level Rust wrapper that hides any FFI types. Also outside the scope of this PR, but I do see value in eventually shipping a dedicated crate for this and updating the example to use it.

libwebrtc/src/native/video_source.rs
```diff
@@ -130,6 +130,59 @@ impl NativeVideoSource {
         );
     }
 
+    /// Captures a Jetson DMA-buffer backed video frame.
+    ///
+    /// `pixel_format` is `0` for NV12 and `1` for YUV420M.
+    pub fn capture_dmabuf_frame(
```
ladvoc, line 139: suggestion: This and the following method should only be compiled for Linux.

examples/local_video/src/argus.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub acquire_wait_ns: u64,
+    /// Time spent copying the acquired EGLStream frame into the DMA buffer.
+    pub blit_ns: u64,
+}
+
+// The C++ session is single-threaded but we move it across the tokio runtime.
+unsafe impl Send for ArgusCaptureSession {}
+
+extern "C" {
+    fn lk_argus_create_session(
```
alan-george-lk, line 37: This example is cool--if I'm tracking the pieces right, couldn't this code be used by consumers in library form (vs example runnable form) to do Argus frame capture? Or is the expectation that user pipelines will do their own Argus interfacing, this is just an example of how it connects with LiveKit?

Or am I not following right?

webrtc-sys/include/livekit/dmabuf_video_frame_buffer.h
```diff
@@ -0,0 +1,69 @@
...
+  webrtc::scoped_refptr<webrtc::VideoFrameBuffer> CropAndScale(
+      int offset_x,
+      int offset_y,
+      int crop_width,
+      int crop_height,
+      int scaled_width,
+      int scaled_height) override;
+
+  // DMA buffer accessors
+  int dmabuf_fd() const { return dmabuf_fd_; }
```
alan-george-lk, line 56: Nit: in the C++ SDK we aligned all methods to `camelBack()` case, since this isn't in the SDK not a biggie just for your awareness. If the webrtc stuff does `snake_case` for base accessors and `PascalCase` for operations (which it appears to from the `override` lines above), nothing to change such that it's consistent.

webrtc-sys/src/jetson/av1_encoder_impl.h
```diff
@@ -0,0 +1,84 @@
+#ifndef WEBRTC_JETSON_AV1_ENCODER_IMPL_H_
+#define WEBRTC_JETSON_AV1_ENCODER_IMPL_H_
```
alan-george-lk, line 18: Nit: `#pragma once` is more modern and we've standardized it in the SDK. If this is the convention elsewhere in this folder/WebRTC style, disregard

ladvoc (approved):

webrtc-sys/src/jetson/av1_encoder_impl.h
```diff
@@ -0,0 +1,84 @@
...
+  void SetRates(const RateControlParameters& rc_parameters) override;
+
+  EncoderInfo GetEncoderInfo() const override;
+
+ private:
+  int32_t ProcessEncodedFrame(std::vector<uint8_t>& packet,
+                              const ::webrtc::VideoFrame& input_frame,
+                              bool is_keyframe);
+
+  const webrtc::Environment& env_;
```
alan-george-lk, line 76: Minor: I am generally avoidant of inheritance (esp double inheritance, given this already extends `VideoEncoder`, but there seems to be a lot of shared state between these `_impl` classes. 

Consider a `JetsonVideoEncoder` class that extends `VideoEncoder` which these then extend, or for composition instead, a helper object that defines all the common variables/methods and then is instantiated in each of these. One layer of indirection, but less copy/pasta

webrtc-sys/src/jetson/av1_encoder_impl.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include "jetson_mmapi_encoder.h"
+
+namespace webrtc {
+
+class JetsonAV1EncoderImpl : public VideoEncoder {
+ public:
+  struct LayerConfig {
+    int simulcast_idx = 0;
+    int width = -1;
```
alan-george-lk, line 37: Nit: `std::optional` is more modern/clear than default negative values, but this certainly works given the context of the data, and is more lightweight/C-like

webrtc-sys/src/jetson/h264_encoder_impl.cpp
```diff
@@ -0,0 +1,582 @@
...
+  }
+  RTC_HISTOGRAM_ENUMERATION("WebRTC.Video.H264EncoderImpl.Event",
+                            kH264EncoderEventError, kH264EncoderEventMax);
+  has_reported_error_ = true;
+}
+
+int32_t JetsonH264EncoderImpl::InitEncode(
+    const VideoCodec* inst,
+    const VideoEncoder::Settings& settings) {
+  const bool debug = std::getenv("LK_ENCODER_DEBUG") != nullptr;
```
alan-george-lk: This looks like a new environment variable just for encoder debugging (checked this symbol on `main`). Might be worth documenting in `.md` somewhere that it exists and what it does, otherwise it's kind of "magic number" territory. Also is it worth considering tying it into the `env_logger` hooks for debug?

webrtc-sys/src/jetson/h264_encoder_impl.cpp
```diff
@@ -0,0 +1,582 @@
...
+  std::vector<uint8_t> packet;
+  bool is_keyframe = false;
+
+  // Check for DmaBuf zero-copy path first.
+  auto* dmabuf = livekit::DmaBufVideoFrameBuffer::FromNative(
+      input_frame.video_frame_buffer().get());
+  if (dmabuf) {
+    if (!logged_first_encode.exchange(true)) {
+      std::fprintf(stderr,
+                   "[H264Impl] First Encode() via DmaBuf: fd=%d, %dx%d, "
```
alan-george-lk: Minor: This doesn't seem like an error, and the code execution doesn't stop here, is this meant to be a debug log or actually stdout? A lot of stderr usage in this PR for debug lines, which is fine, I would just be cognizant of the happy path to avoid confusion when reviewing log output. This comment applies to a few spots in this file and may apply elsewhere

webrtc-sys/src/jetson/jetson_encoder_factory.cpp
```diff
@@ -0,0 +1,116 @@
...
+  // the software libaom encoder.
+  if (livekit::JetsonMmapiEncoder::IsCodecSupported(
+          livekit::JetsonCodec::kAV1)) {
+    absl::InlinedVector<ScalabilityMode, kScalabilityModeCount>
+        scalability_modes;
+    scalability_modes.push_back(ScalabilityMode::kL1T1);
+    supported_formats_.push_back(
+        SdpVideoFormat(SdpVideoFormat::AV1Profile0(), scalability_modes));
+    RTC_LOG(LS_INFO) << "Jetson MMAPI AV1 encoder available.";
+    std::cout << "Jetson MMAPI AV1 encoder available." << std::endl;
```
alan-george-lk, line 55: Does `RTC_LOG` not hit `stdout`?

webrtc-sys/src/jetson/jetson_mmapi_encoder.cpp
```diff
@@ -0,0 +1,1967 @@
...
+  } else if (verbose) {
+    std::fprintf(stderr, "[MMAPI] AV1_HEADERS_WITH_FRAME disabled (WebRTC OBU)\n");
+    std::fflush(stderr);
+  }
+}
+
+bool JetsonMmapiEncoder::ConfigureAv1Encoder() {
+  const bool verbose = std::getenv("LK_ENCODER_DEBUG") != nullptr;
+
+#ifdef V4L2_CID_MPEG_VIDEOENC_AV1_TILE_CONFIGURATION
```
alan-george-lk, line 389: I don't see any other references to this `define` in this PR, does this come from WebRTC headers?

webrtc-sys/src/jetson/jetson_mmapi_encoder.cpp
```diff
@@ -0,0 +1,1967 @@
...
+                   errno, strerror(errno));
+      std::fflush(stderr);
+    }
+  }
+
+  return true;
+}
+
+bool JetsonMmapiEncoder::ConfigureEncoder() {
+  const bool verbose = std::getenv("LK_ENCODER_DEBUG") != nullptr;
```
alan-george-lk: Nit: `debug` is used for this bool elsewhere, `verbose` here

webrtc-sys/src/jetson/jetson_mmapi_encoder.cpp
```diff
@@ -0,0 +1,1967 @@
...
+  const bool have_y = mmap_sync_supported_ &&
+      GetPitchAndHeightFromNvBufSurfaceFd(buffer->planes[0].fd, 0, &y_pitch, &y_h, &y_np);
+  const bool have_u = mmap_sync_supported_ &&
+      (buffer->n_planes > 1) &&
+      GetPitchAndHeightFromNvBufSurfaceFd(buffer->planes[1].fd, 1, &u_pitch, &u_h, &u_np);
+  const bool have_v = mmap_sync_supported_ &&
+      (!output_is_nv12_ && buffer->n_planes > 2) &&
+      GetPitchAndHeightFromNvBufSurfaceFd(buffer->planes[2].fd, 2, &v_pitch, &v_h, &v_np);
+
+  auto stride_from_plane = [&](int plane_index,
```
alan-george-lk: I wonder if this lambda (and similar sections of this method) could be written in a way that don't depend on HW state, such that they could be unit tested

webrtc-sys/src/jetson/jetson_mmapi_encoder.h
```diff
@@ -0,0 +1,147 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
+
+class NvVideoEncoder;
+
+namespace livekit {
+
+enum class JetsonCodec { kH264, kH265, kAV1 };
```
alan-george-lk: Nit: suggest `JetsonCodec : uint8_t` to scope this enum to 256 values, otherwise it's 32 bits for a tiny enum

webrtc-sys/build.rs
```diff
@@ -211,6 +212,57 @@ fn main() {
                 }
             }
 
+            if arm {
+                let jetson_mmapi_include = PathBuf::from("/usr/src/jetson_multimedia_api/include");
+                if jetson_mmapi_include.exists() {
+                    let jetson_classes_dir =
+                        PathBuf::from("/usr/src/jetson_multimedia_api/samples/common/classes");
```
alan-george-lk, line 219: Should we check if this path is valid as well?

alan-george-lk (approved):

examples/local_video/src/argus.rs
chenosaurus: good call out, I've created a new project to create new `livekit-capture` crate and migrate all capture related code there.  Will refactor examples once we have that.

examples/local_video/src/argus.rs
chenosaurus: yep, we plan on creating a dedicated capture crate that contains all the different platform capture paths.  For now this is just inline in the example app so it can run on a jetson.  I would expect customers to do their own integration, and they will probably use C++ SDK if they were using libargus anyways.

webrtc-sys/src/jetson/av1_encoder_impl.h
chenosaurus: I just followed the style from the other encoder implementations

webrtc-sys/src/jetson/h264_encoder_impl.cpp
chenosaurus: good catch, gonna remove this I don't think we need to keep it in here

webrtc-sys/src/jetson/jetson_mmapi_encoder.cpp
chenosaurus: this comes from jetson's v4l2 extension headers

webrtc-sys/build.rs
chenosaurus: hmm, I wonder if we should just copy the files over and include them in our repo like we do w/ the nvenc related files

chenosaurus: @cloudwebrtc any opinions?  I see for nvenc we added NvEncoder.x NvDecoder.x into the repo.

webrtc-sys/src/jetson/jetson_mmapi_encoder.h
chenosaurus: fixed

webrtc-sys/src/jetson/jetson_mmapi_encoder.cpp
chenosaurus: refactored into a new file, added unit tests

webrtc-sys/src/jetson/jetson_mmapi_encoder.cpp
chenosaurus: removed all the debug stuff

examples/local_video/src/publisher.rs
chenosaurus: moved them to the argus mod

examples/local_video/src/publisher.rs
chenosaurus: yea good idea

PR #1172, opened 2026-06-16
perf(livekit-uniffi): omit Android unwind tables in release builds

davidliu:
## Summary

- Add **`-C force-unwind-tables=no`** for livekit-uniffi **Android release** native builds to reduce `.eh_frame` / unwind metadata in shipped `.so` files.
- Scope the flag to **`cargo make --profile release`** (via `RUSTFLAGS` on `build-android`) so the default development profile still emits unwind tables — better for local crash backtraces when iterating on Rust code.
- Make Android packaging **profile-aware** (mirroring iOS): development builds debug Rust artifacts; release builds optimized artifacts and run the existing arm64 size gate. Gradle continues to produce the **release AAR** in both cases.

## Why

Modern Rust may emit unwind tables even with `panic = "abort"`. For mobile SDKs that trade crash-time unwinding for binary size, `force-unwind-tables=no` is a standard knob. The change only affects release builds, development builds still keep the tables for debuggability.

## Size impact

Measured on **`blaze/uniffi-size-trim`** branch (due to some of the savings overlapping) with symmetric `aarch64-linux-android` release builds:
| Variant | Size |
|---------|------|
| Baseline (no flag) | **836,136 B (816.5 KiB)** |
| `-C force-unwind-tables=no` | **718,744 B (701.9 KiB)** |
| **Savings** | **117,392 B (114.6 KiB, ~14%)** |

## Tradeoff

Release Android `.so` files may produce **sparser on-device crash stacks** before symbolication. Acceptable for a `panic = "abort"` UniFFI library optimized for size; debug builds are unaffected.

## Validation

* `cargo make --profile release android-package` passes `android-check-size`
* Release arm64-v8a `.so` in the AAR is smaller than without the flag
* `cargo make android-package` (development) still builds debug `.so` files without the flag

### DIFF
--- a/livekit-uniffi/Makefile.toml
+++ b/livekit-uniffi/Makefile.toml
@@ -51,14 +51,26 @@ else
 end
 """
 
+[tasks.android-shared]
+private = true
+
+[tasks.android-shared.env]
+ANDROID_CARGO_OUT_DIR = { source = "${CARGO_MAKE_PROFILE}", mapping = { development = "debug", release = "release", default = "debug" } }
+ANDROID_AAR_BASENAME = "livekit-uniffi-android-release"
+
 [tasks.build-android]
+extend = "android-shared"
 install_crate = { crate_name = "cargo-ndk", binary = "cargo-ndk", test_arg = "--help" }
-script_runner = "@duckscript"
+script_runner = "@shell"
 script = """
 echo "TARGET: ${TARGET}"
-exec rustup target add ${TARGET}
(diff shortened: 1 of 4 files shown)


ladvoc (approved):

pblazej (approved):
👏

davidliu:
For iOS, looks like it was already not emitting unwind tables, so this is probably a big reason why there was a noticeable size disparity between the two previously.

PR #1174, opened 2026-06-16
Explicitly drop FFI handles on dispose

alan-george-lk:
### Before you submit your PR

### PR description

This PR resolves an issue observed in the downstream C++ SDK during repeated unit/integration test runs, where the FFI handle state was not completely cleared and caused issues for different test cases intended to be atomic, but within the same process boundary. This specifically raised a crash in PlatformAudio tests.

### Breaking changes

N/A

### MSRV

### Testing

Validated in a downstream C++ SDK run against this branch.

### Async

### DIFF
--- a/livekit-ffi/src/server/mod.rs
+++ b/livekit-ffi/src/server/mod.rs
@@ -176,8 +176,9 @@ impl FfiServer {
 
         self.logger.set_capture_logs(false);
 
-        // Drop all handles
         *self.config.lock() = None; // Invalidate the config
+        self.ffi_handles.clear();
+        self.handle_dropped_txs.clear();
     }
 
     pub fn send_event(&self, message: proto::ffi_event::Message) -> FfiResult<()> {
--- a/.changeset/ffi-handle-cleanup.md
+++ b/.changeset/ffi-handle-cleanup.md
@@ -1,0 +1,5 @@
+---
+livekit-ffi: patch
+---
+
+Clear remaining FFI handles during dispose so platform audio resources are released across repeated initialize/shutdown cycles.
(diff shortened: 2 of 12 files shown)


stephen-derosa (approved):

livekit-ffi/src/server/mod.rs
```diff
@@ -176,8 +176,10 @@ impl FfiServer {
 
         self.logger.set_capture_logs(false);
 
-        // Drop all handles
-        *self.config.lock() = None; // Invalidate the config
+        // Drop remaining FFI handles so native resources they own are released on dispose.
+        *self.config.lock() = None;
+        self.ffi_handles.clear();
```
ladvoc, line 180: suggestion: Log how many handles remain at dispose time for visibility:
```rust
log::debug!("{} FFI handles remaining at dispose", self.ffi_handles.len());
```

ladvoc (approved):

livekit-ffi/src/server/mod.rs
alan-george-lk: Added

xianshijing-lk (approved):

livekit-ffi/src/server/mod.rs
```diff
@@ -176,8 +176,21 @@ impl FfiServer {
...
         *self.config.lock() = None; // Invalidate the config
+
+        // Drop remaining FFI handles so native resources they own are released on dispose.
+        log::debug!("{} FFI handles remaining at dispose before clearing", self.ffi_handles.len());
+        let platform_audio_handles = self
+            .ffi_handles
+            .iter()
+            .filter(|handle| handle.value().is::<platform_audio::FfiPlatformAudio>())
+            .count();
+        log::debug!(
```
xianshijing-lk: unless we add an explicit API to shutdown the PlatformAudio, otherwise it is expected that some platform_audio_handles will remain valid here.

alan-george-lk:
Moving this to https://github.com/livekit/rust-sdks/pull/1193, closing this one.

PR #1175, opened 2026-06-17
fix: escalate to full reconnect if connection failed during a resume

davidzhao:
if connections cannot be successfully resumed, previously it could be stuck in a resume cycle without correctly escalating it to a full reconnect loop.

added test cases to cover this scenario. the tests were failing before the fix

### DIFF
--- a/livekit/src/room/mod.rs
+++ b/livekit/src/room/mod.rs
@@ -830,6 +830,21 @@ impl Room {
         self.inner.rtc_engine.simulate_scenario(scenario).await
     }
 
+    /// Test-only: force the next `count` resume attempts to fail, exercising the
+    /// resume-failure → full-reconnect escalation path end-to-end.
+    #[cfg(feature = "__lk-e2e-test")]
+    pub fn fail_next_resume_attempts(&self, count: u32) {
+        self.inner.rtc_engine.fail_next_resume_attempts(count);
+    }
+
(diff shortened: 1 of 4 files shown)


livekit/src/room/mod.rs
```diff
@@ -830,6 +830,21 @@ impl Room {
...
+    #[cfg(feature = "__lk-e2e-test")]
+    pub fn fail_next_resume_attempts(&self, count: u32) {
+        self.inner.rtc_engine.fail_next_resume_attempts(count);
+    }
+
+    /// Test-only: arm a one-shot fault so the next resume simulates a concurrent
+    /// transport failure (then still succeeds), reproducing the resume-reports-
+    /// success-while-a-failure-was-pending race.
+    #[cfg(feature = "__lk-e2e-test")]
+    pub fn escalate_during_next_resume(&self) {
```
xianshijing-lk: nit, regarding naming, consider fail_transport_during_next_resume(&self) ?  if that makes things a bit more clear.

livekit/src/room/mod.rs
```diff
@@ -830,6 +830,21 @@ impl Room {
...
+        self.inner.rtc_engine.escalate_during_next_resume();
```
xianshijing-lk: ditto, consider name the function fail_transport_during_next_resume()

xianshijing-lk (approved):
Just one nit, lgtm

PR #1180, opened 2026-06-19
add user_data trailer type

chenosaurus:


libwebrtc/src/video_frame.rs
```diff
@@ -55,13 +55,20 @@ pub enum VideoBufferType {
 /// Metadata carried alongside a video frame via the packet trailer mechanism.
 ///
 /// Each field corresponds to an independently negotiable packet trailer feature
-/// (`PTF_USER_TIMESTAMP`, `PTF_FRAME_ID`), so individual fields are `Option`.
-#[derive(Debug, Clone, Copy)]
+/// (`PTF_USER_TIMESTAMP`, `PTF_FRAME_ID`, `PTF_USER_DATA`), so individual fields
+/// are `Option`.
+#[derive(Debug, Clone)]
 pub struct FrameMetadata {
```
ladvoc, line 61: question: Could we define a helper on `FrameMetadata` so the user can calculate their exact max `user_data` length based on the other fields they have set?

chenosaurus: I'm not sure if FrameMetadata would be able to calculate it since all the fields get set at the same time usually.  Otherwise it would need to know about the Features that were enabled on the TrackPublishOptions.

examples/local_video/src/publisher.rs
```diff
@@ -1310,6 +1328,7 @@ async fn run_capture_loop(
     height: u32,
     display_shared: Option<Arc<Mutex<SharedYuv>>>,
     publish_timing_state: Option<Arc<Mutex<PublisherTimingState>>>,
+    user_data_channels: Option<Arc<Mutex<[f32; user_data::NUM_CHANNELS]>>>,
```
ladvoc, line 1331: suggestion(non-blocking): Calling this data channel might imply this is related to WebRTC data channels, not frame metadata.

ladvoc (approved):
LGTM ✅

PR #1183, opened 2026-06-23
Add dart/flutter bindings to livekit-uniffi

jhugman:
### Before you submit your PR

- [X] I have read the [contributing guidelines](https://github.com/livekit/rust-sdks/blob/main/CONTRIBUTING.md) and validated that this PR will be accepted.
- [X] I have read and followed the principles regarding breaking changes, testing, and code quality.

### PR description

Added Dart bindings with the uniffi-dart bindgen. This generates a `Native Assets` API. Exercise with `cargo make bindgen-dart`.

This adds a uniffi-bindgen-dart feature to the `livekit-uniffi` crate: followup would be to separate out the uniffi-tooling binaries in to a workspace centred tooling crate.

Added consumable package, with a hook/build.dart. This works both locally, and by downloading a prebuilt cdylib.

Added CI which builds the cdylibs per OS/architecture on release.

### Breaking changes

No changes breaking changes are made, however:

- `uniffi-dart` bindgen needed a bug fix which means the version of uniffi-dart is from main (pinned to the hash).
-  the version of uniffi is incremented to 0.31. This is unlikely to cause breakage in the Swift, Python, Kotlin bindings, but _may_ cause problems for `uniffi-bindgen-node`. Followup: move `uniffi-bindgen-node` to `uniffi-bindgen-react-native`.

### MSRV

No changes. uniffi 0.31 requires >= 1.85

### Testing

- Dart package tests (`support/dart/tests`): a smoke test including a negative case.
- CI build & hook/build.dart tested via dry run.

### Async


livekit-uniffi/support/dart/hook/build.dart.tera
```diff
@@ -0,0 +1,141 @@
...
+
+  if (!await libFile.exists()) {
+    throw Exception('$libName not found in $zipUrl');
+  }
+  output.dependencies.add(libFile.uri);
+  return libFile.uri;
+}
+
+Future<List<int>> _downloadAndVerify(Uri zipUrl) async {
+  final zipResp = await http.get(zipUrl);
```
1egoman: thought: Does this request have a default timeout? I'm assuming so but it would be worth confirming.

thought: Also, does this `http.get` follow `3xx` redirects properly? If not, I think it would be worth making sure it does, that opens up a lot of future flexibility.

livekit-uniffi/support/dart/hook/build.dart.tera
```diff
@@ -0,0 +1,141 @@
...
+  if (zipResp.statusCode != 200) {
+    throw Exception('Failed to download $zipUrl: HTTP ${zipResp.statusCode}');
+  }
```
1egoman: nitpick: Also in the aim of future flexibility, maybe making this verify 2xx not strictly 200 might be a good idea.

livekit-uniffi/Cargo.toml
```diff
@@ -14,18 +14,32 @@ publish = false
...
+uniffi-dart = { git = "https://github.com/Uniffi-Dart/uniffi-dart", rev = "90f2c6f29cbf88c8bc2cf515e6a0c2314a48844c", optional = true }
+camino = { version = "1", optional = true }
+
+[features]
+# Enables the `uniffi-bindgen-dart` bin. Not enabled for library builds.
+dart-bindgen = ["dep:uniffi-dart", "dep:camino"]
 
 [build-dependencies]
-uniffi = { version = "0.30.0", features = ["build", "scaffolding-ffi-buffer-fns"] }
+uniffi = { version = "0.31", features = ["build", "scaffolding-ffi-buffer-fns"] }
```
1egoman, line 32: nitpick: You probably know better than I do, but I just wanted to 100% confirm that all other bindgens being used here support this new version.

I know `uniffi-bindgen-node` does not but we previously discussed that you were going to migrate this to UBRN shortly so I think that's ok given this isn't being used for anything right now. Just want to confirm there's no other platforms that were missed.

livekit-uniffi/support/dart/hook/build.dart.tera
```diff
@@ -0,0 +1,141 @@
...
+
+    output.assets.code.add(
+      CodeAsset(
+        package: input.packageName,
+        // Dart prefixes this with `package:<packageName>/`.
+        name: 'uniffi:$_cdylibName',
+        linkMode: DynamicLoadingBundled(),
+        file: libFile,
+      ),
+    );
```
1egoman, line 52: thought: An interesting related blog post I came across in my research: https://www.simonbinder.eu/posts/native_assets/

I'm not sure how much of this applies since this is a net new package, but the thing that stood out to me was that it seems like by using native assets (I think that's what you are doing here?) you potentially force consumers to also enable the native assets feature. Maybe that's fine but could be problematic once this gets folded into the dart sdk and could maybe break existing users

(Feel free to ignore this if it's not relevant, you're probably done more research into dart packaging than I have at this point)

1egoman (commented):
A few high level thoughts. I am deferring an official review approval to folks with more first hand flutter experience.

livekit-uniffi/Cargo.toml
jhugman: yes. `uniffi-bindgen-dart`, `ubrn`, `uniffi` all support `0.31.x`.

livekit-uniffi/support/dart/hook/build.dart.tera
jhugman: As we discussed (but duplicated here for posterity):
- `uniffi-dart` generates Native Assets bindings, and `uniffi-bindgen-dart` generates DynamicLibrary based bindings. 
- I think this would mean changing the bindings, i.e. changing upstream bindgens.
- @hiroshihorie expressed no strong opinion about which flavour of bindings to support.

Additionally:
- this post dates from when Native Assets were experimental, 2024-early 2025; it was the flipping of the `--enable-experiment=native-assets` flag's virality that was a concern for this post.
- Native Assets are now the blessed path forward, gated on dart version [3.10 / flutter 3.38](https://docs.flutter.dev/release/release-notes/release-notes-3.38.0)

livekit/client-sdk-flutter specifies min versions: [3.60 / 3.27](https://github.com/livekit/client-sdk-flutter/blob/main/pubspec.yaml#L21-L23) so this would cause churn for users when we migrate from a pure-dart to a Rust based SDK. Checking with @hiroshihorie for advice.

livekit-uniffi/support/dart/hook/build.dart.tera
jhugman: `package:http`'s `get` follows `3xx` for GET by default `followRedirects = true, maxRedirects = 5`.

There is no timeout, so I added one for 60 seconds.

livekit-uniffi/support/dart/pubspec.yaml.tera
```diff
@@ -0,0 +1,20 @@
...
+
+dev_dependencies:
+  # Native Assets build-hook support. `any` tracks the SDK-bundled versions.
+  code_assets: any
+  hooks: any
+  # Used by hook/build.dart to download and verify prebuilt libraries.
+  archive: ^4.0.0
+  crypto: ^3.0.0
+  http: ^1.2.0
+  test: ^1.25.0
```
hiroshihorie, line 20: Since consumers compile and run `hook/build.dart` at their own build time, its imports need to be regular dependencies. A dependency's `dev_dependencies` are not resolved for downstream apps, so as written the hook fails to compile once an app depends on `livekit_uniffi` (the [hooks doc](https://dart.dev/tools/hooks) calls this out). It works in-repo because the package is the root package there. I reproduced both the failure and the fix with a small path-dependency consumer.

```suggestion
dependencies:
  ffi: ^2.1.0
  # Native Assets build-hook support. `any` tracks the SDK-bundled versions.
  code_assets: any
  hooks: any
  # Used by hook/build.dart to download and verify prebuilt libraries.
  archive: ^4.0.0
  crypto: ^3.0.0
  http: ^1.2.0

dev_dependencies:
  test: ^1.25.0
```

`test` is the only one that is genuinely dev-only. 🙂

livekit-uniffi/support/dart/hook/build.dart.tera
```diff
@@ -0,0 +1,150 @@
...
+  } else if (os == OS.windows) {
+    if (arch == Architecture.arm64) return 'aarch64-pc-windows-msvc';
+    if (arch == Architecture.x64) return 'x86_64-pc-windows-msvc';
+  } else if (os == OS.android) {
+    if (arch == Architecture.arm64) return 'aarch64-linux-android';
+    if (arch == Architecture.arm) return 'armv7-linux-androideabi';
+    if (arch == Architecture.x64) return 'x86_64-linux-android';
+  } else if (os == OS.iOS) {
+    if (arch == Architecture.arm64) return 'aarch64-apple-ios';
+  }
```
hiroshihorie, line 170: thought: Is iOS Simulator in scope here? `_targetTriple` maps iOS + arm64 straight to the device triple `aarch64-apple-ios` without checking `code.iOS.targetSdk`, and the workflow matrix builds no simulator artifacts (`aarch64-apple-ios-sim`, `x86_64-apple-ios`). So a simulator run on an Apple Silicon Mac downloads the device dylib and fails later in the Xcode link step with a Mach-O platform mismatch ("building for iOS Simulator, but linking in dylib built for iOS"), while an Intel Mac throws `UnsupportedError`.

If simulator support is intended, the fix is a `targetSdk` check in the hook plus the two simulator targets in the matrix. If device-only is intentional for this PR, a guard like this would turn the confusing late link error into a clear one:

```dart
if (code.iOS.targetSdk == IOSSdk.iPhoneSimulator) {
  throw UnsupportedError('iOS Simulator is not supported yet');
}
```

Either way works, mostly flagging it since the matrix building `aarch64-apple-ios` makes iOS look fully supported.

livekit-uniffi/support/dart/pubspec.yaml.tera
jhugman: Ok, everything else to `dependencies`.

livekit-uniffi/support/dart/hook/build.dart.tera
jhugman: Added support for the simulator.

hiroshihorie (approved):

PR #1191, opened 2026-06-25
NVENC AV1 support

chenosaurus:
- enable Nvidia GPU NVENC AV1 encoding


webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
```diff
@@ -119,30 +197,81 @@ bool NvidiaVideoEncoderFactory::IsSupported() {
...
+        << "NVIDIA AV1 NVENC encoder is not supported on this GPU.";
+  }
+
+  /*std::map<std::string, std::string> highParameters = {
+      {"profile-level-id", "4d0032"},
+      {"level-asymmetry-allowed", "1"},
+      {"packetization-mode", "1"},
+  };
+
+  supported_formats_.push_back(SdpVideoFormat("H264", highParameters));
```
xianshijing-lk: why these code is commented out ?

webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
```diff
@@ -119,30 +197,81 @@ bool NvidiaVideoEncoderFactory::IsSupported() {
...
+  */
+}
+
+NvidiaVideoEncoderFactory::~NvidiaVideoEncoderFactory() {}
+
+bool NvidiaVideoEncoderFactory::IsSupported() {
+  const NvencProbeResult probe = ProbeNvencSupport();
```
xianshijing-lk: how about using function local to avoid calling ProbeNvencSupport() more than once ? 
And log it regardless if it is available or not.

Function-local static (recommended)
bool NvidiaVideoEncoderFactory::IsSupported() {
  static const NvencProbeResult probe = [] {
    NvencProbeResult result = ProbeNvencSupport();
    RTC_LOG(LS_INFO)
        << "NVIDIA NVENC hardware encoder "
        << (result.encoder_supported ? "is available." : "is not available.");

    return result;
  }();

  return probe.encoder_supported;
}

webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
```diff
@@ -50,16 +129,19 @@ bool NvidiaVideoEncoderFactory::IsSupported() {
...
   auto NvEncodeAPIGetMaxSupportedVersion =
       (NVENCSTATUS(NVENCAPI*)(uint32_t*))dlsym(
           hModule, "NvEncodeAPIGetMaxSupportedVersion");
   auto NvEncodeAPICreateInstance =
       (NVENCSTATUS(NVENCAPI*)(NV_ENCODE_API_FUNCTION_LIST*))dlsym(
           hModule, "NvEncodeAPICreateInstance");
 
-  bool supported = false;
+  NV_ENCODE_API_FUNCTION_LIST fnList = {NV_ENCODE_API_FUNCTION_LIST_VER};
+  CUcontext cuCtx = nullptr;
```
xianshijing-lk, line 143: preferably, in C++, the variables should be declared just right before they are used.

webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
```diff
@@ -12,35 +20,106 @@
...
+  if (nvStatus != NV_ENC_SUCCESS || format_count == 0) {
+    return false;
+  }
 
-bool NvidiaVideoEncoderFactory::IsSupported() {
+  std::vector<NV_ENC_BUFFER_FORMAT> formats(format_count);
+  uint32_t written_format_count = 0;
+  nvStatus = fnList.nvEncGetInputFormats(
+      hEncoder, encodeGuid, formats.data(), format_count, &written_format_count);
+  if (nvStatus != NV_ENC_SUCCESS) {
```
xianshijing-lk: nit, probably you want to add || written_format_count == 0 in this check ?

webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
```diff
@@ -12,35 +20,106 @@
...
+      fnList.nvEncGetEncodeGUIDCount(hEncoder, &guid_count);
+  if (nvStatus != NV_ENC_SUCCESS || guid_count == 0) {
+    return false;
+  }
+
+  std::vector<GUID> guids(guid_count);
+  uint32_t written_guid_count = 0;
+  nvStatus = fnList.nvEncGetEncodeGUIDs(hEncoder, guids.data(), guid_count,
+                                        &written_guid_count);
+  if (nvStatus != NV_ENC_SUCCESS) {
```
xianshijing-lk: what should happen if written_guid_count is 0 here ?

webrtc-sys/src/nvidia/av1_encoder_impl.h
```diff
@@ -0,0 +1,94 @@
...
+                              const ::webrtc::VideoFrame& input_frame,
+                              bool is_keyframe);
+
+  const webrtc::Environment& env_;
+  EncodedImageCallback* encoded_image_callback_ = nullptr;
+
+  std::unique_ptr<NvEncoder> encoder_;
+  CUcontext cu_context_;
+  CUmemorytype cu_memory_type_;
+  CUarray cu_scaled_array_;
```
xianshijing-lk: initialize it to nullptr

webrtc-sys/src/nvidia/av1_encoder_impl.h
```diff
@@ -0,0 +1,94 @@
...
+ private:
+  int32_t ProcessEncodedFrame(std::vector<uint8_t>& packet,
+                              const ::webrtc::VideoFrame& input_frame,
+                              bool is_keyframe);
+
+  const webrtc::Environment& env_;
+  EncodedImageCallback* encoded_image_callback_ = nullptr;
+
+  std::unique_ptr<NvEncoder> encoder_;
+  CUcontext cu_context_;
```
xianshijing-lk: nit, should they be const, like 
const CUcontext cu_context_;
const CUmemorytype cu_memory_type_;
const NV_ENC_BUFFER_FORMAT nv_format_;

webrtc-sys/src/nvidia/av1_encoder_impl.h
```diff
@@ -0,0 +1,94 @@
...
+
+  int32_t Encode(const VideoFrame& frame,
+                 const std::vector<VideoFrameType>* frame_types) override;
+
+  void SetRates(const RateControlParameters& rc_parameters) override;
+
+  EncoderInfo GetEncoderInfo() const override;
+
+ private:
+  int32_t ProcessEncodedFrame(std::vector<uint8_t>& packet,
```
xianshijing-lk: nit, is packet mutable ? if not, use const std::vector<uint8_t>& packet,

webrtc-sys/src/nvidia/av1_encoder_impl.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+  GUID encodeGuid = NV_ENC_CODEC_AV1_GUID;
+  GUID presetGuid = NV_ENC_PRESET_P4_GUID;
+
+  encoder_->CreateDefaultEncoderParams(&nv_initialize_params_, encodeGuid,
+                                       presetGuid,
+                                       NV_ENC_TUNING_INFO_ULTRA_LOW_LATENCY);
+
+  nv_initialize_params_.frameRateNum =
+      static_cast<uint32_t>(configuration_.max_frame_rate);
```
xianshijing-lk: is max_frame_rate guaranteed bigger than 1 ? otherwise it will be divided by 0

xianshijing-lk: maybe it is safer to do 
nv_initialize_params_.frameRateNum =
    std::max<uint32_t>(1, static_cast<uint32_t>(std::round(configuration_.max_frame_rate)));

webrtc-sys/src/nvidia/av1_encoder_impl.cpp
```diff
@@ -0,0 +1,470 @@
...
+void NvidiaAV1EncoderImpl::ReportInit() {
+  if (has_reported_init_) {
+    return;
+  }
+  RTC_HISTOGRAM_ENUMERATION("WebRTC.Video.AV1EncoderImpl.Event",
+                            kAV1EncoderEventInit, kAV1EncoderEventMax);
+  has_reported_init_ = true;
+}
+
+void NvidiaAV1EncoderImpl::ReportError() {
```
xianshijing-lk, line 66: should ReportError() take the error as param and report the right errors?

webrtc-sys/src/nvidia/av1_encoder_impl.cpp
```diff
@@ -0,0 +1,470 @@
...
+  info.supports_native_handle = false;
+  info.implementation_name = "NVIDIA AV1 Encoder";
+  info.scaling_settings = VideoEncoder::ScalingSettings::kOff;
+  info.is_hardware_accelerated = true;
+  info.supports_simulcast = false;
+  info.preferred_pixel_formats = {VideoFrameBuffer::Type::kI420};
+  return info;
+}
+
+void NvidiaAV1EncoderImpl::SetRates(
```
xianshijing-lk, line 434: It looks like SetRate() only set the local codec_ and configuration_ cache values. 

Should it trigger encoder_->Reconfigure(... if encoder_is already running ? or that is not an supported use case

webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
chenosaurus: I removed it, we only support h264 baseline

webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
chenosaurus: they are outside of the do {} block so they can be cleaned up after

webrtc-sys/src/nvidia/nvidia_encoder_factory.cpp
chenosaurus: also add check for || written_guid_count == 0 to short circuit and return

webrtc-sys/src/nvidia/av1_encoder_impl.h
chenosaurus: made it const

webrtc-sys/src/nvidia/av1_encoder_impl.cpp
chenosaurus: it's possible, this was carried over from the other encoder implementations.  I've fixed it across av1, h264, & h265.

webrtc-sys/src/nvidia/av1_encoder_impl.cpp
chenosaurus: I think this is an existing problem in all of the nvenc encoder implementations, I will address in a different PR to fully wire up SetRate().

ladvoc (approved):
LGTM ✅

PR #1196, opened 2026-06-28
feat: auto failover APIs with LK Cloud

davidzhao:
retries in alternative datacenters on 5xx and transport failures


livekit-api/src/services/api_test.rs
```diff
@@ -0,0 +1,211 @@
...
+macro_rules! skip_if_offline {
+    ($base:expr) => {
+        if !reachable(&$base).await {
+            eprintln!("skipping: mock test server not reachable at {}", $base);
+            return;
+        }
+    };
+}
+
+#[tokio::test]
```
ladvoc, line 85: suggestion(non-blocking): If capturing log output from these tests is desired, the _test_log_ macro can be added: `#[test_log::test(tokio::test)]`.

livekit-api/src/services/sip.rs
```diff
@@ -154,6 +155,17 @@ pub struct CreateSIPParticipantOptions {
     pub ringing_timeout: Option<Duration>,
     pub max_call_duration: Option<Duration>,
     pub enable_krisp: Option<bool>,
+    /// SIP headers sent as-is on the INVITE; may help the SIP endpoint identify
+    /// the call as coming from LiveKit.
+    pub headers: Option<HashMap<String, String>>,
```
ladvoc, line 160: issue: Adding public fields to a non-exhaustive struct is technically a breaking change even if default is implemented, so the changeset level should be minor for this crate.

livekit-api/src/services/twirp_client.rs
```diff
@@ -91,29 +95,154 @@ impl TwirpClient {
...
+    /// Resolves the next untried region, fetching the region list lazily on the
+    /// first retryable failure and reusing it thereafter.
+    async fn next_region(
+        &self,
+        original: &Url,
+        forward: &HeaderMap,
+        region_urls: &mut Option<Vec<String>>,
+        attempted: &[String],
+    ) -> Option<Url> {
+        if region_urls.is_none() {
```
ladvoc: suggestion: Consider using pattern matching to avoid unwrap:
```rust
let region_urls = match region_urls {
    Some(urls) => urls,
    None => region_urls.insert(failover::region_urls(original, forward).await),
};
failover::pick_next(region_urls, attempted)
```

livekit-api/src/services/failover.rs
```diff
@@ -0,0 +1,236 @@
...
+    {
+        let _ = d;
+    }
+}
+
+/// Region discovery (`/settings/regions`) uses a short timeout so a slow or
+/// unreachable endpoint doesn't stall the failover path.
+const DISCOVERY_TIMEOUT: Duration = Duration::from_secs(2);
+
+struct CacheEntry {
```
ladvoc: suggestion: Region caching was recently implemented by @lukasIO in #1150 for the signaling client; we should consider reusing here if possible.

ladvoc (approved):

cloudwebrtc (approved):

.changeset/feat_auto_failover_apis_with_lk_cloud.md
```diff
@@ -0,0 +1,8 @@
+---
+livekit: patch
+livekit-api: patch
+livekit-ffi: patch
+livekit-uniffi: patch
```
xianshijing-lk: I don't see changes in livekit-ffi and livekit-uniffi folders, are they indirectly affected ? 

And should it be minor rather than patch ?

livekit-api/src/services/failover.rs
```diff
@@ -0,0 +1,236 @@
...
+/// async backend retries without delay (the failover path is short).
+pub(crate) async fn backoff_sleep(d: Duration) {
+    if d.is_zero() {
+        return;
+    }
+    #[cfg(feature = "services-tokio")]
+    {
+        tokio::time::sleep(d).await;
+    }
+    #[cfg(not(feature = "services-tokio"))]
```
xianshijing-lk: curiously, is tokio a must for production ? or optional ? 
I wonder if that will creates confusion or different behaviors if we support non-tokio usage

livekit-api/src/services/twirp_client.rs
```diff
@@ -91,29 +95,154 @@ impl TwirpClient {
...
+                        return Ok(R::decode(resp.bytes().await?)?);
+                    }
+                    // 4xx is terminal; only 5xx is retryable.
+                    if is_last || status.as_u16() < 500 {
+                        let err: TwirpErrorCode = resp.json().await?;
+                        return Err(TwirpError::Twirp(err));
+                    }
+                    match self.next_region(&original, &forward, &mut region_urls, &attempted).await
+                    {
+                        Some(next) => {
```
xianshijing-lk: this looks like we are unconditionally retrying?

For some transport errors ( like connection dropped, timeout), take some mutating apis as example, like a dropped connection does not prove the server skipped the operation; it may have created the room, started egress, ..etc.
Retrying those against another region can duplicate side effects ? 

If that is a real concern, should we restrict transport error retries to some idempotent/read methods ?

livekit-api/src/services/connector.rs
```diff
@@ -223,34 +248,52 @@ impl ConnectorClient {
...
+        let headers =
+            self.base.auth_header(VideoGrants { room_create: true, ..Default::default() }, None)?;
+
+        // Accept can block until the call is answered, so default the request
+        // timeout to the standard ring window; the caller overrides via
+        // `options.timeout` and should set it above the ringing_timeout passed to
+        // `dial_whatsapp_call`. Without waiting, the request returns promptly and
+        // the client default applies.
+        let timeout = if wait_until_answered {
+            Some(options.timeout.unwrap_or(DEFAULT_RINGING_TIMEOUT))
```
xianshijing-lk, line 281: should you honor ringing_timeout when computing request timeout here ?

livekit-api/src/services/failover.rs
```diff
@@ -0,0 +1,236 @@
...
+            if entry.fetched_at.elapsed() < entry.ttl {
+                return entry.urls.clone();
+            }
+        }
+    }
+
+    match fetch(base, headers).await {
+        Ok((urls, ttl)) => {
+            // A zero TTL (e.g. Cache-Control: max-age=0) means "do not cache".
+            if !ttl.is_zero() {
```
xianshijing-lk: curiously, for max-age=0, should we also clear the stale cache ?

livekit-api/src/services/twirp_client.rs
```diff
@@ -91,29 +95,154 @@ impl TwirpClient {
             pkg: pkg.to_owned(),
             prefix: prefix.unwrap_or(DEFAULT_PREFIX).to_owned(),
             client: http_client::Client::new(),
+            failover: FailoverConfig::default(),
+            request_timeout: failover::DEFAULT_REQUEST_TIMEOUT,
         }
     }
 
+    /// Enables or disables region failover (enabled by default). Failover only
```
jhugman, line 103: idiomatic: consider `derive_builder` here.

livekit-api/src/services/sip.rs
davidzhao: is there a way we could prevent this from becoming a breaking change each time we add fields?

livekit-api/src/services/failover.rs
```diff
@@ -0,0 +1,236 @@
...
+        let _ = d;
+    }
+}
+
+/// Region discovery (`/settings/regions`) uses a short timeout so a slow or
+/// unreachable endpoint doesn't stall the failover path.
+const DISCOVERY_TIMEOUT: Duration = Duration::from_secs(2);
+
+struct CacheEntry {
+    urls: Vec<String>,
```
jhugman: idiomatic: consider using a `Url`, or at least a newtype for `Url`.

livekit-api/src/services/twirp_client.rs
davidzhao: yes, we need to make sure the server APIs are idempotent. if a success is not returned, then it's unclear to the client if there has been side effects, so the only option is to retry.

livekit-api/src/services/failover.rs
davidzhao: i'm not sure that's needed. it's a tiny entry with a fixed number of entries

livekit-api/src/services/failover.rs
```diff
@@ -0,0 +1,236 @@
...
+    static CACHE: OnceLock<Mutex<HashMap<String, CacheEntry>>> = OnceLock::new();
+    CACHE.get_or_init(|| Mutex::new(HashMap::new()))
+}
+
+/// Returns the alternative region URLs for `base`, fetching `/settings/regions`
+/// if the cache is stale. Best-effort: on a fetch failure it serves a stale
+/// cached list when available, otherwise an empty list (the caller then stops
+/// failing over). Forwards the caller's headers so a valid token — and any test
+/// directives — reach the discovery endpoint.
+pub(crate) async fn region_urls(base: &Url, headers: &HeaderMap) -> Vec<String> {
```
jhugman, line 139: Oh we're already using `Url`!

livekit-api/src/services/sip.rs
```diff
@@ -435,49 +462,64 @@ impl SIPClient {
...
+            play_dialtone: options.play_dialtone.unwrap_or(false),
+            hide_phone_number: options.hide_phone_number.unwrap_or(false),
+            max_call_duration: Self::duration_to_proto(options.max_call_duration),
+            ringing_timeout: Self::duration_to_proto(ringing_timeout),
+            krisp_enabled: options.enable_krisp.unwrap_or(false),
+            headers: options.headers.unwrap_or_default(),
+            include_headers: options.include_headers.map(|h| h as i32).unwrap_or_default(),
+            media_encryption: options.media_encryption.map(|e| e as i32).unwrap_or_default(),
+            ..Default::default()
+        };
```
jhugman, line 493: Wow

livekit-api/src/services/twirp_client.rs
```diff
@@ -91,29 +95,154 @@ impl TwirpClient {
...
+                                current.host_str().unwrap_or_default(),
+                                err,
+                                next,
+                            );
+                            failover::backoff_sleep(self.backoff(attempt)).await;
+                            attempted.push(failover::host_key(&next));
+                            current = next;
+                        }
+                        None => return Err(err.into()),
+                    }
```
jhugman: This is _nearly_ the same code as above. Can we refactor this?

livekit-api/src/services/sip.rs
ladvoc: Yes, I think we should eventually move toward using the builder pattern in more places. With that approach, struct fields stay private, and when a new field is added, we expose a corresponding setter as an additive API change.

For example, `TwirpClient` currently follows this pattern:

```rust
/// Overrides the default per-attempt request timeout (10s) applied to calls
/// that don't pass their own. Each failover attempt gets the full budget.
pub fn with_request_timeout(mut self, timeout: Duration) -> Self {
    self.request_timeout = timeout;
    self
}
```

That said, applying this here would require a larger breaking change that is probably outside the scope of this PR. Still, I think it’s worth doing at some point.

livekit-api/src/services/failover.rs
davidzhao: that's a good point, probably better to require tokio in order to get proper delay before fallback

livekit-api/src/services/failover.rs
```diff
@@ -0,0 +1,236 @@
...
+    {
+        let c = cache().lock().unwrap();
+        if let Some(entry) = c.get(&key) {
+            if entry.fetched_at.elapsed() < entry.ttl {
+                return entry.urls.clone();
+            }
+        }
+    }
+
+    match fetch(base, headers).await {
```
jhugman, line 149: If another thread calls `region_urls` while a first one is fetching, it will redo the fetch. I'm not sure what the pattern of behaviour is (how many threads likely calling this, how big the payload etc), so I can't say if this is acceptable or not.

livekit-api/src/services/connector.rs
davidzhao: `ringing_timeout` is set in a different step, and not set here, so it'll not be possible for us to figure out what it is.

.changeset/feat_auto_failover_apis_with_lk_cloud.md
davidzhao: i think the refactor still touches livekit & livekit-ffi (due to signal client changes).. I'm not sure about uniffi, happy to remove that one

ladvoc: _livekit-uniffi_ isn't required for this change since it only exposes access token/verification functionality.

jhugman (commented):
Drive by review looking at idiomaticity. Feel free to ignore.

livekit-api/src/services/failover.rs
davidzhao: fixed

livekit-api/src/services/failover.rs
davidzhao: yeah good point. though the payload is tiny, and it's fine if they race (another thread repopulates it)

PR #1205, opened 2026-07-01
Turn single peerconnection off by default

cnderrauber:


livekit/src/room/mod.rs
```diff
@@ -434,7 +434,7 @@ impl Default for RoomOptions {
             },
             join_retries: 3,
             sdk_options: RoomSdkOptions::default(),
-            single_peer_connection: false,
+            single_peer_connection: true,
```
davidzhao, line 437: this does the reverse? is it a dupe?

davidzhao (approved):

livekit/src/room/mod.rs
cnderrauber: The title is wrong.. duplicate with https://github.com/livekit/rust-sdks/pull/1206

PR #1212, opened 2026-07-02
Make AdmProxy worker-thread-affine

hiroshihorie:
## What

Rewrites `AdmProxy` (webrtc-sys) so that every touch of the platform ADM, the synthetic ADM, and all proxy state happens on the WebRTC worker thread. Every public method marshals once at the boundary (`RunOnWorker`, a `BlockingCall` that runs inline when the caller is already the worker), all mutable state is worker-owned and annotated `RTC_GUARDED_BY(worker_thread_)`, and both mutexes are removed. Mode transitions are now plain sequential worker code.

On top of that, the platform ADM is now created lazily on the first `AcquirePlatformAdm()` call on all platforms, extending the model Android already used. Apps that never use `PlatformAudio` never construct the platform ADM, so creating a `PeerConnectionFactory` has no platform audio footprint.

## Why

Platform ADM implementations (Windows CoreAudio2, iOS, and the upcoming Apple AudioEngine device) bind a sequence checker to their construction thread and `RTC_DCHECK_RUN_ON` every control method. Previously the proxy called the platform ADM directly on whatever thread the Rust FFI caller was on, guarded only by a mutex, violating that contract and racing with WebRTC's own worker-thread calls. On Android this is more than a DCHECK violation: the lazily created ADM makes JNI calls, and a JNI call from an unattached caller thread hard-aborts the process (livekit/client-sdk-unity#368). This PR is also a prerequisite for switching the Apple platform ADM to the AudioEngine device (follow-up: #1215, building on the now-merged webrtc-sdk/webrtc#260).

Making the proxy thread-affine removes by construction the races a mutex couldn't prevent: stale mode snapshots interleaving transitions, the synthetic ADM being driven from caller threads, and `Terminate()` racing mode switches.

This mirrors upstream WebRTC's own pattern: `RTCPeerConnectionFactory` constructs the ADM inside `_workerThread->BlockingCall(...)`, and `WebRtcVoiceEngine`/`AudioState` guard every ADM call with worker-thread sequence checkers.

## ADM creation timing

Before this PR the platform ADM was created eagerly in the proxy constructor on every platform except Android (pre-existing behavior on main, though the old design doc incorrectly described a lazy lifecycle). It is now lazy everywhere:

- `EnsurePlatformAdmCreated()` runs on the first acquire, inside the worker-marshaled acquire path, so the fresh ADM binds its sequence checker to the worker thread exactly as an eagerly constructed one would. The worker-thread-affine design is what makes cross-platform laziness safe: on main, Android's lazy creation ran on whatever Rust thread called acquire.
- State configured before the first acquire is cached and replayed onto the fresh ADM: the `AudioTransport` WebRTC registers at factory init, device selection, and the channel configuration the voice engine requests via `SetStereoPlayout`/`SetStereoRecording`.
- Once created, the instance still persists until the proxy is destroyed. Release only switches back to synthetic mode. Re-creating the ADM caused iOS KVO races, so laziness applies only to the first creation.
- For synthetic-only apps this removes all platform audio plumbing at factory creation. With the default ADMs eager creation was already side-effect light (no AVAudioSession configuration, no observers, no mic prompt before `InitPlayout`/`InitRecording`), but the AudioEngine ADM arriving in #1215 registers an audio session delegate at construction, so deferring construction keeps synthetic-only apps fully untouched.

## Also fixed along the way

- **Android: lazily created ADM never received the audio transport.** WebRTC registers its `AudioTransport` at factory init, before the ADM exists. `EnsurePlatformAdmCreated` now passes it (and any pre-acquire device selection and channel config) to the fresh ADM. Previously platform recording delivered no audio to the pipeline.
- **Shutdown race:** `AudioDeviceController` now holds a `shared_ptr<RtcRuntime>` (same pattern as AudioTrack/VideoTrack) so the worker thread outlives every proxy reference reachable from Rust.
- Removed dead state: write-only `playout_initialized_`/`recording_initialized_` flags and never-read device GUID tracking.
- `docs/ADM_PROXY_DESIGN.md` rewritten to match the actual lifecycle and to document the threading model.

## Threading contract

- The proxy must be constructed on the worker thread (the factory already does this via `BlockingCall`). The synthetic ADM binds its sequence checker there, and the lazily created platform ADM binds there too via the marshaled acquire path.
- Destruction may happen on any thread. The destructor hops to the worker for teardown.
- Calls from WebRTC internals (AudioState, voice engine) are already on the worker and run inline, with no added hops on those paths.
- Rust FFI calls (control plane, low frequency) each pay one blocking hop to the worker, replacing the previous mutex acquisition.
- Realtime audio callbacks (`AudioTransport` data) are NOT marshaled. They flow directly through the sub-ADMs' own I/O threads, matching the platform ADMs' design (separate io/render thread checkers).

## Behavior parity

Apart from the marshaling, the lazy creation, and the fixes listed above, behavior is unchanged: all no-platform-ADM defaults, state-update ordering, and gate semantics match the previous implementation method-for-method. The pre-acquire defaults were already exercised in production on Android, which was lazy before this PR.

## Testing

- `cargo check -p webrtc-sys` and `cargo build -p livekit` pass.
- Reviewed method-by-method against the previous implementation for semantic drift, plus lifetime/deadlock analysis of the `BlockingCall` boundary (factory teardown ordering, dtor-on-any-thread, SyntheticAudioDevice task queue). Threading contract verified against the ADM implementations in webrtc-sdk (Windows/iOS/AudioEngine sequence checkers, voice engine and AudioState worker guards).
- **New runtime exerciser** (included in this PR): `cargo run -p livekit --example platform_audio`. Runs offline, no LiveKit server. Passing on macOS (arm64), now also exercising lazy creation on every teardown/reacquire cycle:
  - lifecycle, device enumeration/selection, hot-swap while recording, AEC/AGC/NS reconfiguration
  - 16 threads x 50 iterations of concurrent enumeration/getters/recording churn/acquire-release churn: 0 errors, no deadlock
  - 5 asserted full runtime teardown/reacquire cycles (factory + AdmProxy destroyed and rebuilt, half dropped mid-recording)
- Not covered offline: platform playout under a live room (`playing_ = true` switch branch) and DCHECK-level thread validation (release prebuilt compiles `RTC_DCHECK` out, needs a debug webrtc via `LK_CUSTOM_WEBRTC`).

### DIFF
--- a/docs/ADM_PROXY_DESIGN.md
+++ b/docs/ADM_PROXY_DESIGN.md
@@ -100,9 +100,11 @@ These two methods must coexist without interference.
                                  │                         │
                                  ▼                         ▼
                   ┌────────────────────────────────────────────┐
-                  │        Platform ADM (Lazy Init)            │
-                  │  Created when: AcquirePlatformAdm()        │
-                  │  Destroyed when: ref_count → 0             │
+                  │              Platform ADM                  │
+                  │  Created: eagerly at proxy construction    │
+                  │  (lazily on first acquire on Android)      │
+                  │  Active when: ref_count > 0                │
+                  │  Destroyed: with the proxy                 │
                   └──────────────────┬─────────────────────────┘
                                      │
                           ┌──────────┴──────────┐
@@ -180,26 +182,33 @@ These two methods must coexist without interference.
 3. **NativeAudioSource**: Existing API for manual audio frame pushing
 4. **external_audio_source.patch**: WebRTC patch to prevent audio mixing conflicts
 
-### Lazy Initialization + Reference Counting Pattern
+### Reference Counting Pattern
 
-The Platform ADM is **not** created at startup. Instead, it's created lazily when first needed:
+The Platform ADM instance exists for the proxy's whole lifetime, but stays idle until acquired. The ref count and gates decide whether it actually touches the hardware:
 
 ```cpp
 // adm_proxy.h
 class AdmProxy : public webrtc::AudioDeviceModule {
-  // Platform ADM is created lazily, NOT at startup
-  webrtc::scoped_refptr<webrtc::AudioDeviceModule> platform_adm_;
+  // All state is owned by the worker thread, no locks needed
 
-  // Reference count for Platform ADM lifecycle
-  int platform_adm_ref_count_ = 0;
+  // Synthetic ADM pumps the pipeline when platform playout is inactive
+  webrtc::scoped_refptr<SyntheticAudioDevice> synthetic_adm_
+      RTC_GUARDED_BY(worker_thread_);
+
+  // Platform ADM for real audio I/O
+  webrtc::scoped_refptr<webrtc::AudioDeviceModule> platform_adm_
+      RTC_GUARDED_BY(worker_thread_);
+
+  // Reference count for Platform ADM users
+  int platform_adm_ref_count_ RTC_GUARDED_BY(worker_thread_) = 0;
 
   // Gate controls whether microphone recording is active
   // Default: FALSE - NativeAudioSource works without interference
-  bool recording_enabled_ = false;
+  bool recording_enabled_ RTC_GUARDED_BY(worker_thread_) = false;
 
   // Gate controls whether playout goes through platform speakers
   // Default: FALSE - synthetic mode (FFI callbacks to application)
-  bool playout_enabled_ = false;
+  bool playout_enabled_ RTC_GUARDED_BY(worker_thread_) = false;
 };
 ```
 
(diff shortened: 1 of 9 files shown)


docs/ADM_PROXY_DESIGN.md
```diff
@@ -180,39 +182,66 @@ These two methods must coexist without interference.
...
 |--------|--------|
-| `AcquirePlatformAdm()` | Increments ref_count. Creates Platform ADM on first call. |
-| `ReleasePlatformAdm()` | Decrements ref_count. Terminates Platform ADM when count reaches 0. |
+| `AcquirePlatformAdm()` | Increments ref_count. Switches to platform mode on 0 → 1 (and creates the ADM on Android). |
+| `ReleasePlatformAdm()` | Decrements ref_count. Switches back to synthetic mode on 1 → 0. The Platform ADM instance stays alive until the proxy is destroyed. |
+
+**Creation timing:**
 
-**Why Lazy Initialization?**
+On most platforms the Platform ADM is created eagerly in the proxy constructor. iOS requires early AVAudioSession setup to avoid KVO race conditions, and keeping a single instance for the proxy's lifetime avoids the races that re-creation caused. Whether the platform hardware is actually used is controlled purely by the ref count and the gates below, so eager creation does not interfere with synthetic mode. Android is the exception, its ADM is created lazily on first acquire because JNI may not be initialized when the proxy is constructed.
```
xianshijing-lk, line 224: what will happen on IOS if PlatformAudio is not being used, and developers are using synthetic mode to play / capture the audio ? 
will it be a problem as our ADM proxy grab the AVAudioSession ?

MaxHeimbrock:
I think we need this to fix this issue reported by a customer:
https://github.com/livekit/client-sdk-unity/issues/368

docs/ADM_PROXY_DESIGN.md
MaxHeimbrock: I have tested with Unity on iOS, MacOS and Android both with and without Platform Audio and all look good. So this case was covered.

MaxHeimbrock (approved):
Works for me and looks good, thanks 👍

PR #1214, opened 2026-07-02
Dynacast: keep SVC tracks active while any quality is subscribed

MaxHeimbrock:
Stacked on #1213 (targets that branch; rebase to `main` once it merges).

## What

Ports the SVC special case from client-sdk-js that #1003 didn't bring over: for SVC codecs (VP9/AV1) all spatial layers ride in a single encoded stream and the SFU selects layers server-side. When a `SubscribedQualityUpdate` has **any** quality enabled, the whole encoding must stay active, instead of matching qualities per encoding (which can deactivate the track, e.g. if the SFU enables only a quality that doesn't map to the rid-less encoding).

JS reference: `setPublishingLayersForSender` in `livekit-client` — `if (isSVC) { if (qualities.some(q => q.enabled)) qualities.forEach(q => q.enabled = true); }`, with `isSVCCodec = vp9 || av1`.

## Changes

- `LocalVideoTrack::set_publishing_layers` takes `is_svc` and applies the any-enabled rule.
- `handle_subscribed_quality_update` derives `is_svc` from the publication's codec (VP9/AV1), mirroring `isSVCCodec` in JS.
- New e2e test `test_dynacast_svc` (VP9, `L3T3_KEY`): the single SVC encoding stays active across LOW/HIGH quality requests, deactivates when the last subscriber unsubscribes, and reactivates on resubscribe.
- `SolidColorTrack::publish_with_options` test helper for publishing with custom `TrackPublishOptions`.

## Testing

All dynacast e2e tests pass against a local `livekit-server --dev`:

```
test test_dynacast ... ok
test test_dynacast_multiple_subscribers_only_publish_requested_tracks ... ok
test test_dynacast_svc ... ok
```

Observed SFU updates in the new test confirm the flow: `vp9:[Low=true, Medium=false, High=false]` on a LOW request keeps the encoding ON; `[Low=false, Medium=false, High=false]` after unsubscribe turns it off.

### DIFF
--- a/livekit/src/room/mod.rs
+++ b/livekit/src/room/mod.rs
@@ -1928,10 +1928,17 @@ impl RoomSession {
             }
         };
 
+        let video_codec = publication.publish_options().video_codec;
+        // SVC codecs carry all spatial layers in one encoded stream.
+        let is_svc = matches!(
+            video_codec,
+            crate::options::VideoCodec::VP9 | crate::options::VideoCodec::AV1
+        );
+
         let qualities: Vec<proto::SubscribedQuality> = if !update.subscribed_codecs.is_empty() {
             // This is the requested codec, which we also advertise in simulcast_codecs and use
             // for sender codec preferences, so it should match the SFU's subscribed codec key.
-            let codec = publication.publish_options().video_codec.as_str().to_lowercase();
+            let codec = video_codec.as_str().to_lowercase();
             log::info!(
                 "dynacast: SFU quality update for {}: subscribed_codecs={:?}, looking for codec '{}'",
                 track_sid,
@@ -1980,7 +1987,7 @@ impl RoomSession {
             update.subscribed_qualities.clone()
         };
 
-        if let Err(e) = video_track.set_publishing_layers(&qualities) {
+        if let Err(e) = video_track.set_publishing_layers(&qualities, is_svc) {
             log::error!("dynacast: failed to set publishing layers for {}: {}", track_sid, e);
         }
     }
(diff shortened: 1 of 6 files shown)


livekit/tests/dynacast_test.rs
```diff
@@ -428,3 +428,110 @@ async fn test_dynacast_multiple_subscribers_only_publish_requested_tracks() -> R
...
+#[cfg(feature = "__lk-e2e-test")]
+#[test_log::test(tokio::test)]
+async fn test_dynacast_svc() -> Result<()> {
+    let mut pub_room_opts = RoomOptions::default();
+    pub_room_opts.dynacast = true;
+    let pub_options = TestRoomOptions { room: pub_room_opts, ..Default::default() };
+    let sub_options = TestRoomOptions::default();
+
+    let mut rooms = test_rooms_with_options([pub_options, sub_options]).await?;
+    let (pub_room, _pub_events) = rooms.remove(0);
```
ladvoc, line 449: nitpick: You can use `rooms.pop()` here.

livekit/src/room/mod.rs
```diff
@@ -1928,10 +1928,17 @@ impl RoomSession {
             }
         };
 
+        let video_codec = publication.publish_options().video_codec;
+        // SVC codecs carry all spatial layers in one encoded stream.
+        let is_svc = matches!(
```
ladvoc, line 1933: suggestion(non-blocking): Consider breaking this out into a `const` helper function.

ladvoc (approved):

PR #1218, opened 2026-07-04
feat: support building for 32-bit ARM Linux (linux-arm)

ibrahim-mubarak:
## What

Adds support for building the SDK for **32-bit ARM Linux** (`armv7-unknown-linux-gnueabihf`, i.e. `linux-arm`), targeting armv7-a + NEON hard-float as a baseline — which also covers armv8 cores running in AArch32 mode (e.g. Cortex-A32/A35 embedded SoCs).

There are no prebuilt `linux-arm` webrtc artifacts, so consumers build libwebrtc from source and point `LK_CUSTOM_WEBRTC` at it:

```sh
cd webrtc-sys/libwebrtc
./build_linux.sh --arch arm --profile release
# artifacts land in linux-arm-release/
```

## Changes

- **`webrtc-sys/libwebrtc/build_linux.sh`**
  - accept `--arch arm`; sets `arm_float_abi="hard"` and `arm_use_neon=true` (armv7+NEON baseline, same config as the other Linux builds otherwise — X11/PipeWire stay enabled, they are lazy-loaded/dlopen'd at runtime).
  - new optional `CLANG_BASE_PATH` env: Chromium only publishes x86-64 Linux clang binaries, so on other build hosts (e.g. aarch64) you can point this at a matching LLVM release; it is passed as `clang_base_path` to gn and its `llvm-objcopy` is used for the boringssl symbol prefixing step.
- **`webrtc-sys/build.rs`**
  - `add_gio_headers`: map `arm` → `debian_bullseye_armhf-sysroot` / `arm-linux-gnueabihf`.
  - `add_lazy_load_so`: select a new `arm-linux-gnueabihf` trampoline directory for `arm`.
- **`webrtc-sys/src/lazy_load_deps_for/desktop_capturer/arm-linux-gnueabihf/`**: Implib.so-generated trampolines for the desktop-capturer X11/drm/gbm libraries (generated with `implib-gen.py --target armhf` against the Debian bullseye armhf sysroot libs, same procedure as the existing x86_64/aarch64 files).
- **`yuv-sys/build.rs`**: mark `YuvConstants` as an opaque type for bindgen. Its SIMD vector members give it an alignment bindgen cannot mirror on 32-bit ARM (the generated layout test fails with align 2 vs 8); it is only ever passed by pointer from Rust.

## Testing

- Built libwebrtc `linux-arm-release` from source (webrtc-sdk/webrtc @ m144_release, commit `aaeeee80…751ef663` matching the `webrtc-51ef663` tag) on an aarch64 Fedora host with LLVM 22.1.8 via `CLANG_BASE_PATH`.
- `cargo build -p livekit-ffi --release --target armv7-unknown-linux-gnueabihf` succeeds against it (cross gcc 13.4, glibc 2.39 sysroot).
- Resulting `liblivekit_ffi.so` is a clean 32-bit ARM EABI5 hard-float object whose only hard deps are libc/libm/libstdc++/libgcc; verified `livekit::initialize()`/`shutdown()` round-trip through the FFI under qemu-arm against an embedded (Realtek Cortex-A32) sysroot.
- A companion PR to client-sdk-cpp adds the CMake-side cross-compilation support.

https://claude.ai/code/session_01L4fDRyTiTphQnidL2Myptq


ibrahim-mubarak:
Closing for now — will resubmit once ready.

PR #1220, opened 2026-07-06
introduce LiveKitAPI construct, added smoke tests

davidzhao:
- simpler constructor for interacting with various services
- ability to be used without api key/token, but with just a secret
- full smoke tests to ensure correct token permissions, request encoding
- added missing SIP APIs, achieves parity with other SDKs

### DIFF
--- a/livekit-api/Cargo.toml
+++ b/livekit-api/Cargo.toml
@@ -45,8 +45,8 @@ __signal-client-async-compatible = [
 ]
 
 
-services-tokio = ["dep:reqwest", "dep:tokio", "tokio/time"]
-services-async = ["dep:isahc", "dep:futures-timer"]
+services-tokio = ["dep:reqwest", "dep:tokio", "tokio/time", "dep:livekit-runtime", "livekit-runtime/tokio"]
+services-async = ["dep:isahc", "dep:livekit-runtime", "livekit-runtime/async"]
 access-token = ["dep:jsonwebtoken", "dep:hmac", "dep:signature"]
 webhooks = ["access-token", "dep:serde_json", "dep:base64"]
 
@@ -121,7 +121,7 @@ hmac = { version = "0.12", optional = true }
 signature = { version = "2", optional = true }
 
 # signal_client
-livekit-runtime = { workspace = true, optional = true}
+livekit-runtime = { workspace = true, optional = true }
 tokio-tungstenite = { version = "0.29", features = ["url"], optional = true }
 async-tungstenite = { version = "0.29", features = [ "async-std-runtime", "async-native-tls", "url"], optional = true }
 tokio = { workspace = true, default-features = false, features = ["sync", "macros", "signal", "io-util", "net"], optional = true }
@@ -134,8 +134,6 @@ bytes = { workspace = true, optional = true }
 http = "1.1"
 reqwest = { version = "0.12", default-features = false, features = [ "json" ], optional = true }
 isahc = { version = "1.7.2", default-features = false, features = [ "json", "text-decoding" ], optional = true }
-# Runtime-agnostic timer for failover backoff on the non-tokio (isahc) backend.
-futures-timer = { version = "3", optional = true }
 
 flate2 = { version = "1", optional = true }
 scopeguard = "1.2.0"
@@ -145,3 +143,5 @@ device-info = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "net", "time", "macros", "io-util"] }
+# Minimal executor to drive the runtime-agnostic `services-async` (isahc) tests.
+futures = "0.3"
--- a/livekit-api/src/services/connector.rs
+++ b/livekit-api/src/services/connector.rs
@@ -89,7 +89,7 @@ pub struct ConnectTwilioCallOptions {
 #[derive(Debug)]
 pub struct ConnectorClient {
     base: ServiceBase,
-    client: TwirpClient,
+    pub(crate) client: TwirpClient,
 }
 
 impl ConnectorClient {
--- a/livekit-api/src/services/connector.rs
+++ b/livekit-api/src/services/connector.rs
@@ -178,7 +178,34 @@ impl ConnectorClient {
             .map_err(Into::into)
     }
 
-    /// Disconnects a WhatsApp call
+    /// Disconnects a WhatsApp call initiated by the business.
+    ///
+    /// This is the `BusinessInitiated` case; use [`disconnect_whatsapp_call_with_reason`]
+    /// to disconnect with a different [`DisconnectReason`].
+    ///
+    /// [`disconnect_whatsapp_call_with_reason`]: Self::disconnect_whatsapp_call_with_reason
+    /// [`DisconnectReason`]: proto::disconnect_whats_app_call_request::DisconnectReason
+    ///
+    /// # Arguments
+    /// * `call_id` - Call ID sent by Meta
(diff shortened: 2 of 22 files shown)


livekit-api/src/services/connector.rs
```diff
@@ -63,14 +62,10 @@ pub struct AcceptWhatsAppCallOptions {
...
-    /// Optional - Wait for the call to be answered before returning. When set,
-    /// the request blocks until the call is answered or fails.
+    /// Optional - Wait until the inbound party joins before returning.
     pub wait_until_answered: Option<bool>,
-    /// Optional - Per-request timeout override. When `wait_until_answered` is
-    /// set, defaults to a longer value (dialing takes time) and is raised, if
-    /// needed, to stay above `ringing_timeout`; otherwise the client default.
+    /// Optional - Per-request timeout override. When `wait_until_answered` is set
+    /// it defaults to the standard ring window; otherwise the client default applies.
     pub timeout: Option<Duration>,
```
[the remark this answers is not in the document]
davidzhao: this was recently added and is incorrect

livekit-api/src/services/mod.rs
```diff
@@ -43,15 +46,18 @@ pub enum ServiceError {
...
 
 pub type ServiceResult<T> = Result<T, ServiceError>;
 
 struct ServiceBase {
     api_key: String,
     api_secret: String,
+    // When set, requests carry this token verbatim and grants are ignored; the
+    // caller (typically a browser client) signed it out of band.
+    token: Option<String>,
 }
```
1egoman: suggestion: Consider making this an enum with disctinct `api_key`/`api_secret` and `token` cases.

Maybe something like:
```suggestion
enum ServiceBase {
    ApiKeySecret {
        api_key: String,
        api_secret: String,
    },
    Token(String),
}
```

livekit-api/src/services/mod.rs
```diff
@@ -62,23 +68,106 @@ impl Debug for ServiceBase {
...
         sip: Option<SIPGrants>,
     ) -> Result<HeaderMap, AccessTokenError> {
-        let mut tok =
-            AccessToken::with_api_key(&self.api_key, &self.api_secret).with_grants(grants);
-        if let Some(sip_grants) = sip {
-            tok = tok.with_sip_grants(sip_grants);
-        }
-        let token = tok.to_jwt()?;
+        let token = if let Some(token) = &self.token {
+            token.clone()
```
1egoman: suggestion: [same comment](https://github.com/livekit/node-sdks/pull/695#discussion_r3529281441) as I left on the node one, it might be nice to have an "token expired" event here and some way a user can regenerate the token before continuing with making the request.

livekit-api/src/services/twirp_client.rs
```diff
@@ -27,26 +27,61 @@ use crate::http_client;
...
+pub enum ServerError {
     #[cfg(feature = "services-tokio")]
     #[error("failed to execute the request: {0}")]
     Request(#[from] reqwest::Error),
     #[cfg(feature = "services-async")]
     #[error("failed to execute the request: {0}")]
     Request(#[from] std::io::Error),
-    #[error("twirp error: {0}")]
+    #[error("server error: {0}")]
     Twirp(TwirpErrorCode),
```
1egoman: question: How bad an idea would be to make the breaking change and migrate `Twirp` => `Server` here? There's also a few other places in the diff where I noticed a `Twirp` enum name was kept.

I am mostly asking because in the past there's been a bit less emphasis on perfect backwards compatibility in the rust sdk (ie, adding new struct fields) and I'd think this would be a somewhat confusing remnant that would be nice to clean up.

livekit-api/Cargo.toml
```diff
@@ -145,3 +143,5 @@ device-info = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "net", "time", "macros", "io-util"] }
+# Minimal executor to drive the runtime-agnostic `services-async` (isahc) tests.
+futures = "0.3"
```
1egoman, line 147: thought: Is there a benefit to this extra test specific dependency, or would it be just worth using `tokio`?

davidzhao: as long as we are supporting the async runtime, I think we'd need a way to test that path without tokio

livekit-api/src/services/mod.rs
davidzhao: I don't know that this makes a difference.. the server will be the authority here and it'll specify all the reasons why the token is incorrect.

client-side clocks could be desynced, so it's possible for the client to think it's not expired, but for the server to indicate that it is.. or vice versa.

livekit-api/src/services/livekit_api.rs
```diff
@@ -0,0 +1,141 @@
...
+/// A single entry point to every LiveKit server API, exposing each service
+/// through an accessor (`room()`, `egress()`, `ingress()`, `sip()`,
+/// `agent_dispatch()`, `connector()`).
+///
+/// Construct it with an API key and secret ([`with_api_key`](Self::with_api_key),
+/// or [`new`](Self::new) to read them from the environment) for backend use, or
+/// with a pre-signed token ([`with_token`](Self::with_token)) for client-side use
+/// where the API secret must not be exposed.
+#[derive(Debug)]
+pub struct LiveKitApi {
```
ladvoc, line 36: note: The way this is currently setup, each inner API client owns its own HTTP client which means each gets its own connection pool; ideally, when constructed together through `LiveKitApi`, they should reference the same HTTP client:

From the reqwest [docs](https://docs.rs/reqwest/0.12.28/reqwest/struct.Client.html):
> The Client holds a connection pool internally to improve performance by reusing connections and avoiding setup overhead, so it is advised that you create one and reuse it.

Isahc has a similar [note](https://docs.rs/isahc/latest/isahc/struct.HttpClient.html).

livekit-api/src/services/twirp_client.rs
davidzhao: done

livekit-api/src/services/livekit_api.rs
davidzhao: great point! I'll make this change

davidzhao:
@ladvoc @1egoman PTAL

livekit-api/src/services/livekit_api.rs
```diff
@@ -63,14 +66,28 @@ impl LiveKitApi {
...
+    /// Points every service at one shared HTTP client so they reuse a single
+    /// connection pool instead of each opening its own.
+    fn share_http_client(&mut self) {
+        let http = http_client::Client::new();
+        self.room.client.set_http_client(http.clone());
+        self.egress.client.set_http_client(http.clone());
+        self.ingress.client.set_http_client(http.clone());
+        self.sip.client.set_http_client(http.clone());
+        self.agent_dispatch.client.set_http_client(http.clone());
+        self.connector.client.set_http_client(http);
```
1egoman: nitpick: It might be cleaner to just pass in the http client at construction time rather than setting it after the fact? So something like `::with_token_and_client(token, client)` / `::with_api_key_client(key, secret, client)` (and the old constructors defaulting to making their own http clients for backwards compatibility) for downstream structs.

Maybe I'm missing something though and that is nonviable here for some reason.

1egoman (approved):

livekit-api/src/services/livekit_api.rs
davidzhao: true, I was just trying to not touch the constructors.. handled it with a build function

PR #1223, opened 2026-07-06
Add pre-encoded video passthrough publish path

chenosaurus:
This PR contains core SDK changes that add the ability for the Rust-SDK to ingest pre-encoded video frames.  This is only the core plumbing to allow these frames to be `captured` and passed thru a passthru encoder.  

- **webrtc-sys**
  - New `passthrough_video_encoder.{h,cpp}`: a `webrtc::VideoEncoder` that passes pre-encoded H264/H265/VP8/VP9/AV1 access units through without re-encoding, forwarding WebRTC rate-control targets and keyframe requests back to the encoded source.
  - New `encoded_video_frame_buffer.{h,cpp}`: carries encoded payloads through the `VideoFrameBuffer` pipeline.
  - `video_track.{h,cpp,rs}`, `video_encoder_factory.cpp`, `rtp_sender.cpp`: plumbing to capture encoded frames into a source and select the passthrough encoder.
  - `jetson/jetson_av1_bitstream.{h,cpp}` now compiles on **all** platforms (was Jetson-only) for AV1 OBU validation on ingest; small fixes in the Jetson AV1 encoder path.
- **libwebrtc**: new public API — `EncodedVideoFrame`, `EncodedVideoCodec`, `EncodedFrameType`, `EncodedRateControl`, `VideoEncoderBackend::PreEncoded`, `capture_encoded_frame` on the native video source.

### DIFF
--- a/.changeset/preencoded-passthrough-core.md
+++ b/.changeset/preencoded-passthrough-core.md
@@ -1,0 +1,7 @@
+---
+"webrtc-sys": patch
+"libwebrtc": patch
+"livekit": patch
+---
+
+Add a pre-encoded video publish path: a passthrough video encoder and encoded video frame buffer in webrtc-sys, and `EncodedVideoFrame`/`EncodedVideoCodec`/`EncodedFrameType` publish APIs with a `VideoEncoderBackend::PreEncoded` backend in libwebrtc. WebRTC rate-control targets and keyframe requests are forwarded to encoded sources, and pre-encoded AV1 and H265 access units are validated on ingest.
--- a/.changeset/preencoded-passthrough-core.md
+++ b/.changeset/preencoded-passthrough-core.md
@@ -1,7 +1,7 @@
 ---
-"webrtc-sys": patch
-"libwebrtc": patch
-"livekit": patch
+"webrtc-sys": minor
+"libwebrtc": minor
+"livekit": minor
 ---
 
 Add a pre-encoded video publish path: a passthrough video encoder and encoded video frame buffer in webrtc-sys, and `EncodedVideoFrame`/`EncodedVideoCodec`/`EncodedFrameType` publish APIs with a `VideoEncoderBackend::PreEncoded` backend in libwebrtc. WebRTC rate-control targets and keyframe requests are forwarded to encoded sources, and pre-encoded AV1 and H265 access units are validated on ingest.
--- a/libwebrtc/src/native/video_frame.rs
+++ b/libwebrtc/src/native/video_frame.rs
@@ -51,9 +51,7 @@ pub fn new_video_frame_buffer(
             vfb_sys::ffi::VideoFrameBufferType::NV12 => Box::new(vf::NV12Buffer {
                 handle: NV12Buffer { sys_handle: sys_handle.pin_mut().get_nv12() },
             }),
-            _ => {
-                Box::new(vf::I420Buffer { handle: I420Buffer { sys_handle: sys_handle.to_i420() } })
-            }
+            _ => unreachable!(),
         }
     }
 }
--- a/libwebrtc/src/native/video_frame.rs
+++ b/libwebrtc/src/native/video_frame.rs
@@ -51,7 +51,9 @@ pub fn new_video_frame_buffer(
             vfb_sys::ffi::VideoFrameBufferType::NV12 => Box::new(vf::NV12Buffer {
                 handle: NV12Buffer { sys_handle: sys_handle.pin_mut().get_nv12() },
             }),
-            _ => unreachable!(),
+            _ => {
+                Box::new(vf::I420Buffer { handle: I420Buffer { sys_handle: sys_handle.to_i420() } })
+            }
         }
     }
 }
--- a/libwebrtc/src/native/video_source.rs
+++ b/libwebrtc/src/native/video_source.rs
@@ -13,13 +13,15 @@
 // limitations under the License.
 
 use std::{
-    sync::Arc,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 use cxx::SharedPtr;
 use livekit_runtime::interval;
-use parking_lot::Mutex;
 use webrtc_sys::{video_frame as vf_sys, video_frame::ffi::VideoRotation, video_track as vt_sys};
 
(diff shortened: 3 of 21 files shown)


.changeset/preencoded-passthrough-core.md
```diff
@@ -0,0 +1,7 @@
+---
+"webrtc-sys": patch
+"libwebrtc": patch
+"livekit": patch
```
stephen-derosa: FMU: why is this a patch and not a feature?

libwebrtc/src/video_frame.rs
```diff
@@ -71,6 +71,73 @@ pub struct FrameMetadata {
...
+    /// Capture timestamp in microseconds.
+    pub timestamp_us: i64,
+    /// Encoded frame type.
+    pub frame_type: EncodedFrameType,
+    /// Encoded frame width in pixels.
+    pub width: u32,
+    /// Encoded frame height in pixels.
+    pub height: u32,
+    /// Optional metadata to attach through packet trailers.
+    pub frame_metadata: Option<FrameMetadata>,
```
stephen-derosa, line 113: nice!

webrtc-sys/include/livekit/encoded_video_frame_buffer.h
```diff
@@ -0,0 +1,119 @@
...
+  bool has_request = false;
+  uint64_t target_bitrate_bps = 0;
+  double framerate_fps = 0.0;
+};
+
+// Latest-wins rate-control mailbox shared between the pass-through encoder and
+// the Rust capture side.
+class EncodedRateControlState {
+ public:
+  void Store(uint64_t target_bitrate_bps, double framerate_fps);
```
stephen-derosa, line 52: does the state live independent of the request? If not, consider just using the constructor instead of a `Store` function

stephen-derosa (approved):
looks good to me, no major feedback. Definitely would appreciate some examples/tests here!

libwebrtc/src/native/video_frame.rs
```diff
@@ -51,7 +51,9 @@ pub fn new_video_frame_buffer(
             vfb_sys::ffi::VideoFrameBufferType::NV12 => Box::new(vf::NV12Buffer {
                 handle: NV12Buffer { sys_handle: sys_handle.pin_mut().get_nv12() },
             }),
-            _ => unreachable!(),
+            _ => {
```
ladvoc: question: Is this safe?

libwebrtc/src/native/video_source.rs
```diff
@@ -139,6 +159,60 @@ impl NativeVideoSource {
...
+        };
+
+        let capture_ts = if frame.timestamp_us == 0 {
+            let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+            now.as_micros() as i64
+        } else {
+            frame.timestamp_us
+        };
+
+        self.inner.lock().captured_frames += 1;
```
ladvoc: suggestion(non-blocking): `captured_frames` is a scalar and the only field in `VideoSourceInner`, so it would benefit performance wise from using an atomic instead of a mutex, especially given this is a high-frequency path. Note: this was not introduced in this PR.

libwebrtc/src/video_frame.rs
```diff
@@ -71,6 +71,73 @@ pub struct FrameMetadata {
...
+    /// Encoded video codec.
+    pub codec: EncodedVideoCodec,
+    /// Encoded access-unit payload.
+    pub payload: &'a [u8],
+    /// Capture timestamp in microseconds.
+    pub timestamp_us: i64,
+    /// Encoded frame type.
+    pub frame_type: EncodedFrameType,
+    /// Encoded frame width in pixels.
+    pub width: u32,
```
ladvoc: suggestion: Use `VideoResolution` here instead of discrete width/height fields.

libwebrtc/src/video_source.rs
```diff
@@ -75,10 +84,33 @@ pub mod native {
...
+        pub fn new_encoded(resolution: VideoResolution) -> Self {
+            Self { handle: vs_imp::NativeVideoSource::new_encoded(resolution) }
+        }
+
         pub fn capture_frame<T: AsRef<dyn VideoBuffer>>(&self, frame: &VideoFrame<T>) {
             self.handle.capture_frame(frame)
         }
 
+        /// Captures one pre-encoded video access unit.
+        pub fn capture_encoded_frame(&self, frame: &EncodedVideoFrame<'_>) -> bool {
```
ladvoc, line 98: suggestion(non-blocking): I am assuming these new methods should only ever be called if the video source is pre-encoded and `capture_frame` shouldn't be? If so, this is a good place to use the type-state pattern to model this at the API level:

```rust
// Marker-types
struct Encoded;
struct PreEncoded;

pub struct NativeVideoSource<T = Encoded> {
    pub(crate) handle: vs_imp::NativeVideoSource,
    marker: PhantomData<T>
}

impl NativeVideoSource<Encoded> {
    pub fn new(resolution: VideoResolution, is_screencast: bool) -> Self {
        // ...
    }

    pub fn capture_frame<T: AsRef<dyn VideoBuffer>>(&self, frame: &VideoFrame<T>) {
        // ...
    }
}

impl NativeVideoSource<PreEncoded> {
    pub fn new_encoded(resolution: VideoResolution) -> Self {
        // ...
    }

    pub fn capture_encoded_frame(&self, frame: &EncodedVideoFrame<'_>) -> bool {
        // ...
    }
}

```

Example:
```rust
let a = NativeVideoSource::new(resolution, false); // NativeVideoSource<Encoded>
a.capture_frame(/* ... */);
a.capture_encoded_frame(/* ... */); // ERROR: this method is not defined in this context

let b = NativeVideoSource::new_encoded(resolution); // NativeVideoSource<PreEncoded>
b.capture_frame(/* ... */); // ERROR: this method is not defined in this context
b.capture_encoded_frame(/* ... */); 
```

ladvoc (approved):

.changeset/preencoded-passthrough-core.md
chenosaurus: good call, this should be a minor bump since it does add new APIs

libwebrtc/src/native/video_frame.rs
chenosaurus: reverted this change

cloudwebrtc (approved):
lgtm!

PR #1224, opened 2026-07-07
fix: don't log expected publisher data channel close as unexpected

longcw:
Fixes the spurious ERROR logs reported in livekit/agents#6250.

When the server tears a session down (e.g. the room is deleted) it closes the publisher data channels from the remote side before the engine sets its `closed` flag, so the close was logged at ERROR as "publisher data channel closed unexpectedly" even though it was expected.

This records the terminal disconnect (server `Leave{Disconnect}`) so the close it triggers isn't logged as unexpected. Genuine unexpected closes (e.g. an oversized message silently killing the channel) still log, and Resume/Reconnect are unaffected.

### DIFF
--- a/.changeset/fix_publisher_dc_closed_unexpectedly.md
+++ b/.changeset/fix_publisher_dc_closed_unexpectedly.md
@@ -1,0 +1,6 @@
+---
+livekit: patch
+livekit-ffi: patch
+---
+
+Don't log an expected publisher data channel close as unexpected - #1224 (@longcw)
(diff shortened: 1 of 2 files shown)


davidzhao (commented):
we should also set this bit when the user disconnects intentionally.. so [close](https://github.com/livekit/rust-sdks/blob/main/livekit/src/rtc_engine/rtc_session.rs#L1887) would have a reason of ClientInitiated.

longcw:
> we should also set this bit when the user disconnects intentionally.. so [close](https://github.com/livekit/rust-sdks/blob/main/livekit/src/rtc_engine/rtc_session.rs#L1887) would have a reason of ClientInitiated.

user disconnects using `close()` already set `closed` that prevents the error log
```rust
                if !inner.closed.load(Ordering::Acquire)
                    && !inner.disconnecting.load(Ordering::Acquire)
                    && inner.publisher_pc.is_connected()
                {
                    log::error!("publisher data channel '{}' closed unexpectedly", label);
                }
```

theomonnom (approved):

PR #1192, opened 2026-06-26
Data streams v2

1egoman:
### Overview

This change introduces data streams v2, a set of major updates to data streams to add three things:
1. Single packet data streams for short (< 15kb), finite length payloads
2. DEFLATE compression, when both the sender and all receivers support compression/decompression
3. A maximum 15kb size for data stream attributes to close this as a DOS vector

For information about important decisions made in data streams v2, [see the spec](https://github.com/livekit/client-sdk-js/blob/data-streams-v2/DATA_STREAMS_SPEC.md). In particular, parts 1-5 provide a good high level overview of data streams v1 + v2, what these protocol improvements are aiming to acheive, and why certain choices were made when designing the protocol.

Also maybe of interest is [the corresponding web sdk pull request](https://github.com/livekit/client-sdk-js/pull/1985).

These updates are fully backwards compatible - the sdk will fall back to data streams v1 whenever an old participant is in the room.

There is also a significant amount of new test coverage to data streams (both v1 and v2). There were some small bugs I found along the way which I fixed as well.

PR #1227, opened 2026-07-07
Move data streams into their own crate

1egoman:
Breaks out a subset of the changes from https://github.com/livekit/rust-sdks/pull/1192 into a pull request which exclusively moves the existing data streams implementation into its own crate, `livekit-data-streams`, and adds a new `livekit-common` crate (kept fairly minimal for now) for common utilities which `livekit-data-streams` and `livekit` both need. 

This is largely being scoped to a "lift and shift" type effort, more extensive refactoring of the `livekit-data-streams` crate will be done in https://github.com/livekit/rust-sdks/pull/1192.

### Breaking changes

There should not be any breaking changes from the perspective of `livekit`'s public api. If so, please surface them.

### Testing

This pull request adds the relevant v1 data streams tests from https://github.com/livekit/rust-sdks/pull/1192. More test coverage will be introduced as a follow up in the data streams v2 change.


livekit-data-stream/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+mod utf8_chunk;
+
+mod incoming;
+pub use incoming::{
+    AnyStreamReader, ByteStreamReader, IncomingStreamManager, StreamReader, TextStreamReader,
+};
+
+mod outgoing;
+pub use outgoing::*;
```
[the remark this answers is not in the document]
1egoman: This is by design.

livekit/src/utils/take_cell.rs
```diff
@@ -26,18 +26,6 @@ impl<T> TakeCell<T> {
         Self { value: Arc::new(RwLock::new(Some(value))) }
     }
 
-    /// Take ownership of the value in the cell if it matches some predicate.
-    ///
-    /// This method will only take the value if the provided predicate returns `true` when called with the current value.
```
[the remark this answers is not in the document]
1egoman: This was also by design, there are no callers of `TakeCell::take_if` / `TakeCell::take_if_raw`.

livekit-data-stream/src/incoming.rs
```diff
@@ -253,9 +239,18 @@ struct ManagerInner {
 }
 
 impl IncomingStreamManager {
-    pub fn new() -> (Self, UnboundedReceiver<(AnyStreamReader, String)>) {
+    pub fn new(
+        reserved_topics: Vec<String>,
+    ) -> (Self, UnboundedReceiver<(AnyStreamReader, String)>) {
```
1egoman, line 244: Notable change: Instead of looking at a well known global, `IncomingStreamManager` now takes a list of `reserved_topics` so that upstream layers can pass down a list of reserved topics in a decoupled fashion.

livekit/src/room/data_stream/mod.rs
```diff
@@ -252,12 +254,26 @@ pub(crate) enum AnyStreamInfo {
...
+        }
+    }
+
+    pub fn encryption_type(&self) -> EncryptionType {
+        match self {
+            Self::Byte(info) => info.encryption_type(),
+            Self::Text(info) => info.encryption_type(),
+        }
+    }
 }
```
1egoman, line 261: Note to reviewers: I have dropped `libwebrtc::enum_dispatch` and written the implementation myself to avoid making `libwebrtc` a dependency of `livekit-data-streams`.

livekit/src/utils/take_cell.rs
```diff
@@ -26,18 +26,6 @@ impl<T> TakeCell<T> {
...
-    /// If the predicate returns `false` or the value has already been taken, this method returns `None`.
-    pub(crate) fn take_if_raw(&self, predicate: impl FnOnce(&T) -> bool) -> Option<T> {
-        if self.value.read().as_ref().map_or(false, |v| predicate(v)) {
-            self.take()
-        } else {
-            None
-        }
-    }
-
     /// Take ownership of the value in the cell. If the value has,
```
1egoman, line 41: Note to reviewers - this was dead code.

livekit-data-stream/src/lib.rs
```diff
@@ -13,27 +13,29 @@
...
+/// The stream managers only need to know that a send failed (they map it to
+/// [`StreamError::SendFailed`]); the concrete engine error type stays in the `livekit` crate,
+/// which bridges the outgoing packet channel to the RTC engine.
+#[derive(Debug, Clone)]
+pub struct SendError;
+
+/// Generates a random stream identifier (UUID v4).
+pub(crate) fn create_random_uuid() -> String {
+    uuid::Uuid::new_v4().to_string()
 }
```
1egoman, line 46: Note to reviewers: previously, this used `libwebrtc::native::create_random_uuid`. I didn't want to make `libwebrtc` a dependency of this new crate so I swapped this out for `uuid::Uuid`. I think this should be fine but please surface if there was a good reason `create_random_uuid` was being used instead to generate data stream ids here.

livekit-data-stream/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[features]
+# End-to-end testing hooks (exposes is_compressed/is_inline on stream info). Forwarded from
+# the `livekit` crate's `__lk-e2e-test` feature.
+__e2e-test = []
+# Exposes constructors used by downstream crates' test suites (e.g. `TextStreamReader::new_for_test`).
+test-utils = []
```
1egoman, line 13: Note to reviewers: should these be separate features or just one?

livekit/src/room/data_stream/mod.rs
ladvoc: How about moving `enum_dispatch` to `livekit-common`?

livekit-data-stream/src/lib.rs
ladvoc: I think the only reason this was used originally was to avoid another dependency, but given the architectural direction, I think it is ok to pull in `uuid` which is lightweight.

livekit/src/room/data_stream/mod.rs
1egoman: Ah, yea, I could do that - done in 433b85d.

I've kept the libwebrtc copy around so that it isn't coupled to `livekit-common`, IMO the goal should be to migrate all users to `livekit_common::enum_dispatch` and then drop the libwebrtc version.

livekit-data-stream/src/lib.rs
1egoman: Sounds good!

livekit-data-stream/Cargo.toml
ladvoc: If we need to keep them, they should be separate (I plan to eventually drop the special feature for E2E testing and instead just skip E2E tests based on env).

However, there's the bigger question of if we should be using non-public methods in integration tests in the first place.

1egoman: > However, there's the bigger question of if we should be using non-public methods in integration tests in the first place.

Right now, that's how it works so I don't think this makes it worse. IMO I'd rather decouple that conversation and revisit as part of the data streams v2 pull request since I will be making some more extensive updates there.

Cargo.toml
```diff
@@ -4,6 +4,8 @@ members = [
     "livekit",
     "livekit-api",
     "livekit-protocol",
+    "livekit-common",
+    "livekit-data-stream",
```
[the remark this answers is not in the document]
1egoman: Fixed.

livekit-data-stream/src/lib.rs
```diff
@@ -13,27 +13,34 @@
 // limitations under the License.
 
 use chrono::{DateTime, Utc};
-use libwebrtc::enum_dispatch;
+use livekit_common::EncryptionType;
 use livekit_protocol::data_stream as proto;
 use std::collections::HashMap;
 use thiserror::Error;
```
ladvoc, line 22: suggestion: Include _README_ as doc string so its contents shows up in generated docs:

```suggestion
#![doc = include_str!("../README.md")]
```

livekit-data-stream/src/incoming.rs
```diff
@@ -382,3 +377,205 @@ impl ManagerInner {
...
+            SENDER.to_string(),
+            EncType::None,
+        );
+        let (reader, identity) = rx.recv().await.expect("a reader should be dispatched");
+        assert_eq!(identity, SENDER);
+        assert_eq!(text_info(&reader).attributes.get("foo"), Some(&"bar".to_string()));
+        mgr.handle_chunk(chunk("s1", 0, text.as_bytes().to_vec()), EncType::None);
+        mgr.handle_trailer(trailer("s1"));
+        assert_eq!(read_text(reader).await.unwrap(), text);
+    }
```
1egoman, line 497: Notable improvement: I've added in an initial set of data streams v1 unit tests which exercises some happy paths in the data streams outgoing manager and data streams incoming manager. These will be further developed as part of the data streams v2 pull request.

ladvoc:
suggestion: To prevent future abuse of the _livekit-common_ crate, it might be a good idea to describe it in _AGENTS.md_ (or add a new one scoped to the crate) laying out what belongs here and what doesn't (e.g., lightweight functions, types, etc.).

livekit/src/room/mod.rs
```diff
@@ -2298,16 +2300,30 @@ async fn incoming_data_stream_task(
...
 
+/// Data stream topics reserved for internal SDK use (e.g. RPC). Events for these topics are
+/// handled within the `livekit` crate and never surfaced through `RoomEvent`; the list is also
+/// passed to `IncomingStreamManager` so it can flag internal streams.
+const INTERNAL_DATA_STREAM_TOPICS: &[&str] = &[rpc::RPC_REQUEST_TOPIC, rpc::RPC_RESPONSE_TOPIC];
+
+fn is_internal_topic(topic: &str) -> bool {
+    INTERNAL_DATA_STREAM_TOPICS.contains(&topic)
+}
+
```
1egoman, line 2311: Note to reviewers: migrated from existing code, [here is the original](https://github.com/livekit/rust-sdks/blob/3dafb1d8e8fc49b907a7c4a74cf342dfcbb24c4f/livekit/src/room/data_stream/mod.rs#L30-L37)

1egoman:
> suggestion: To prevent future abuse of the livekit-common crate, it might be a good idea to describe it in AGENTS.md (or add a new one scoped to the crate) laying out what belongs here and what doesn't (e.g., lightweight functions, types, etc.).

Done in 4445b08ad

livekit-data-stream/Cargo.toml
1egoman: > If we need to keep them, they should be separate (I plan to eventually drop the special feature for E2E testing and instead just skip E2E tests based on env).

I've combined them into one feature for now, `test-utils`.

livekit-common/src/enum_dispatch.rs
```diff
@@ -0,0 +1,56 @@
...
+            $crate::enum_dispatch!(@match [$($variant),+]: $fnc, $self, ($($arg,)*))
+        }
+    };
+
+    ($variants:tt; $($vis:vis fn $fnc:ident$args:tt -> $ret:ty;)+) => {
+        $(
+            $crate::enum_dispatch!(@fnc $variants: $vis fn $fnc$args -> $ret);
+        )+
+    };
+}
```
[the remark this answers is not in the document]
1egoman: This is by design.

livekit-common/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+// -------------------------------------------------------------------------------------------------
+
+/// Legacy client.
+pub const CLIENT_PROTOCOL_DEFAULT: i32 = 0;
+
+/// RPC v2 (see RPC spec).
+pub const CLIENT_PROTOCOL_DATA_STREAM_RPC: i32 = 1;
+
+// -------------------------------------------------------------------------------------------------
+// ParticipantIdentity
```
ladvoc, line 35: suggestion(non-blocking): Consider keeping top-level `lib.rs` just for declaring mods and exports and moving all declarations into submodules. A few ideas for organization:
- We will probably have more identifier types live here eventually, so `ParticipantIdentity` could be moved to an `identifiers` module.
- Consts can all go under a `consts` module (this is a convention I have seen before).

livekit-data-stream/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[dependencies]
+livekit-common = { workspace = true }
+livekit-protocol = { workspace = true }
+log = { workspace = true }
+thiserror = { workspace = true }
+parking_lot = { workspace = true }
+bytes = { workspace = true }
+tokio = { workspace = true, default-features = false, features = ["sync", "fs", "io-util", "rt"] }
```
ladvoc, line 22: note: Outside the scope of this PR and a pre-existing issue, but we should only depend on _tokio_ with the _sync_ feature unconditionally.

ladvoc (approved):

xianshijing-lk (approved):
great work, lgtm

livekit-common/src/lib.rs
1egoman: I have already done this as a follow up in the `data-streams-v2` branch, I wanted to keep this more 1:1 so it was easier to review. Let's discuss it more in the review for that one shortly!

livekit-data-stream/Cargo.toml
1egoman: Good to know, I will investigate in the `data-streams-v2` follow up pull request.

PR #1228, opened 2026-07-08
Handle data track SID reassignment

ladvoc:
Closes BOT-456

### DIFF
--- a/livekit-datatrack/src/remote/manager.rs
+++ b/livekit-datatrack/src/remote/manager.rs
@@ -243,7 +243,7 @@ impl Manager {
         info: &DataTrackInfo,
     ) -> bool {
         // Publisher identity and pub handle are stable across republications.
-        let Some((old_sid, desc)) = self.descriptors.iter().find(|(_, desc)| {
+        let Some((old_sid, descriptor)) = self.descriptors.iter().find(|(_, desc)| {
             desc.publisher_identity.as_ref() == publisher_identity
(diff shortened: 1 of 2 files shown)


stephen-derosa (approved):
lgtm -- nice tests!

livekit-datatrack/src/remote/manager.rs
```diff
@@ -229,6 +232,59 @@ impl Manager {
...
+        }
+        let old_sid = old_sid.clone();
+
+        let new_sid = info.sid();
+        log::debug!("SID reassigned: {} -> {}", old_sid, new_sid);
+
+        let Some(descriptor) = self.descriptors.remove(&old_sid) else {
+            return false;
+        };
+        *descriptor.info.sid.write().unwrap() = new_sid.clone();
```
1egoman, line 268: Just want to call out, in case this makes you more confident that this is correct - I did something very similar to this in a recent [node sdk change](https://github.com/livekit/node-sdks/pull/671) (also ported this [to python](https://github.com/livekit/python-sdks/pull/679)) for regular tracks to ensure that upon republish just the sid is updated: https://github.com/livekit/node-sdks/blob/e4a769c9717e6657f03ef44528640c2fb74973a8/packages/livekit-rtc/src/room.ts#L607-L615

1egoman (approved):
Generally makes sense to me!

PR #1229, opened 2026-07-08
Refactor http and websocket implementations into its own crate

jhugman:
This PR adds a `livekit-net` crate which is for HTTP and WebSocket traffic. i.e. signalling shaped tasks.

SDKs that want to implement their own networking can call the `setPlatformTransport` function with their own foreign trait implementation.

It needs a Parallel Change: 
- [ ] Expand PR: this one.
- [ ] Migrate
- [ ] Contract

Still to come:
- enabling the features on a per SDK process
- supporting Rust websockets during the expand phase

### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async

### DIFF
--- a/.changeset/livekit-net-transport-seam.md
+++ b/.changeset/livekit-net-transport-seam.md
@@ -1,0 +1,19 @@
+---
+livekit-api: major
+livekit: major
+livekit-uniffi: minor
+---
+
+Route LiveKit signalling through a pluggable, host-providable transport (new `livekit-net` crate).
+
+The signalling WebSocket and the two pre-connect HTTP GETs (validate, region discovery) now go through a transport trait resolved from a process-global registry, so mobile/UniFFI builds can drop the Rust TLS/WebSocket/HTTP stack and use the platform's network stack. Native desktop builds are unchanged in behavior.
+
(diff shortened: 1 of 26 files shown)


.changeset/livekit-net-transport-seam.md
```diff
@@ -0,0 +1,19 @@
+---
+livekit-api: major
+livekit: major
```
1egoman, line 3: question: Should this be a major version bump? For a user of the rust crates directly (ie, not via `livekit-ffi` / `livekit-uniffi`), what about this change is breaking?

UPDATE: talked to lukas about this, I forgot that knope has some unintuitive behavior where `major` when under version `1.0.0` actually means `minor`. So, I think this is ok.

1egoman (commented):
I think I could use some guidance on how to best review this change - what areas are most important to focus on? Surfacing one thing that suck out to me now though.

.changeset/livekit-net-transport-seam.md
ladvoc: TIL

1egoman:
@jhugman Can this be closed now that #1258 has been opened?

jhugman:
closed in favour of #1258

PR #1234, opened 2026-07-09
Add e2e signalling tests

lukasIO:
### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async

### DIFF
--- a/livekit-api/src/signal_client/signal_test.rs
+++ b/livekit-api/src/signal_client/signal_test.rs
@@ -1,0 +1,528 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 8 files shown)


xianshijing-lk (approved):
lgtm, great tests. 

Curious,  any plans to add tests that verify media (audio / video tracks) as well? For example, making sure tracks are correctly restored after a successful reconnect, or verifying what happens when media negotiation fails?

livekit-api/src/signal_client/signal_test.rs
```diff
@@ -0,0 +1,528 @@
...
+}
+
+// -- happy path -------------------------------------------------------------
+
+/// `happy` — the WS sends a `JoinResponse` populated with the room from the
+/// token and non-zero ping config (so the client arms keepalive).
+#[tokio::test]
+async fn happy_join() {
+    let base = base_url();
+    skip_if_offline!(base);
```
1egoman: nitpick: It looks like `skip_if_offline!` is used in all tests. Is there ever a reason `base_url` would be called without `skip_if_offline!`, and if not, maybe drop the macro to reduce duplication and indirection?

```rust
fn online_lk_test_server() -> Option<String> {
    let base_url = std::env::var("LK_TEST_SERVER_URL").unwrap_or_else(|_| "http://127.0.0.1:9999".to_owned());
    if !reachable(&base_url).await {
        eprintln!("skipping: mock test server not reachable at {base_url}");
        None
    } else {
        Some(base_url)
    }
}

// And in each test:
let Some(base_url) = online_lk_test_server() else {
    return;
};
```

1egoman (approved):

livekit-api/src/signal_client/signal_test.rs
lukasIO: good idea

PR #1239, opened 2026-07-10
harden changeset dependency detection

lukasIO:
### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async

### DIFF
--- a/.github/scripts/changeset_detect.py
+++ b/.github/scripts/changeset_detect.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python3
+# Copyright 2026 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Detect which knope-managed packages a PR affects and validate changeset coverage.
+
+A changed crate requires a version bump not just for itself, but for every crate
+that (transitively) depends on it. This computes that closure from `cargo metadata`
+and compares it against the bumps already declared in the PR's changeset files.
+
+Inputs (environment variables):
+  CHANGED_FILES     newline-separated list of files changed by the PR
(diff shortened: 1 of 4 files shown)


.github/scripts/changeset_detect.py
```diff
@@ -0,0 +1,217 @@
...
+import subprocess
+import sys
+
+BUMP_ORDER = {"patch": 0, "minor": 1, "major": 2}
+
+
+def emit(data):
+    """Write the result as JSON and exit successfully."""
+    json.dump(data, sys.stdout)
+    sys.exit(0)
```
xianshijing-lk, line 53: curiously, Is exiting with status 0 even on failures intentional?

.github/scripts/changeset_detect.py
```diff
@@ -0,0 +1,217 @@
...
+    return visited
+
+
+def match_changed_packages(changed_files, pkg_to_dir):
+    """Map changed files to knope packages (longest directory prefix wins)."""
+    sorted_pkgs = sorted(pkg_to_dir.items(), key=lambda x: len(x[1]), reverse=True)
+    direct = set()
+    for f in changed_files:
+        for pkg_name, pkg_dir in sorted_pkgs:
+            if f.startswith(pkg_dir + "/"):
```
xianshijing-lk, line 125: will the script run on windows?

.github/scripts/changeset_detect.py
xianshijing-lk, line 53: I wonder if the sys.exit() should be called on the upper level rather than emit() function

xianshijing-lk (approved):
some questions

.github/scripts/changeset_detect.py
lukasIO: unlikely, currently the intention for this script is to be run in CI only, but happy to add windows support if you think that's useful

.github/scripts/changeset_detect.py
lukasIO: yeah, right now the script is treated as being run successfully as the consumer (CI job) gets additional failure metadata via emit. 

I can restructure this to exit with a different code on failures if you feel strongly about it?

lukasIO: `emit` is the last step in the script, basically it's `emit_and_exit`, I can rename to make the intention clearer?

lukasIO:
@xianshijing-lk I'll merge this for now, if you have any follow ups on the comments above, I'll address them in a separate PR

.github/scripts/changeset_detect.py
ladvoc: I don't see a case where we would need to run this for Windows in CI.

PR #1241, opened 2026-07-10
Renovate per crate updates

1egoman:
The renovate pull requests are [getting out of control](https://github.com/livekit/rust-sdks/pull/1208). This change splits up the pull requests to be per-crate rather than to be one big one for the whole project. Each individual crate will tag the respective code owners of that crate for review.

This will result in more, but better scoped pull requests which can hopefully be more easily merged outright rather than one massive almost always broken pull request which needs human / LLM fixing before merging.


ladvoc (approved):
LGTM ✅

PR #1243, opened 2026-07-10
Generate Go package for UniFFI

ladvoc:
Experimental Go package support for _livekit-uniffi_. Outstanding tasks:
- [ ] Investigate package name-spacing (currently has same issues as other languages)
- [ ] Setup distribution path


ladvoc:
No longer considering sharing Rust code with Go in the near term, will close for now.

PR #1246, opened 2026-07-13
Add livekit-capture crate

chenosaurus:
A new crate providing codec-neutral video capture for publishing to LiveKit: a VideoCaptureSource facade over decoded-frame camera capture (macOS AVFoundation, Linux V4L2, Jetson libargus) and pre-encoded ingest (TCP byte streams, RTSP-over-TCP, GStreamer appsink) with shared H264/H265/VP8/VP9/AV1 ingress helpers. Encoded sources honor WebRTC rate-control targets and keyframe requests, validate AV1/H265 access units on ingest, and report capture-origin timing.

### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async


livekit-capture/Cargo.toml
```diff
@@ -0,0 +1,79 @@
...
+version = "0.1.0"
+readme = "README.md"
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[dependencies]
+base64 = { workspace = true, optional = true }
+bytes = { workspace = true }
+gstreamer = { workspace = true, optional = true }
```
ladvoc: suggestion: `gstreamer` and `gstreamer-app` don't have to be workspace dependencies since they will only ever be consumed by this crate.

livekit-capture/build.rs
```diff
@@ -0,0 +1,65 @@
...
+            mmapi_root.display()
+        );
+        return;
+    }
+
+    println!("cargo:rerun-if-changed=src/sources/lk_argus.cpp");
+
+    cc::Build::new()
+        .cpp(true)
+        .file("src/sources/lk_argus.cpp")
```
ladvoc: issue: It is unconventional to build C/C++ lib sources inline in a higher-level Rust crate; see [_*-sys_ Packages](https://doc.rust-lang.org/stable/cargo/reference/build-scripts.html?utm_source=chatgpt.com#-sys-packages):

> It is common to have a companion package without the _-sys_ suffix that provides a safe, high-level abstractions on top of the sys package. For example, the _git2_ crate provides a high-level interface to the _libgit2-sys_ crate.

We don't necessarily need an intermediate _libargus_ to provide high-level abstractions on top of the sys package, but we should at least isolate the C++ dependency in a _libargus-sys_ crate.

livekit-capture/build.rs
```diff
@@ -0,0 +1,65 @@
...
+        .include(&mmapi_include)
+        .flag("-std=c++14")
+        .flag("-Wno-deprecated-declarations")
+        .compile("lk_argus");
+
+    println!("cargo:rustc-cfg=livekit_capture_argus");
+    println!("cargo:rustc-link-lib=dylib=nvargus_socketclient");
+    println!("cargo:rustc-link-lib=dylib=nvbufsurface");
+
+    let tegra_lib_dir = PathBuf::from("/usr/lib/aarch64-linux-gnu/tegra");
```
ladvoc: question: Should this also be overridable via env in the same way as `JETSON_MULTIMEDIA_API_DIR`?

livekit-capture/build.rs
```diff
@@ -0,0 +1,65 @@
...
+    if tegra_lib_dir.exists() {
+        println!("cargo:rustc-link-search=native={}", tegra_lib_dir.display());
+    }
```
ladvoc: suggestion: Warn if this directory doesn't exist:
```suggestion
    if tegra_lib_dir.exists() {
        println!("cargo:rustc-link-search=native={}", tegra_lib_dir.display());
    } else {
        println!(
            "cargo:warning=Tegra library directory {} not found",
            tegra_lib_dir.display()
        );
    }
```

ladvoc:
Closing in favor of #1299.

PR #1248, opened 2026-07-13
Strip DWARF from iOS FFI release archives

### LINKED ISSUE #1087: iOS staticlib (liblivekit_ffi.a) ships with unstripped DWARF — 462 MB vs ~45 MB
### Summary

The published iOS slices of `liblivekit_ffi.a` (both `ffi-ios-arm64` and `ffi-ios-sim-arm64`) carry unstripped DWARF debug info. The same code on macOS/Android ships at the expected size:

| Platform        | File                       | Size   |
|-----------------|----------------------------|--------|
| iOS arm64       | `liblivekit_ffi.a`         | 462 MB |
| iOS sim arm64   | `liblivekit_ffi.a`         | 463 MB |
| macOS arm64     | `liblivekit_ffi.dylib`     | 17 MB  |
| Android arm64   | `liblivekit_ffi.so`        | 18 MB  |

Reproduced against `livekit-ffi 0.12.52` (bundled in `io.livekit.livekit-sdk` v1.3.6 for Unity).

### Measured composition of the iOS arm64 archive

Extracted all 3,294 object files and summed Mach-O section sizes with `otool -l`:

| Section            | Size     | % of file |
|--------------------|----------|-----------|
| `__TEXT` (code)    | 17.3 MB  | 3.9%      |
| `__DWARF` (debug)  | 393.0 MB | 88.9%     |
| symbol tables, etc | 31.5 MB  | 7.1%      |

`__TEXT` matches the macOS/Android slices, confirming the code itself is the same size on every platform — the iOS difference is entirely embedded debug info.

After `strip -S` + `ranlib`, the archive shrinks **462 MB → 45 MB** with all 102,672 exported symbols preserved (including `livekit_ffi_request`, `livekit_ffi_drop_handle`, `livekit_ffi_initialize`). Linking and functional behavior verified unchanged.

### Why it happens

iOS uses `cargo rustc --crate-type staticlib`, which produces an `ar` archive of object files — no link step happens at this stage, so no symbol stripping is applied. macOS/Android ship dynamic libraries (`.dylib`/`.so`) whose link step naturally drops DWARF.

### Why most users don't notice

Apple's linker strips DWARF when the .a is linked into the host app, so the final IPA is normal size. UaaL/XCFramework consumers read & re-link the archive multiple times per build (device + simulator slices, then again when the host app embeds the resulting framework) and pay the input-side cost on every cycle. Build times more than doubled on our project after adding LiveKit.

### Suggested fix

Two equivalent options:

1. **Post-build strip** in the iOS release CI step:
   ```sh
   strip -S target/aarch64-apple-ios/release/liblivekit_ffi.a
   ranlib  target/aarch64-apple-ios/release/liblivekit_ffi.a
   # same for aarch64-apple-ios-sim

2. Cargo profile in livekit-ffi/Cargo.toml:
[profile.release]
strip = "debuginfo"
2. Worth testing whether this actually strips the C++ WebRTC objects too — Rust's strip profile setting historically only stripped Rust's own emitted code. If it doesn't fully cover WebRTC, fall back to option 1.

Environment

- io.livekit.livekit-sdk v1.3.6 / livekit-ffi 0.12.52
- Unity 6, Xcode 16+, macOS arm64 (Apple Silicon)
- Unity-as-a-Library workflow producing an XCFramework consumed by a Swift host app

drewcrawford:
## Summary

- Strip DWARF debug information from the release iOS device and simulator
  `liblivekit_ffi.a` archives before packaging.
- Rebuild each archive's symbol index with `ranlib`.
- Leave manually requested debug artifacts unstripped.
- Add a patch changeset for `livekit-ffi`.

Fixes #1087.

## Background

The iOS FFI builds produce static archives, so there is no final dynamic-link step
to discard DWARF information. Although the workspace release profile enables
symbol stripping, that does not remove debug information embedded in every
archive member, including the bundled C/C++ WebRTC objects.

Post-processing the completed archive with Apple's `strip -S` handles the whole
artifact while preserving global symbols required by downstream linkers.

## Measured impact

| Target | Raw archive | Compressed archive | Packaging time |
| --- | --- | --- | --- |
| `aarch64-apple-ios` | 462.23 → 44.92 MiB | 122.26 → 14.37 MiB | 13.59 → 1.54s |
| `aarch64-apple-ios-sim` | 463.84 → 45.07 MiB | 122.58 → 14.44 MiB | 13.76 → 1.51s |

Across both published iOS archives, this reduces raw size by 90.28% and
compressed size by 88.23%.

DWARF section size falls to zero. All defined external symbols and all 3,435
archive members are preserved, including `livekit_ffi_request`,
`livekit_ffi_initialize`, and `livekit_ffi_drop_handle`.

## Design tradeoffs

Using a Cargo profile setting alone was avoided because it does not reliably
strip debug information from all bundled C/C++ archive members.

Post-build stripping adds roughly 61 seconds to each iOS release matrix job on
my laptop. Faster packaging saves about 12 seconds per artifact, and the
smaller upload and downstream processing costs should offset additional time
outside the build itself.

## Testing

- Built release `liblivekit_ffi.a` for:
  - `aarch64-apple-ios`
  - `aarch64-apple-ios-sim`
- Applied `xcrun strip -S` and `xcrun ranlib` to both archives.
- Verified zero remaining DWARF bytes.
- Compared archive members and defined external symbols before and after.
- Verified required LiveKit FFI entry points remain exported.
- Verified Apple's linker can consume both post-strip archives.
- Built the macOS arm64 dynamic-library control and confirmed it
  already contains zero DWARF.
- Ran `cargo fmt --all -- --check`, workflow YAML parsing, changeset validation,
  and `git diff --check`.

## CI note

This PR also adds one line to `gen-node-proto.yml`: the checkout step now
sets `repository:` to the PR head repository. The job previously fetched
the head branch name from the base repository, which fails for any
fork-based PR matching the workflow's path filter — including this one,
via `.github/**`. Same-repo PRs are unaffected
(`head.repo.full_name` equals `github.repository` there).

### DIFF
--- a/.changeset/strip-ios-ffi-dwarf.md
+++ b/.changeset/strip-ios-ffi-dwarf.md
@@ -1,0 +1,5 @@
+---
+livekit-ffi: patch
+---
+
+Strip DWARF debug information from release iOS FFI archives.
--- a/.github/workflows/ffi-builds.yml
+++ b/.github/workflows/ffi-builds.yml
@@ -203,6 +203,15 @@ jobs:
           cd livekit-ffi
           cargo rustc --crate-type staticlib --profile "$CARGO_RUST_PROFILE" --target ${{ matrix.target }} ${{ matrix.buildargs }}
 
+      - name: Strip iOS release archive
+        if: ${{ matrix.platform == 'ios' }}
+        run: |
+          if [ "$CARGO_RUST_PROFILE" = "release" ]; then
+            archive="target/${{ matrix.target }}/$CARGO_PROFILE_DIR/${{ matrix.dylib }}"
+            xcrun strip -S "$archive"
+            xcrun ranlib "$archive"
+          fi
+
       - name: Build (Windows)
         if: ${{ matrix.platform == 'windows' }}
         run: |
--- a/.github/workflows/gen-node-proto.yml
+++ b/.github/workflows/gen-node-proto.yml
@@ -35,6 +35,7 @@ jobs:
         with:
           submodules: true
           ref: ${{ github.event.pull_request.head.ref || github.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
 
       - name: Install Protoc
         uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0


drewcrawford:
@ladvoc mind taking a look when you get a chance? Still green and merges cleanly. Takes the iOS FFI archive from 462 MB to ~45 MB (#1087). Happy to rebase, or split out the `gen-node-proto.yml` fix if that’s easier to review.

ladvoc:
Thank you for your contribution! @pblazej, when you have a chance, would you mind reviewing this?

pblazej (approved):
I think there's no better option for static lib and Unity is a hard blocker here (can't use dylib at all)?

cc @MaxHeimbrock

MaxHeimbrock (approved):
Looks good, I tested it locally and can confirm the great improvements in binary size.

PR #1250, opened 2026-07-14
Ensure participant disconnects are synthesized after connection resume

lukasIO:
### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async


livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -1363,13 +1399,23 @@ impl SessionInner {
                 );
             }
             proto::signal_response::Message::Update(mut update) => {
+                #[cfg(feature = "__lk-e2e-test")]
```
ladvoc, line 1426: question: Why is this gated behind the E2E test feature?

lukasIO: this is used/injected in e2e tests only as faulty behaviour during a signal disconnect to ensure we can mimic losing/missing participant disconnect events after a resume.

livekit/src/rtc_engine/mod.rs
```diff
@@ -114,6 +114,12 @@ pub enum EngineEvent {
     ParticipantUpdate {
         updates: Vec<proto::ParticipantInfo>,
     },
+    /// A signal resume fully recovered; any known participant whose identity
+    /// is not in `seen_identities` left the room while the signal link was
+    /// down and its disconnection must be synthesized.
+    ParticipantReconcile {
+        seen_identities: HashSet<String>,
```
1egoman: nitpick: It might be better to make this `HashSet<ParticipantIdentity>`:


```suggestion
        seen_identities: HashSet<ParticipantIdentity>,
```

Then that would avoid some of the unwrapping code and the `String` -> `&str` conversions in places, and also add the type safety nicety of ensuring you aren't passing in a random string which is a non participant identity in.

livekit/src/rtc_engine/rtc_session.rs
1egoman: I agree this is a bit surprising on first read, maybe a comment calling out this isn't a mistake on the line above might be a good idea.

1egoman (approved):
I think I generally get it!

livekit/src/rtc_engine/mod.rs
lukasIO: good idea!

PR #1252, opened 2026-07-14
Fix publisher transport self-deadlock during renegotiation

drewcrawford:
Summary

- Release the publisher transport's inner mutex before starting a deferred renegotiation.
- Add a loopback regression test for a second negotiation requested while an offer is awaiting its answer.
- Verify that applying the answer completes and emits the deferred follow-up offer.

## Problem

`PeerTransport::set_remote_description` held `self.inner`, a `tokio::sync::Mutex`, while processing an answer. If another publisher negotiation had been requested while the previous offer was in flight, `create_and_send_offer` recorded that request by setting `inner.renegotiate = true`.

After applying the answer, `set_remote_description` saw that flag and called `create_and_send_offer` without releasing the mutex guard. `create_and_send_offer` immediately tried to lock `self.inner` again. Because Tokio's mutex is not reentrant, the task waited forever for the lock it already held.

This leaves the room reporting `Connected` while publisher operations that need the transport mutex remain blocked. Further track negotiation and ICE restart cannot proceed without tearing down the connection.

## Production reachability

The deadlock is reachable through the ordinary `HaveLocalOffer` path when
`fast_publish` is disabled and publisher changes use the 150 ms debouncer:

1. A publish or unpublish operation creates an offer and the publisher enters `HaveLocalOffer`.
2. After that debounced negotiation completes, but before the server's answer is applied, another
   publish or unpublish change schedules a second debounced negotiation.
3. `create_and_send_offer` sees the offer in flight and sets `renegotiate = true`.
4. The answer arrives and `set_remote_description` attempts the deferred renegotiation.

This can occur when track changes are separated by the debounce window and signaling latency keeps the first offer in flight long enough for the second negotiation to run. When `fast_publish` is
enabled, subsequent publisher negotiations are queued outside `PeerTransport`, generally avoiding
this exact overlap.

## Fix

Clear the deferred-renegotiation flag and drop the `inner` guard before calling `create_and_send_offer`. The nested call can then acquire the mutex normally.

## Testing

The new `renegotiation_does_not_deadlock` test uses two in-process peer connections:

1. Alice creates an ordinary publisher offer and enters `HaveLocalOffer`.
2. Bob prepares an answer, but Alice does not apply it yet.
3. A second offer request sets the deferred-renegotiation flag.
4. Alice applies Bob's answer under a timeout.
5. The test verifies that the call completes and that a second offer is emitted.

- `cargo fmt --all -- --check`
- `cargo test -p livekit --lib`

### DIFF
--- a/.changeset/fix-publisher-renegotiation-deadlock.md
+++ b/.changeset/fix-publisher-renegotiation-deadlock.md
@@ -1,0 +1,6 @@
+---
+livekit: patch
+livekit-ffi: patch
+---
+
+Fix a publisher-transport deadlock during renegotiation. When another negotiation was requested while an offer was awaiting its answer, `set_remote_description` re-entered `create_and_send_offer` while holding the transport's non-reentrant inner mutex, permanently wedging publishing.
--- a/livekit/Cargo.toml
+++ b/livekit/Cargo.toml
@@ -59,6 +59,9 @@ base64 = "0.22"
 [dev-dependencies]
 # Enable data-stream test constructors (e.g. TextStreamReader::new_for_test) for our test suites.
 livekit-data-stream = { workspace = true, features = ["test-utils"] }
+# Async test runtime + timers, needed by the peer_transport deadlock reproducer.
+# (rt/time are already built elsewhere in the workspace, so no new crates enter the lockfile.)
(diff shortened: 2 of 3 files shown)


xianshijing-lk (approved):
great find. 

lgtm

PR #1256, opened 2026-07-15
Fix uniffi release: publish directly via knope, react on the release event

pblazej:
Every `livekit-uniffi` release has stranded (`v0.1.2`–`v0.1.5` are unpublished drafts with no tags): the `assets = "marker"` draft was never published by CI, and the one-shot draft lookup also raced Knope's creation. Since the Dart/Flutter cdylibs — the only asset the GH release ever hosted — are experimental and deferred, the release no longer needs to stage assets, so this drops the marker and lets Knope publish the `livekit-uniffi/v*` release + tag directly (like every normal crate). `uniffi-packages.yml` now reacts to `on: release: [published]` (filtered to `livekit-uniffi/v*`) and publishes the Swift xcframework + Android AAR — no draft, no lookup/retry, no manual finalize. The release event fires only for genuine livekit-uniffi releases, so the "released this cycle" gate is free (same publish-directly spirit as the Node/npm flow; since Maven — unlike npm — can't dedupe versions, this event-gating replaces npm's implicit idempotency). `ffi-builds.yml` and its marker are left unchanged.

**Note:** `on: release` firing for a Knope-App-published release is expected (GitHub Apps trigger workflow events; knope-bot's merges already trigger `push`/`pull_request` here) but hasn't been exercised in this repo yet — `workflow_dispatch` is kept as a manual fallback, and it's worth confirming on the first release. The already-stranded drafts (`v0.1.2`–`v0.1.5`) are left untouched (non-destructive); publish/clean them up manually as desired.


pblazej:
@ladvoc feel free to close it, there must be some more elegant solution to both this and normal ffi

ladvoc (approved):
Generally makes sense to me, but I will defer to @lukasIO for the Knope config.

lukasIO (approved):

PR #1258, opened 2026-07-15
Refactor networking from signalling into a new livekit-net crate

jhugman:
Supersedes #1229.
Fixes CLT-2177

This PR is the first of three:

- [x] refactor. This moves networking from signalling into a new livekit-net crate.
- [ ] expand. This adds support for the livekit-net transport to be injected by the host platform
- [ ] migrate
- [ ] contract. This removes support for the default native transport provided by reqwest and tungstenite.

`livekit-net` introduces a `PlatformTransport` trait, which client crates then use to call out to the network. Start in the `livekit-net/src/transport.rs` file.

This covers the websocket interface and http, so feasibly we might have also call this network client.

The large line difference:

- about +762-804 is just for `Cargo.lock`
- almost all of the new livekit-net crate is just cut/paste/adapt from livekit-api: +946 -0 for the livekit-net, then +214 −962 for the removal commit.

### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async

### DIFF
--- a/livekit-api/Cargo.toml
+++ b/livekit-api/Cargo.toml
@@ -11,39 +11,25 @@ readme = "README.md"
 # By default ws TLS is not enabled
 default = ["services-tokio", "access-token", "webhooks"]
 
-signal-client-tokio = [
-    "dep:tokio-tungstenite",
+# Signalling client, blind to the transport backend. Pulls livekit-net (no
+# backend), the tokio crate (signal_client uses tokio::sync directly), and
+# protobuf helpers. The livekit-runtime flavor comes from the signal-client-<rt>
+# wrappers below or the top crate — not here, so async/dispatcher don't collide
+# with tokio under the one-runtime guard. Needs a runtime flavor to compile, and
+# a registered livekit_net::transport() at runtime.
+signal-client = [
+    "dep:livekit-net",
+    "dep:livekit-runtime",
     "dep:tokio",
-    "dep:futures-util",
-    "dep:reqwest",
-    "dep:livekit-runtime",
-    "livekit-runtime/tokio",
     "dep:base64",
     "dep:flate2",
-    "dep:bytes"
+    "dep:bytes",
+    "dep:serde_json",
 ]
 
-signal-client-async = [
-    "__signal-client-async-compatible",
-    "livekit-runtime/async"
-]
-
-signal-client-dispatcher = [
-    "__signal-client-async-compatible",
-    "livekit-runtime/dispatcher"
-]
-
-__signal-client-async-compatible = [
-    "dep:async-tungstenite",
-    "dep:tokio", # For macros and sync
-    "dep:futures-util",
-    "dep:isahc",
-    "dep:livekit-runtime",
-    "dep:base64",
-    "dep:flate2",
-    "dep:bytes"
-]
-
+signal-client-tokio      = ["signal-client", "livekit-net/native-tokio", "livekit-runtime/tokio"]
+signal-client-async      = ["signal-client", "livekit-net/native-async",      "livekit-runtime/async"]
+signal-client-dispatcher = ["signal-client", "livekit-net/native-dispatcher",  "livekit-runtime/dispatcher"]
 
 services-tokio = ["dep:reqwest", "dep:tokio", "tokio/time", "dep:livekit-runtime", "livekit-runtime/tokio"]
 services-async = ["dep:isahc", "dep:livekit-runtime", "livekit-runtime/async"]
@@ -53,8 +39,7 @@ webhooks = ["access-token", "dep:serde_json", "dep:base64"]
 # TLS Configuration
 # -----------------
 # These features control TLS behavior for WebSocket and HTTP connections.
-# Note: These features only change the behavior of tokio-tungstenite and reqwest.
-# They don't change the behavior of libwebrtc/webrtc-sys.
+# TLS is delegated to livekit-net; no local tungstenite/rustls deps.
 #
 # IMPORTANT FOR CONTAINER DEPLOYMENTS:
 # When using `rustls-tls-native-roots`, the SDK relies on the operating system's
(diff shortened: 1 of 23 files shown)


livekit-net/src/transport.rs
```diff
@@ -0,0 +1,57 @@
...
+/// A record wrapper, not a bare `Arc<dyn PlatformConnection>`: uniffi 0.31 cannot
+/// lift a trait object returned from an async `with_foreign` method.
+pub struct PlatformConnectResult {
+    pub connection: Arc<dyn PlatformConnection>,
+}
+
+/// A host- or Rust-provided network transport. Opens WebSockets and performs the
+/// pre-WS HTTP GETs. Knows nothing about LiveKit/protobuf.
+#[async_trait::async_trait]
+pub trait PlatformTransport: Send + Sync {
```
1egoman: thought: Is `-Transport` sufficiently clear here? I would think it is somewhat ambiguous whether this means "webrtc transport" or "websocket transport". It might make sense to consider a name that either has "websockets" or "signaling" in it to disambiguate.

livekit-net/src/transport.rs
```diff
@@ -0,0 +1,57 @@
...
+        headers: Vec<Header>,
+        timeout_ms: u64,
+    ) -> Result<PlatformConnectResult, TransportError>;
+
+    /// The pre-WS HTTP GETs (validate, region discovery).
+    async fn http_get(
+        &self,
+        url: String,
+        headers: Vec<Header>,
+    ) -> Result<HttpResponse, TransportError>;
```
1egoman: thought: This might be part of a future phase, but I wonder if it makes sense to break out this method into a second trait here which is HTTP specific.

This trait could for now just have `http_get` but in the future could be expanded to add all the verbs. This would be useful as part of the [rust token source pull request](https://github.com/livekit/rust-sdks/pull/749) which @MaxHeimbrock is going to be taking over and getting across the line shortly. I think he'll likely need a `_post` variant for `TokenSourceEndpoint` eventually.

Then it would be up to the caller to implement both or either depending on what downstream crates they use.

livekit-net/tests/native_parity.rs
```diff
@@ -0,0 +1,113 @@
...
+        );
+        sock.write_all(resp.as_bytes()).await.unwrap();
+    });
+    port
+}
+
+#[tokio::test]
+async fn ws_send_recv_roundtrip() {
+    // Minimal echo WS server using tokio-tungstenite (dev-dep).
+    let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
```
1egoman, line 37: Very cool, I've never come across `:0` before, [good to know](https://networkengineering.stackexchange.com/questions/11234/tcp-port-0-reserved-for-what-purpose)!

livekit-net/src/native/mod.rs
```diff
@@ -0,0 +1,156 @@
+// Copyright 2025 LiveKit, Inc.
```
1egoman: nitpick: Incorrect license year in this file's header, as well as the majority of this change.

```suggestion
// Copyright 2026 LiveKit, Inc.
```

livekit-api/src/signal_client/mod.rs
```diff
@@ -914,6 +930,57 @@ fn get_livekit_url(
...
+                SignalError::Client(code, String::new())
+            } else {
+                SignalError::Server(code, String::new())
+            }
+        }
+        TE::Closed => SignalError::Closed,
+        TE::Connection(m) => SignalError::Connection(m),
+        TE::Other(m) => SignalError::Connection(m),
+    }
+}
```
1egoman: thought: Would be better implemented as a `impl From<livekit_net::TransportError> for SignalError`?

In other crates (I know `livekit-datatrack` does this), there's a lot of usage of [`from_variant`](https://crates.io/crates/from_variant) so that errors can nicely be cascaded up a stack with just `.into()`.

livekit-api/Cargo.toml
```diff
@@ -11,39 +11,25 @@ readme = "README.md"
...
-]
-
-signal-client-async = [
-    "__signal-client-async-compatible",
-    "livekit-runtime/async"
-]
-
-signal-client-dispatcher = [
-    "__signal-client-async-compatible",
-    "livekit-runtime/dispatcher"
```
1egoman, line 33: FYI @ladvoc, you I think had the best understanding of how all the network / async features worked previously and this changes them fairly significantly, so it would be good to get your eyes here specifically.

livekit-api/src/http_client.rs
```diff
@@ -123,23 +72,13 @@ mod async_std {
...
         }
 
-        #[cfg(feature = "services-async")]
         pub fn post(&self, url: url::Url) -> RequestBuilder {
             RequestBuilder {
                 body: Vec::new(),
                 builder: isahc::http::Request::post(url.as_str()),
                 client: self.0.clone(),
             }
         }
```
1egoman, line 81: suggestion: This would be a great place to consume that http specific trait I mentioned in an earlier comment, which would let you drop even more of this code I think.

In practice I'm not sure that the "non native" paths would be useful here so I think the primary benefit would be around dropping `#[cfg(feature = "services-tokio")]` / `#[cfg(feature = "services-async")]` around the big `mod tokio` / `mod async_std` blocks earlier in this file.

livekit-net/src/transport.rs
jhugman: Agreed, I've split this up into a `WsClient` and `HttpClient`, and renamed `PlatformConnection` to `WsConnection`

livekit-net/src/transport.rs
jhugman: Agreed. I added a `HttpClient` trait, which when implemented by the foreign side is just `request`, but is decorated by get and post methods.

livekit-api/src/http_client.rs
jhugman: I left this for a followup PR, given that these services seem to be all server side.

1egoman (approved):

livekit-net/src/native/proxy.rs
```diff
@@ -0,0 +1,235 @@
...
+    let proxy_env = if url.scheme() == "wss" {
+        env::var("HTTPS_PROXY").or_else(|_| env::var("https_proxy"))
+    } else {
+        env::var("HTTP_PROXY").or_else(|_| env::var("http_proxy"))
+    };
+
+    // Connect directly or through proxy
+    let ws_stream = if let Ok(proxy_url) = proxy_env {
+        if !proxy_url.is_empty() {
+            log::info!("Using proxy: {}", proxy_url);
```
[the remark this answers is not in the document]
1egoman: ^ This is a good point, but for another reason - websocket connection urls can contain tokens in the query string (admittedly which are short lived). In other clients I believe there is logic to redact these in logs. I'm not sure if the rust sdk was doing this previously but if it was that behavior should definitely continue.

livekit-net/src/lib.rs
```diff
@@ -25,6 +25,18 @@ pub use types::{Header, HttpResponse, TransportError};
...
+/// port, and path.
+pub fn redact_url(url: &url::Url) -> String {
+    let mut u = url.clone();
+    let _ = u.set_username("");
+    let _ = u.set_password(None);
+    u.set_query(None);
+    u.set_fragment(None);
+    u.to_string()
+}
+
```
1egoman, line 39: @jhugman Should these be set to like `xxx` or something? I worry that omitting them might make people reading the relevant logs think that these aren't being set.

PR #1261, opened 2026-07-16
Handle room closure for ReadyForRoomEventRequest

ladvoc:
Timeout was originally put in place to prevent connect from hanging indefinitely if FFI client never sends `ReadyForRoomEventRequest`. However, it has been observed slow Python applications can still hit this timeout. Rather than increasing the timeout higher, the correct solution to prevent waiting indefinitely is to make the wait operation cancellable when the room is closed.

Follow-up to #1188
Closes BOT-473

### DIFF
--- a/livekit-ffi/src/server/room.rs
+++ b/livekit-ffi/src/server/room.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use std::collections::HashMap;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use std::{collections::HashSet, slice, sync::Arc};
 
 use livekit::{prelude::*, registered_audio_filter_plugins, PluginError};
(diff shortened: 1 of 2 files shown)


alan-george-lk (approved):
LGTM!

livekit-ffi/src/server/room.rs
```diff
@@ -242,28 +244,34 @@ impl FfiRoom {
...
-                        drop(handle);
-                        ffi_room.close(server, DisconnectReason::ConnectionTimeout).await;
-                        server.drop_handle(handle_id);
-                        server.send_panic(Box::new(FfiError::InvalidRequest(msg.into())));
-                        return;
+                    // Forward room events only once the client signals it is
+                    // ready to receive them; cancelled if the room is closed
+                    // first (returning releases the handle lock so close() can
+                    // proceed).
+                    let wait_start = Instant::now();
```
stephen-derosa, line 251: discussed offline, lets add a log or something if the buffer size is incremented and we havent received a ReadyForRoomEventRequest. Reason being we want to mask a bug from any of the client SDKs if there is one

PR #1267, opened 2026-07-20
livekit-api: Strip query params from info-level connect log

gasmith:
The signal websocket URL's query string can include a rather large encoded `join_request` blob. I'm not sure whether it contains secrets, but it has that appearance, and it's a lot of clutter for an info-level log statement.

This change updates the connect method to log the URL without query parameters at info level. The full URL is still logged at debug level.

@jhugman: Looks like this might conflict with #1258.

### DIFF
--- a/.changeset/trim-info-log.md
+++ b/.changeset/trim-info-log.md
@@ -1,0 +1,5 @@
+---
+livekit-api: patch
+---
+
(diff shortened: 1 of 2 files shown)


gasmith:
Looks like this is no longer required after #1258, thanks to `livekit_net::redact_url`.

PR #1273, opened 2026-07-22
allow for audio filters to be registered after initial room connection

lukasIO:
## Problem

Audio-filter plugins are authenticated in a one-time pass: when a room
connects, `FfiRoom::connect` calls `on_load(url, token)` for every plugin
registered *at that moment*. A plugin registered **after** the room has
connected is never initialized for it — its later `create` then fails
because the connection was never authenticated (surfacing downstream as
"failed to initialize the audio filter").

Because `room.connect()` is async, any registration that lands in that
window loses the race.

## Fix

When a plugin is registered (`LoadAudioFilterPlugin`), also run its `on_load`
against every already-connected room, using each room's connect URL and
current token — so registration order no longer matters.

- New `Room::token()` accessor (livekit): current signalling token, needed to
  authenticate late-registered plugins.
- New `FfiServer::list_rooms()` (livekit-ffi): snapshot of connected rooms
  (refactored out of the existing `dispose` logic).
- `on_load_audio_filter_plugin` runs `on_load` per room on `spawn_blocking`
  (fire-and-forget), so a plugin whose `on_load` does network I/O can't stall
  the request thread. `on_load` is idempotent per URL, so this can't conflict
  with the connect-time pass. Registration happens before the loop so a
  concurrently-connecting room is still covered by the normal path.

## Notes

- Uses the room's connect URL (not the resolved signalling URL) so the auth
  key matches what `create` later looks up.

### DIFF
--- a/livekit-ffi/src/server/requests.rs
+++ b/livekit-ffi/src/server/requests.rs
@@ -17,7 +17,7 @@ use std::{slice, sync::Arc};
 use colorcvt::cvtimpl;
 use livekit::{
     prelude::*,
-    register_audio_filter_plugin, registered_audio_filter_plugin,
+    register_audio_filter_plugin,
     webrtc::{native::apm, native::audio_resampler, prelude::*},
     AudioFilterPlugin, SimulateScenario,
 };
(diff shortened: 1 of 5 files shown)


livekit-ffi/src/server/requests.rs
```diff
@@ -1050,7 +1050,26 @@ fn on_load_audio_filter_plugin(
         }
     };
 
-    register_audio_filter_plugin(request.module_id, plugin);
+    register_audio_filter_plugin(request.module_id, plugin.clone());
```
xianshijing-lk, line 1053: can you add an integration test for this use case ?

xianshijing-lk (commented):
lgtm if you add the test.

livekit-ffi/src/server/requests.rs
```diff
@@ -1039,7 +1039,7 @@ fn on_perform_rpc(
 }
 
 fn on_load_audio_filter_plugin(
```
xianshijing-lk, line 1041: maybe it is a dumb question, do we have code that protect calling on_load_audio_filter_plugin() more than once per session ?

livekit-ffi/src/server/requests.rs
```diff
@@ -1050,7 +1050,26 @@ fn on_load_audio_filter_plugin(
...
+    // authenticated). Run `on_load` now for any already-connected rooms so
+    // registration order doesn't matter.
+    for room in server.list_rooms() {
+        let plugin = plugin.clone();
+        let url = room.inner.url();
+        let token = room.inner.room.token();
+        // Spawn (don't block the request): a plugin's `on_load` may perform
+        // network I/O, and this must not stall the FFI request thread.
+        server.async_runtime.spawn_blocking(move || {
+            if let Err(e) = plugin.on_load(&url, &token) {
```
xianshijing-lk, line 1070: any chance that plugin.on_load() could be called more than once, will it work ?

ladvoc: That would be dependent on plug-in implementation to ensure it is idempotent; we can't enforce here.

livekit-ffi/src/server/requests.rs
lukasIO: I don't think it's needed, do you have a specific concern?

livekit-ffi/src/server/requests.rs
lukasIO: I added a e2e test that loads a dummy audio filter.

xianshijing-lk (approved):

PR #1275, opened 2026-07-22
build(livekit-uniffi): keep uniffi `cli` out of shipped library builds

jhugman:
The uniffi `cli` feature (clap + the bindings backends) is only needed by the `uniffi-bindgen` bin, but it was a hard dependency feature — so it was compiled into every library build of the shipped cdylib/staticlib, and was also pulled in transitively through livekit-datatrack's `uniffi` feature.

Gate it behind an opt-in `cli` feature with `required-features` on the bin (mirroring the existing `dart-bindgen` pattern), drop the spurious `cli` from livekit-datatrack (a plain library that never calls the bindgen CLI), and pass `--features cli` in the Makefile bindgen task. Bindgen already runs in library mode against the compiled cdylib, so it stays fully decoupled.

Removes clap + the bindgen backends (~43 MiB) from the static archive and cuts crate compile time ~3x. The LTO'd shipped dylib already dead-stripped the unreachable cli code, so the dynamic artifact is byte-unchanged.

### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async

### DIFF
--- a/.changeset/remove_uniffi_cli_from_staticlib.md
+++ b/.changeset/remove_uniffi_cli_from_staticlib.md
@@ -1,0 +1,5 @@
+---
+livekit-uniffi: patch
+---
+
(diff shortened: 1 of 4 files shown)


1egoman (approved):

pblazej (approved):
👏

PR #1286, opened 2026-07-27
Data streams v2 uniffi

1egoman:
Adds data streams v2 to the `livekit-uniffi` crate, and exposes it fairly similarly to how data tracks works.

I've also for now added a python test script (It's the most complete bindgen set up in this project which I am familiar with) which exercises this api as an example of what it would look like in practice. I'll remove this before an eventual merge, but I thought it would be useful for reviewers:

```sh
$ python3 datastream_uniffi_test.py
--- OUTGOING:
PACKETS: [b'jP\n$4a501804-960c-4d54-94df-81784ffab6b1\x10\xc6\xa5\xb0\xaf\xf93\x1a\x04test"\ntext/plain(\x0bJ\x00Z\x0bhello world']
--- INCOMING:
TEXT STREAM OPENED: alice CONTENTS: hello world
```


pblazej:
My suggested approach here is to just open a draft PR to _consumers_ e.g. swift and then polish it - that's what I did for data tracks before.

Otherwise the reviewer must generate the bindings anyway, as it's very hard to predict the edge cases around concurrency, memory management, etc.

I'd be super happy to take a look at both in parallel.

livekit-uniffi/src/data_stream/incoming.rs
```diff
@@ -0,0 +1,230 @@
...
+pub struct ByteStreamReader {
+    info: ByteStreamInfo,
+    inner: Mutex<ds_api::ByteStreamReader>,
+}
+
+impl ByteStreamReader {
+    fn new(reader: ds_api::ByteStreamReader) -> Self {
+        Self { info: reader.info().clone().into(), inner: Mutex::new(reader) }
+    }
+}
```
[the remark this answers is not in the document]
1egoman: This is fine, `ds_api::ByteStreamReader` contains an `Arc<RwLock<HashMap<String, String>>>` for the attributes. And calling `.info().attributes()` returns a copy on demand of the data in that hashmap.

livekit-uniffi/uniffi.toml
```diff
@@ -6,6 +6,48 @@ android = true
 package_name = "io.livekit.uniffi"
```
MaxHeimbrock, line 6: Should this be com.livekit.uniffi in the future?

livekit-uniffi/src/data_stream/incoming.rs
```diff
@@ -0,0 +1,216 @@
...
+                    }
+                    ds_api::AnyStreamReader::Text(reader) => {
+                        let reader = Arc::new(TextStreamReader::new(reader));
+                        self.delegate.on_text_stream_opened(reader, identity);
+                    }
+                }
+            }
+            // Deprecated v1 raw chunk/trailer notifications are not surfaced over the FFI boundary.
+            ds::incoming::OutputEvent::ChunkReceived(_)
+            | ds::incoming::OutputEvent::TrailerReceived(_) => {}
```
pblazej, line 258: Agreed on `ChunkReceived`, but `TrailerReceived` is doing double duty: it's the only "stream closed on the wire" signal the core emits, and hosts need it to order handler invocations.

The Swift SDK supports `ordered` text-stream topics (transcription). v1 chained a newly opened stream behind the handlers of streams that had already **closed**. With only `StreamOpened`, the best a host can do is chain on the order streams **opened** in — so a stream that stays open head-of-line-blocks every later stream from that sender on the topic. Restoring the v1 test against this gives:

```
Expectation failed: (received → [])          == ["b"]        // B blocked behind still-open A
Expectation failed: (received → ["a", "b"])  == ["b", "a"]   // then drains in open order
```

The event already carries `trailer.stream_id`, `participant_identity` and `topic`, so forwarding it — or something narrower — would close it:

```rust
fn on_stream_closed(&self, stream_id: String, identity: String);
```

Not Swift-specific: any host doing ordered delivery hits this, so Kotlin and Dart would each reinvent a workaround. The only alternative is parsing trailers before `handle_packet_received`, which defeats the boundary.

livekit-uniffi/src/data_stream/outgoing.rs
```diff
@@ -0,0 +1,221 @@
...
+    manager: ds::outgoing::Manager,
+    registry: Arc<dyn lk_common::RemoteParticipantRegistry>,
+    _guard: DropGuard,
+}
+
+/// Delegate for receiving outbound packets from [`OutgoingDataStreamManager`].
+#[uniffi::export(with_foreign)]
+pub trait OutgoingDataStreamManagerDelegate: Send + Sync {
+    /// Encoded [`livekit_protocol::DataPacket`]s to be sent over the data channel transport.
+    fn on_packets_available(&self, packets: Vec<Bytes>);
```
pblazej: **The `()` return.** You flag this at line 93 already; noting the cost so it can be weighed rather than discovered. `write()` can't throw when a packet doesn't reach the wire (the Swift API documents that it does), `is_open()` can't go false on a send failure, and there's no back-pressure — the responder acks once the host has *buffered*, not sent, so a producer looping on `write()` queues unboundedly. Host-side bounding isn't an option, since dropping stream packets is worse than the memory. A `Result`-returning delegate covers all three.

**The `Vec`.** It's always exactly one packet — the pump does `on_packets_available(vec![packet])` per `recv()`. The plural signature implies batching that never happens, so every host writes a loop that runs once. Either make it singular or actually batch; batching would also make in-batch ordering free and cut the per-packet foreign round trip.

livekit-uniffi/src/data_stream/common.rs
```diff
@@ -0,0 +1,370 @@
...
+
+// MARK: - Wire decode
+
+/// Decodes a serialized [`proto::DataPacket`] carrying a data-stream header/chunk/trailer into an
+/// incoming-manager input event. Returns `None` if the bytes don't decode or the packet isn't a
+/// data-stream packet.
+///
+/// Encryption is defaulted to `None`: end-to-end encryption for data streams over this FFI is a
+/// follow-up (the foreign side is expected to hand us already-decrypted packets).
+pub(crate) fn decode_data_packet(bytes: &[u8]) -> Option<ds::incoming::PacketReceived> {
```
pblazej: Worth recording that this *can't* be plumbed here at all, so nobody later tries to fix it by decoding harder.

`encrypted_packet` is field 18 **inside** `oneof value`, alongside `stream_header`/`stream_chunk`/`stream_trailer`. Hosts decrypt by writing the decrypted member into that same oneof, destroying `encrypted_packet`. By the time these bytes arrive the type isn't defaulted — it's absent from the wire format.

So `descriptor.encryption_type != encryption_type` in `handle_chunk` can never be true over the FFI: both sides are always `None`. That guard is dead code here while still reading as active. Not a regression — the field was equally unreadable in the Swift v1 path — I raise it only because the code suggests otherwise. E2EE-over-FFI will need:

```rust
pub fn handle_packet_received(&self, packet: Bytes, encryption_type: EncryptionType)
```

**Line 310:** `EncryptionTypeMismatch` is a unit variant, so even once it can fire a host can't say which types disagreed — Swift's error is `encryptionTypeMismatch(expected:received:)` and currently fabricates `.none`/`.none`. Carrying both values would fix that.

livekit-uniffi/src/data_stream/incoming.rs
```diff
@@ -0,0 +1,216 @@
...
+#[uniffi::export]
+impl IncomingDataStreamManager {
+    #[uniffi::constructor]
+    pub fn new(
+        delegate: Arc<dyn IncomingDataStreamManagerDelegate>,
+        max_payload_byte_length: Option<u64>,
+    ) -> Arc<Self> {
+        let token = CancellationToken::new();
+        let (manager, input, output) =
+            ds::incoming::Manager::new(max_payload_byte_length.map(|n| n as usize));
```
pblazej, line 81: The cap is fixed at construction with no setter. On the Swift side `maxPayloadSize` comes from room options, which aren't final until `connect` and can differ between sessions on the same `Room` — so honoring a new value means discarding and rebuilding the manager rather than reconfiguring it.

That works cleanly (the `DropGuard` cancels all three tasks, handler state is host-side) and is what I shipped. But it's easy to get wrong in the other direction: the obvious implementation — construct lazily, memoize for the object's lifetime — silently pins the cap to the first session's value. That was in fact the bug I hit.

```rust
pub fn set_max_payload_byte_length(&self, max: Option<u64>)
```

Lower priority than #1–#3; the workaround is sound once you know to write it.

livekit-uniffi/src/data_stream/incoming.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub fn handle_packet_received(&self, packet: Bytes) {
+        if let Some(event) = decode_data_packet(&packet) {
+            let _ = self.input.send(event.into());
+        }
+    }
+
+    /// Aborts all open incoming streams so their readers error instead of hanging (e.g. on
+    /// disconnect). Handler wiring on the foreign side survives, so streams that arrive later
+    /// (e.g. after a reconnect) are still processed.
+    pub fn abort_all_streams(&self) {
```
pblazej, line 112: Minor, testability only. There's no way to observe how many streams are open, so no way to wait for one to *be* open before exercising the abort paths.

The v1 Swift manager exposed `openStreamCount`, and its tests used it to wait for an orphaned header to register before calling `closeStreams(from:)`/`reset()` — exactly the race these methods resolve. Without it I infer "open" by signalling from inside the handler on entry, which measures handler-dispatched rather than descriptor-registered, and would stop being equivalent if those ever move relative to each other.

An `open_stream_count() -> u64` would be enough. Only worth it if cheap.

pblazej (commented):
I integrated this on the Swift side (livekit/client-sdk-swift#1075) and it works end-to-end. Compression and single-packet inlining both kick in correctly once capabilities round-trip, which I verified on the wire.

Below are five gaps in the FFI surface I hit while wiring it up, ordered by what they cost a host rather than by size. None of them blocks the integration; #1 and #2 are the two that cost correctness rather than convenience, and both look cheaper to close now than after this ships and hosts have worked around them.

livekit-uniffi/src/data_stream/outgoing.rs
1egoman: Re batching - let me see what I can do to send multiple packets at once here. For context, my aim was to match the data track [PacketsAvailable](https://github.com/livekit/rust-sdks/blob/f9c47c5a3141a7b9f55218e09c30187ed5e13693/livekit-datatrack/src/local/events.rs#L50) interface here which takes a `Vec<Bytes>`.

livekit-uniffi/src/data_stream/incoming.rs
1egoman: IMO this is better fixed with a docstring which calls it out. I think it's a very similar class of problem as a single room instance which can be `connect()`ed multiple times, so while I think it is a rough edge I don't think it's worth introducing the complexity to dynamically change options later unless really necesary.

livekit-uniffi/src/data_stream/incoming.rs
1egoman: Good idea, done in 6ea6605f3

livekit-uniffi/src/data_stream/incoming.rs
1egoman: Added docstring in d7a2b09e2

livekit-uniffi/src/data_stream/common.rs
1egoman: What I had done here before was a bit sloppy, yes I agree. I have adjusted the call sites which take a `Packet` to also take an `EncryptionType` alongside in da55afa45, and I think that makes it somewhat better.

livekit-uniffi/src/data_stream/incoming.rs
1egoman: Good point, done in 9dbb89b3c

livekit-uniffi/src/data_stream/outgoing.rs
1egoman: For the `Result` so that you can pass back up send failures - I've added a `Result<(), SendError>` return to `on_packets_available` in 1dd54d078. I opted to reuse the existing `SendError` error type, although that has the downside of dropping any sort of "reason" type string. Maybe I could add a log or add that reason string? I'm open to others perspectives here.

For batching, as mentioned above, I've kept `Vec<Bytes>`, but I've changed this so that the vec sending path is used in `send_text` / `send_bytes`. With time this could be expanded but I think this is a good initial use case which also keeps it aligned with PacketsAvailable in the data track subsystem.

pblazej: Prototyped rather than just suggested — pushed as 7a2505fb.

A sync callback can't deliver what the doc promises. "Return once handed to the transport" needs a synchronous transport; Swift's is async throughout, so honoring it means blocking a Rust runtime thread via a primitive that SDK forbids. The only implementable choice was to ack on buffering — so back-pressure and transport errors went unimplemented and `is_open` could never go false.

Four lines: `#[async_trait::async_trait]`, `async fn`, `.await` at the call site. uniffi keeps the trait dyn-compatible itself; `async-trait` was already in the lockfile. Hosts get simpler too — ordering becomes structural, so Swift deleted its pump.

Verified: 104 + 12 Rust tests; 144 Swift tests clean under TSan and `LIBDISPATCH_COOPERATIVE_POOL_STRICT=1`; and a case untestable before — after disconnect `write()` throws and `isOpen` is false, where both previously reported success on an unwritable stream.

Unverified: Kotlin/Dart impact, `async-trait` now a direct dep.

livekit-uniffi/src/data_stream/incoming.rs
pblazej: Leaving this one open — the docs help, but the gap is the immutable field, not the documentation. Swift still discards and rebuilds the manager per session to pick up a new `maxPayloadSize`, which works but is a rebuild forced by a missing setter, and the obvious host implementation (construct lazily, memoize for the object's lifetime) still silently pins the cap to the first session.

Fine to close as wontfix if you'd rather keep the surface small; just didn't want to mark it done when the workaround is still load-bearing.

pblazej:
Pushed 7a2505fb, which makes `on_packets_available` an `async fn` — a synchronous callback can't honor the contract documented above it on hosts whose transport is async — with the reasoning, verification and open Kotlin/Dart question in the thread below; revert freely if you'd rather it went another way.

1egoman:
FYI @pblazej - 1da01b4 was an issue I found when updating the android implementation to take into account the changes from your review. I think it may have some implications on the swift version as it changes the format of the `Packet::Trailer` message to include `encryption_type`. See the commit message for more about the exact context about what was fixed.

pblazej:
> FYI @pblazej - [1da01b4](https://github.com/livekit/rust-sdks/commit/1da01b49c4dc26a0dbb4690e4bbcb1d2935df653) was an issue I found 

👏 I don't see anything else here _at least for Swift_, ready for final review?

livekit-uniffi/uniffi.toml
pblazej: Out of curiosity, why? 😃

1egoman:
(rebased on top of latest `main`)

pblazej (approved):
LGTM

Small cleanup items are probably:
- minor devin comments (docstrings etc.)
- repinning dart?
- removing python tester?

1egoman:
@pblazej Just removed the python testing script in 72bdb58.

IMO, I think let's wait until the end of the week and see if https://github.com/Uniffi-Dart/uniffi-dart/pull/154 is merged. If it takes longer then that, I will plan to merge this pull request as is against my personal fork and make a follow up PR which cuts back over to the official one we can merge once my upstream PR is merged.

Thanks for all your help testing and reviewing this!

1egoman:
(rebased on top of latest `main`)

pblazej:
@1egoman completing flutter would be even better! 💯 cc @hiroshihorie

livekit-uniffi/uniffi.toml
MaxHeimbrock: just because our branding changed from livekit.io to livekit.com. Don't know if this is related here though.

pblazej: ah yes there's a redirect now

Issue #497, opened 2024-11-28
Capture frame failed


vuleetu:
I found a lot of`RtcError` errors in worker logs.

```
Nov 28 01:57:50 python[117393]:     await self._audio_source.capture_frame(frame)
Nov 28 01:57:50 python[117393]:   File "/home/ubuntu/miniconda3/envs/chat/lib/python3.12/site-packages/livekit/rtc/audio_source.py", line 152, in capture_frame
Nov 28 01:57:50 python[117393]:     raise Exception(cb.capture_audio_frame.error)
Nov 28 01:57:50 python[117393]: Exception: an RtcError occured: InvalidState - failed to capture frame {"pid": 121680, "job_id": "AJ_oHY4vMyQc8kr"}
```

Transcriptions were streamed to client normally, but TTS were lost. I need help to figure out how to avoid this issue

btakita:
I'm also getting this error when streaming audio after the room opens. Is there any chance that the error message can provide more detail in what state is invalid?

boyko11:
After I dug down to the Rust code:
https://github.com/livekit/rust-sdks/blob/0773bcec4e24812b382e71166b2ab359bf1a9384/libwebrtc/src/native/audio_source.rs#L87

and with some LLM help, I was able to alleviate the issue in our python code by doubling the size of the queue_size_ms:
```
audio_source = rtc.AudioSource(
            sample_rate=self.global_audio_config.SAMPLING_RATE_DOWNSTREAM,
            num_channels=self.global_audio_config.CHANNELS,
            queue_size_ms=2000
```

I was able to reproduce it fairly consistently by setting the queue_size_ms to 10

vishnukool:
I am facing the exact same issue. TTS is lost but text transmission continues. Why is this marked as Not planned?

ladvoc:
I think this was closed automatically due to inactivity, but I’ll reopen it since it appears that this issue persists.

ladvoc:
@vishnukool, would you mind sharing what version of the agents framework you are using?

vishnukool:
OK returning the favor to the world. This got fixed by updating livekit agents to latest - livekit-agents 1.2.15

Issue #420, opened 2024-09-04
Capture Frame Consistently Hang Audio Output


jkears:
I am using a modified Unity SDK for .NET and I am able to connect, create rooms add participants, process remote participant audio as well as publish audio track/source played back on the client (Blazor using JS SDK).   This works but is highly temperamental to load.

The following diagram depicts a testing approach that I am working in which a .NET console application is creating 100 rooms (amount can vary).   Not illustrated is a a Blazor client app that connects to it's own room and interacts with AI. 

![image](https://github.com/user-attachments/assets/bce83368-9355-4f1a-914e-935d08422d26)

On the backend, the service uses a LiveKit Web Hook to detect the new rooms being created by the Test Client and creates a separate Room Processor, each of which has it's own Ffi-LiveKit instance,  each of which monitors remote participants and audio tracks and captures the audio via VAD to AI and back.  This works well, and in a test of 100 rooms being created plus one that I use via the Blazor client, allows me to run the test whilst speaking via the Blazor app.  

The issue I have had (and from the start of this journey), is that the capture_frame (RUST SDK) function is a blocking function but it is extremely  temperamental to any load that is occurring within the host web service.  If there is only a single Blazor client i.e. single room, this works well, however as I run the test client which creates 100 rooms, it doesn't take long before any ongoing audio stream that is pushed to a track in the same room as the Blazor client hangs and hangs for good.  I can never resend through the broken channel.   

This is improved (but with some frame loss) when instead of calling **Send** in RUST SDK I change to **TrySend** and at least I get notified of the frame loss errors, but it still will lose the channel _even if the load ceases_.

When this locks up, the only way I know to recover is to disconnect the the Blazor client and reconnect.   Also, when this occurs, there are significant errors showing in the JS SDK.

This is the function that will hang forever when using Send but that I modified with TrySend.  

```
pub async fn capture_frame(&self, frame: &AudioFrame<'_>) -> Result<(), RtcError> {
        if self.sample_rate != frame.sample_rate || self.num_channels != frame.num_channels {
            return Err(RtcError {
                error_type: RtcErrorType::InvalidState,
                message: "sample_rate and num_channels don't match".to_owned(),
            });
        }

        //print!("capture_frame locking");

        // Set a maximum lock hold of 1 second
        let lock_result = timeout(Duration::from_secs(1), self.inner.lock()).await;

        // Handle timeout or lock acquisition failure
        let mut inner = match lock_result {
[...]

Issue #408, opened 2024-08-25
capture_frame never completes


jkears:
I have repurposed the Unity SDK for LiveKit into an new .Net Client SDK.  This invokes requests via the LiveKit-Ffi and it works amazingly well ... except for one major issue.  

We have our own internal framework which include our own AI Assistant solution.   When a chat completion is transformed to audio and is applied to an audio source as a capture frame it works for a while but more so than not, it seems that the libwebrtc native audio_source function blocks forever and I am not sure why.  

The reason I know this is occurring in or within this function is that a calling method does not raise the CaptureAudioFrame event is never raised ...

```
let res = source.capture_frame(&audio_frame).await;
let _ = server.send_event(proto::ffi_event::Message::CaptureAudioFrame(
                        proto::CaptureAudioFrameCallback {
                            async_id,
                            error: res.err().map(|e| e.to_string()),
                        },
                    ));
```

In this method ... 
```
 pub async fn capture_frame(&self, frame: &AudioFrame<'_>) -> Result<(), RtcError> {
        if self.sample_rate != frame.sample_rate || self.num_channels != frame.num_channels {
            return Err(RtcError {
                error_type: RtcErrorType::InvalidState,
                message: "sample_rate and num_channels don't match".to_owned(),
            });
        }
        let mut inner = self.inner.lock().await;
        let mut samples = 0;
        // split frames into 10ms chunks
        loop {
            let remaining_samples = frame.data.len() - samples;
            if remaining_samples == 0 {
                break;
            }
            if (inner.len != 0 && remaining_samples > 0) || remaining_samples < self.samples_10ms {
                let missing_len = self.samples_10ms - inner.len;
                let to_add = missing_len.min(remaining_samples);
                let start = inner.len;
                inner.buf[start..start + to_add]
                    .copy_from_slice(&frame.data[samples..samples + to_add]);
                inner.len += to_add;
                samples += to_add;
                if inner.len == self.samples_10ms {
                    let data = inner.buf.clone().to_vec();
                    let _ = self.po_tx.send(data).await;
                    inner.len = 0;
                }
                continue;
            }
            if remaining_samples >= self.samples_10ms {
                // TODO(theomonnom): avoid copying
                let data = frame.data[samples..samples + self.samples_10ms].to_vec();
                let _ = self.po_tx.send(data).await;
                samples += self.samples_10ms;
            }
        }
        Ok(())
    }
}
```
[...]

PR #1289, opened 2026-07-28
Fix buffered AudioSource.capture_frame hanging when the source drain stalls

sam-hark:
### Before you submit your PR

### PR description

**Summary**

- Bound the buffered `NativeAudioSource::capture_frame` completion wait (scaled to the queue duration) so a stalled source drain returns a recoverable `RtcError` instead of hanging forever.
- Split the `AudioTrackSource::InternalSource` lock so the drain's `sink->OnData()` no longer runs under the same mutex as the producer — a wedged/slow sink can no longer block `capture_frame`, while a sink still cannot be freed mid-`OnData`.
- On a stall/`clear_buffer`, release the pending completion so the source stays usable; reclaim the completion context on the `false` return path (previously leaked).

**Problem**

A buffered native audio source (`queue_size_ms > 0`, the default used by `livekit-agents`' `_ParticipantAudioOutput`) can hang `capture_frame` forever, silently killing outbound audio while the room still reports `Connected`. Reported repeatedly and worked around locally, never fixed upstream:

- **#408** "capture_frame never completes" — reporter pinpointed it: *"the buffer never becomes un-full and thus we block forever"*.
- **#420** "Capture Frame Consistently Hang Audio Output" — *"highly temperamental to load … the multi-threaded, task-switching nature … GC playing a role"*.
- **#497** "Capture frame failed" — Python agents: *"TTS is lost … on long stories voice just stops"*.

**Root cause**

The buffered source is a producer/consumer pair across the FFI:

- **Consumer** — `webrtc-sys/src/audio_track.cpp`, `AudioTrackSource::InternalSource`: a 10 ms `RepeatingTask` drains the buffer, pushes 10 ms to each sink via `sink->OnData(...)`, and — only when it runs — fires the deferred completion for the chunk that filled the buffer past the notify threshold.
- **Producer** — `libwebrtc/src/native/audio_source.rs`, `NativeAudioSource::capture_frame`: for each chunk it hands C++ a `oneshot::Sender` and `rx.await`s that completion **with no timeout**.

The drain held `mutex_` across the entire `OnData` loop. So when the drain is slow or CPU-starved, it is almost always holding the lock, and the producer's synchronous FFI `capture_frame` blocks acquiring that same mutex — often before it even reaches the awaited completion. Either way, `capture_frame` never returns: `_wait_for_playout` never completes, generation tasks pile up, and the session goes permanently mute while ICE stays connected.

**Fix**

- **`audio_track.cpp` / `audio_track.h` — split the lock.** `sink_mutex_` guards `sinks_` and is held across the `OnData` loop; `buffer_mutex_` guards the buffer and completion state. The producer (`capture_frame`) only ever takes `buffer_mutex_`, so a wedged/slow sink can no longer block it — while `RemoveSink` still serializes against an in-flight `OnData`, so a sink cannot be freed mid-call. The drain copies the 10 ms frame into a reused scratch buffer (no per-tick allocation).
- **`audio_source.rs` — bound the wait.** `rx.await` → `tokio::time::timeout(...)`, scaled to the queue duration (`5 × queue_ms`, floored at 1 s) so legitimate backpressure never trips it. On timeout, `clear_buffer()` and return `RtcError { InvalidState }`.
- **Keep the source usable + no leak.** `clear_buffer()` now also releases any pending completion (freeing its context and clearing the slot), so a fresh `capture_frame` succeeds after a stall clears rather than erroring forever. The `false`-return path now reclaims the boxed `Sender` it previously leaked.

The timeout and the lock split are coupled: with the drain holding the mutex across `OnData`, the producer blocks in the FFI before the await, so bounding the await alone would be dead code in exactly the scenario it targets.

### Breaking changes

None. `capture_frame` gains a recoverable `RtcError` for the stall case, replacing an unobservable permanent hang; callers already handle `capture_frame` errors.

### MSRV

Unchanged.

### Testing

`libwebrtc/src/native/drain_stall_tests.rs`, exercised by the standard `cargo test`:

- **`capture_frame_recovers_when_drain_stalls`** — attaches a sink whose `OnData` blocks (a deterministic stalled consumer) to a real `AudioTrackSource`, feeds `capture_frame`, asserts it returns the drain-stall `RtcError` (bounded so a wiring regression fails cleanly rather than hanging CI), then — after the stall clears — asserts a fresh `capture_frame` succeeds (the source is not poisoned). Gated to Linux: it needs a real audio track, and on macOS/Windows that brings up the platform AudioDeviceModule, whose real-time thread aborts the process when the blocking sink stalls it. The fix is platform-independent.
- **`capture_frame_completes_under_normal_backpressure`** — runs on every platform (no track/sink); with the drain running, deferred captures complete `Ok`, confirming the timeout does not false-trip on legitimate backpressure. Uses a large queue so the assertion is not coupled to the production floor.

Verified on a real webrtc build: stock hangs; timeout-without-lock-split still hangs (producer wedged on the mutex); the combined fix returns an error in ~1 s and the source recovers; the full `libwebrtc` lib suite passes.

### Async

The bounded wait is the fix itself; the test also uses `tokio::time::timeout` to detect a hang, mirroring the existing `renegotiation_does_not_deadlock` test, and subscribes to a channel signalled the instant the drain enters the sink rather than sleeping for it.

### Out of scope (deferred)

`capture_frame` remains a synchronous FFI call that can block its caller's task while the drain catches up; a fuller change might make the enqueue non-blocking (or run it off the async runtime). This PR keeps that shape and only ensures the wait is bounded and the lock no longer couples the producer to a wedged sink.

### DIFF
--- a/libwebrtc/src/native/audio_source.rs
+++ b/libwebrtc/src/native/audio_source.rs
@@ -204,7 +204,9 @@ impl NativeAudioSource {
 
             // Bound the wait for the drain's completion (timeout rationale above). On a stall,
             // clear_buffer() releases the pending completion so the source stays usable afterward.
-            match tokio::time::timeout(capture_timeout, rx).await {
+            // Route through livekit_runtime so capture_frame stays runtime-neutral (tokio::time
+            // would panic when awaited off a Tokio runtime).
+            match livekit_runtime::timeout(capture_timeout, rx).await {
                 Ok(_) => {}
                 Err(_) => {
                     self.clear_buffer();
--- a/libwebrtc/src/native/audio_source.rs
+++ b/libwebrtc/src/native/audio_source.rs
@@ -18,6 +18,14 @@ use webrtc_sys::audio_track as sys_at;
 
 use crate::{audio_frame::AudioFrame, audio_source::AudioSourceOptions, RtcError, RtcErrorType};
 
+/// Floor for the per-chunk completion wait, for very small queues.
+const CAPTURE_COMPLETE_TIMEOUT_FLOOR: std::time::Duration = std::time::Duration::from_secs(1);
+
+/// Headroom multiple applied to the queue duration when bounding the completion wait: a healthy
+/// drain acknowledges a chunk within one queue-duration, so this leaves margin before the wait is
+/// treated as a stall.
+const CAPTURE_COMPLETE_TIMEOUT_QUEUE_MULTIPLE: u64 = 5;
+
 #[derive(Clone)]
 pub struct NativeAudioSource {
     sys_handle: SharedPtr<sys_at::ffi::AudioTrackSource>,
@@ -153,6 +161,20 @@ impl NativeAudioSource {
             let _ = tx.send(());
         }
 
+        // Bound the per-chunk completion wait on a multiple of the queue duration. The C++
+        // drain acknowledges a chunk from a 10ms RepeatingTask; legitimate backpressure clears
+        // within one queue-duration. A wait beyond that means the drain stalled (CPU-starved,
+        // or a sink blocked in OnData), so surface a recoverable error rather than hang the
+        // caller — and the whole session — forever (see rust-sdks #408 / #420 / #497).
+        let queue_ms = (self.queue_size_samples as u64)
+            .saturating_mul(1000)
+            .checked_div((self.sample_rate as u64) * (self.num_channels as u64))
+            .unwrap_or(0);
+        let capture_timeout = std::time::Duration::from_millis(
+            queue_ms.saturating_mul(CAPTURE_COMPLETE_TIMEOUT_QUEUE_MULTIPLE),
+        )
+        .max(CAPTURE_COMPLETE_TIMEOUT_FLOOR);
+
         // iterate over chunks of self._queue_size_samples
         for chunk in frame.data.chunks(self.queue_size_samples as usize) {
             let nb_frames = chunk.len() / self.num_channels as usize;
@@ -161,7 +183,9 @@ impl NativeAudioSource {
             let ctx_ptr = Box::into_raw(ctx) as *const sys_at::SourceContext;
 
             unsafe {
-                // In the fast path, C++ never store / invoke on_complete / ctx.
+                // C++ only takes ownership of `ctx` when capture_frame returns true; on a false
+                // return (buffer full, or a prior completion still pending) it has not, so reclaim
+                // the Box here to avoid leaking the Sender.
                 if !self.sys_handle.capture_frame(
                     chunk,
                     self.sample_rate,
@@ -170,6 +194,7 @@ impl NativeAudioSource {
                     ctx_ptr,
                     sys_at::CompleteCallback(lk_audio_source_complete),
                 ) {
+                    drop(Box::from_raw(ctx_ptr as *mut oneshot::Sender<()>));
                     return Err(RtcError {
                         error_type: RtcErrorType::InvalidState,
                         message: "failed to capture frame".to_owned(),
@@ -177,7 +202,18 @@ impl NativeAudioSource {
                 }
             }
 
-            let _ = rx.await;
+            // Bound the wait for the drain's completion (timeout rationale above). On a stall,
+            // clear_buffer() releases the pending completion so the source stays usable afterward.
+            match tokio::time::timeout(capture_timeout, rx).await {
+                Ok(_) => {}
+                Err(_) => {
+                    self.clear_buffer();
+                    return Err(RtcError {
+                        error_type: RtcErrorType::InvalidState,
+                        message: "audio capture timed out: source drain stalled".to_owned(),
+                    });
+                }
+            }
         }
 
         Ok(())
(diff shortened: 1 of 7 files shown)


libwebrtc/src/native/audio_source.rs
```diff
@@ -170,14 +194,26 @@ impl NativeAudioSource {
...
                         error_type: RtcErrorType::InvalidState,
                         message: "failed to capture frame".to_owned(),
                     });
                 }
             }
 
-            let _ = rx.await;
+            // Bound the wait for the drain's completion (timeout rationale above). On a stall,
+            // clear_buffer() releases the pending completion so the source stays usable afterward.
+            match tokio::time::timeout(capture_timeout, rx).await {
```
xianshijing-lk: Curiously, in what setup or workload did you actually hit this timeout? My read is that normal Python/agents usage should await capture_frame() and therefore use it as backpressure, so this would only happen if the native webrtc drain stops making progress. Not sure how that could happen.

Was your repro tied to long TTS/audio bursts, high process load, CPU starvation,  shutdown, or other workload sensitive conditions?

webrtc-sys/src/audio_track.cpp
```diff
@@ -154,34 +154,49 @@ AudioTrackSource::InternalSource::InternalSource(
...
+        // sink->OnData() under sink_mutex_ and fire the completion with no lock held. Keeping
+        // OnData off buffer_mutex_ means a wedged or slow sink can never block capture_frame;
+        // keeping it under sink_mutex_ means a sink cannot be freed (via RemoveSink) mid-call.
+        // scratch_ is reused across ticks (this task is the only thread that touches it) to avoid
+        // a per-10ms allocation on the audio path.
+        void (*complete)(const SourceContext*) = nullptr;
+        const SourceContext* complete_ctx = nullptr;
+        {
+          webrtc::MutexLock lock(&buffer_mutex_);
+          scratch_.resize(samples10ms);
```
xianshijing-lk: should we move this   scratch_.resize(samples10ms); to the constructor ? I would think we don't need to call resize() for every tick.

webrtc-sys/src/audio_track.cpp
```diff
@@ -229,8 +243,16 @@ bool AudioTrackSource::InternalSource::capture_frame(
...
 
 void AudioTrackSource::InternalSource::clear_buffer() {
-  webrtc::MutexLock lock(&mutex_);
+  webrtc::MutexLock lock(&buffer_mutex_);
   buffer_.clear();
+  // Release any chunk still awaiting completion so the source is immediately reusable and its
+  // context is freed rather than leaked. Serialized with the drain via buffer_mutex_, so the
+  // completion fires at most once.
+  if (on_complete_) {
+    on_complete_(capture_userdata_);
```
xianshijing-lk: This is a known problem and not related to your change,  but probably we shouldn't call on_complete_() under the buffer_mutex_ lock. Can you change the code to 

void (*complete)(const SourceContext*) = nullptr;
    const SourceContext* complete_ctx = nullptr;
    {
      webrtc::MutexLock lock(&buffer_mutex_);
      buffer_.clear();

      complete = on_complete_;
      complete_ctx = capture_userdata_;
      on_complete_ = nullptr;
      capture_userdata_ = nullptr;
    }

    if (complete)
      complete(complete_ctx);
  }

webrtc-sys/src/audio_track.cpp
```diff
@@ -154,34 +154,49 @@ AudioTrackSource::InternalSource::InternalSource(
...
 
-          buffer_.erase(buffer_.begin(), buffer_.begin() + samples10ms);
-        } else {
-          // Always provide a 10ms frame to avoid playout underruns.
+        {
+          webrtc::MutexLock lock(&sink_mutex_);
           for (auto sink : sinks_)
-            sink->OnData(silence_buffer_.data(), kBitsPerSample, sample_rate_,
-                         num_channels_, samples10ms / num_channels_);
+            sink->OnData(scratch_.data(), kBitsPerSample, sample_rate_, num_channels_,
```
xianshijing-lk, line 195: Is sink->OnData() the main problem that could block forever ? 

WebRTC’s send-side OnData is intended to be fast, this OnData just copies the frame and posts encoding work to an encoder task queue. It shouldn't block for long

webrtc-sys/src/audio_track.cpp
sam-hark: Good point, I updated

webrtc-sys/src/audio_track.cpp
sam-hark: Yes, I updated

libwebrtc/src/native/audio_source.rs
sam-hark: My repro was tied to high process load mostly during long TTS playout, but it has also been very intermittent and difficult to pin down. To be honest, I can't claim this is definitely the root cause of audio stalls I've seen, but it's my leading cause. 

It appears that under load the drain task stops getting scheduled, so the queue stays full and the capture_frame await never returns.

webrtc-sys/src/audio_track.cpp
sam-hark: Likely not, I could remove this from the PR if preferred. This is just trying to enforce that the producer shouldn't be able to block the sink, and it enables adding a deterministic unit test so that the timeout fix can be directly tested.

PR #1291, opened 2026-07-28
Fix pre-encoded frame segfault on macOS

ladvoc:
Capturing pre-encoded frames shortly after publishing a track segfaults on macOS: the sender starts on the real encoder (_VideoToolbox_) because the switch to the passthrough encoder is asynchronous, and any access unit captured before the switch carries a _kNative_ _EncodedVideoFrameBuffer_ that the _ObjC_ encoder blind-casts and retains as a garbage pointer (Linux survives via the defended _ToI420()_ path). Fixed by having non-passthrough encoders drop pre-encoded frames instead of forwarding them until the switch completes.

This was discovered while developing the _gstreamer_ source for _livekit-capture_.

### DIFF
--- a/webrtc-sys/src/video_encoder_factory.cpp
+++ b/webrtc-sys/src/video_encoder_factory.cpp
@@ -17,16 +17,21 @@
 #include "livekit/video_encoder_factory.h"
 
 #include <algorithm>
+#include <atomic>
 #include <cctype>
 #include <cstdlib>
 #include <optional>
 #include <string_view>
 #include <utility>
 
+#include "api/video/video_frame.h"
+#include "modules/video_coding/include/video_error_codes.h"
+
 #include "api/environment/environment_factory.h"
 #include "api/video_codecs/sdp_video_format.h"
 #include "api/video_codecs/video_encoder.h"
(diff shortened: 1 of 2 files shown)


webrtc-sys/src/video_encoder_factory.cpp
```diff
@@ -605,6 +610,80 @@ VideoEncoderFactory::CodecSupport VideoEncoderFactory::QueryCodecSupport(
...
+      if (!logged.exchange(true)) {
+        RTC_LOG(LS_WARNING)
+            << "Dropping pre-encoded access unit sent to a non pass-through "
+               "encoder; waiting for the sender to switch onto the "
+               "pass-through backend";
+      }
+      return WEBRTC_VIDEO_CODEC_OK;
+    }
+    return encoder_->Encode(frame, frame_types);
+  }
```
alan-george-lk, line 661: It feels a bit heavy to re-implement an entire encoder to just redo one of its methods, and then only register that for pre-encoded. I am not familiar with this C++ code very much, but that's my initial read.

chenosaurus (approved):

PR #1293, opened 2026-07-29
fix flaky test_audio: skip a settle window after signal onset

1egoman:
## Problem

`test_audio` fails intermittently in CI (macOS and Windows) with e.g.:

```
Detected sine frequency not within range for channel 0: 38Hz
```

Every occurrence is the `24000Hz,2ch -> 24000Hz,1ch` case reading ~32–38Hz for a 60Hz tone. First seen 2026-07-09.

## Root cause (a real early-audio regression, not just a test artifact)

A CI bisect (6 mac reruns per condition, extracting `test_audio`'s own verdict) showed:

| Condition | test_audio |
|---|---|
| main baseline (server 1.13.4, single-PC) | 3 flake / 6 |
| server pinned 1.13.2 (single-PC) | 0 / 6 |
| dual-PC (server 1.13.4) | 0 / 6 |

Both conditions are required, so it's an interaction:

1. A **single-PC** subscriber bundles audio SRTP and the (bundle-only, port-0) SCTP datachannel onto one DTLS transport.
2. **livekit-server 1.13.3** bumped `pion/webrtc` v4.2.15→v4.2.16, whose only media-affecting change is [`ec1dd730` "Fix bundle-only data channel startup"](https://github.com/pion/webrtc/commit/ec1dd730) — it now runs `startSCTP` for a bundle-only/port-0 datachannel (previously skipped).
3. `startSCTP` performs the SCTP association handshake over the **shared** DTLS transport at `setRemoteDescription`, contending with the first audio packets → NetEq emits ~300–400ms of concealment at stream start.
4. The zero-crossing estimate over the first second then reads low and trips the ±20Hz assertion.

Dual-PC is immune (separate subscriber transport); server ≤1.13.2 is immune (never started SCTP for the bundle-only datachannel). The underlying transport contention is a server/pion concern and is being raised separately.

## This change (test-side mitigation)

Detect signal onset, then discard a **500ms settle window** before measuring 100 frames of steady-state audio. This keeps the estimate off the noisy startup region. (An earlier version of this PR only skipped *leading* silence, which was insufficient — it left the interleaved startup concealment inside the window and still flaked 1/6.)

### DIFF
--- a/livekit/tests/audio_test.rs
+++ b/livekit/tests/audio_test.rs
@@ -65,6 +65,7 @@ async fn test_audio_with(params: TestParams) -> Result<()> {
     const SINE_FREQ: f64 = 60.0;
     const SINE_AMPLITUDE: f64 = 1.0;
     const FRAMES_TO_ANALYZE: usize = 100;
+    const SIGNAL_ONSET_AMPLITUDE: i32 = (i16::MAX / 10) as i32;
 
     let sine_params = SineParameters {
         freq: SINE_FREQ,
@@ -101,6 +102,16 @@ async fn test_audio_with(params: TestParams) -> Result<()> {
                 assert_eq!(frame.num_channels, params.sub_channels);
                 assert_eq!(frame.sample_rate, params.sub_rate_hz);
                 assert_eq!(frame.samples_per_channel, frame.data.len() as u32 / frame.num_channels);
+
+                // The stream can deliver silence between subscribing and the arrival of the
+                // publisher's first decodable packet; those frames have no zero crossings and
+                // would skew the frequency estimate, so start analyzing at signal onset rather
+                // than at the first delivered frame.
+                if frames_analyzed == 0
+                    && !frame.data.iter().any(|&s| (s as i32).abs() >= SIGNAL_ONSET_AMPLITUDE)
+                {
+                    continue;
+                }
 
                 for channel_idx in 0..params.sub_channels as usize {
                     analyzers[channel_idx].analyze(frame.channel_iter(channel_idx));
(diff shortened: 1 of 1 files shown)


1egoman:
## Bisect update — this fix is insufficient; converting to draft

Ran a CI bisect (6 mac reruns per condition, extracting `test_audio`'s specific verdict per attempt):

| Condition | test_audio |
|---|---|
| main baseline (server 1.13.4, single-PC) | 2 flake / 5 (~40%) |
| server pinned 1.13.2 (single-PC) | **0 / 6** |
| dual-PC (server 1.13.4) | **0 / 4** |
| **this branch** (gated test, 1.13.4, single-PC) | **1 / 6** |

Two conclusions:

1. **The amplitude gate does not eliminate the flake** (1/6, same `24000Hz,2ch -> 24000Hz,1ch` case, same 38Hz reading). The residual failure means the low-frequency reading is **not** caused only by leading silence before stream onset — there is degraded/gappy audio *within* the 1s analysis window. The gate addresses the wrong part of the signal.

2. **Reverting either the server (→1.13.2) or the connection topology (→dual-PC) makes it disappear.** Both landed right before the flake first appeared (single-PC default 2026-07-01 #1206; server 1.13.3 2026-07-03). This is a **real audio-delivery regression from the single-PC + server-1.13.3 interaction**, not a test artifact.

Holding this as draft. The real fix is either on the SDK/server interaction side, or a more aggressive test change that skips *all* low-energy frames (which would mask a genuine regression — needs a deliberate call).

ladvoc (approved):

PR #1300, opened 2026-07-30
fix(uniffi): repair Android package build and release-tag checkout

pblazej:
- build-android-*: cargo-make's `extend` replaces the parent env map rather than merging it, so `env = { TARGET = ... }` silently dropped ANDROID_RELEASE_FLAG and RUSTFLAGS — `--profile release` produced debug .so files and android-copy-jniLibs found nothing in target/*/release. Derive the flag from CARGO_MAKE_PROFILE in the script instead.
- android-bindgen-kotlin: TARGET leaks across cargo-make tasks, so bindgen-kotlin's `build` dependency cross-compiled for Android with the host linker ("cannot find -llog"). Pin TARGET back to the host triple and keep symbols, which library-mode bindgen needs on Linux.
- Raise both size gates to measured values: ios-arm64 is 1088 KiB and arm64-v8a 1174 KiB, having grown past the limits set in #1171 when the data-track UniFFI surface landed in #1034.
- Check out inputs.tag_name in both reusable workflows; workflow_dispatch was building the dispatch ref (main) rather than the requested tag.

### Before you submit your PR

### PR description

### Breaking changes

### MSRV

### Testing

### Async

### DIFF
--- a/.github/actions/uniffi-deps/action.yml
+++ b/.github/actions/uniffi-deps/action.yml
@@ -5,8 +5,3 @@ runs:
     - name: Install cargo-make
       uses: taiki-e/install-action@682e7d9e49c5e653d371fc6adbda67653461378a # v2.82.4
       with: { tool: cargo-make@0.37 }
-    - name: Install tera-cli
-      uses: taiki-e/install-action@682e7d9e49c5e653d371fc6adbda67653461378a # v2.82.4
-      with: { tool: tera-cli@0.5.0 }
-    # ^ This would be installed by the cargo-make file automatically, but this ensures
-    # it is already available and downloads binary distribution (no need to build from source).
--- a/.changeset/fix_uniffi_android_package_build.md
+++ b/.changeset/fix_uniffi_android_package_build.md
@@ -1,0 +1,5 @@
+---
+livekit-uniffi: patch
+---
+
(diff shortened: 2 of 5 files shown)


pblazej:
🟢 build (manual) https://github.com/livekit/rust-sdks/actions/runs/30533315426/job/90840523907

lukasIO (approved):
lgtm

.github/actions/uniffi-deps/action.yml
```diff
@@ -5,8 +5,3 @@ runs:
     - name: Install cargo-make
       uses: taiki-e/install-action@682e7d9e49c5e653d371fc6adbda67653461378a # v2.82.4
       with: { tool: cargo-make@0.37 }
-    - name: Install tera-cli
```
ladvoc, line 8: question: Why was this removed? This will mean Tera CLI has to be built from source on every CI run.

pblazej: it installed the wrong binary in practice: 

```

+ tera --env-only -t ./support/swift/Package.swift.tera -o LiveKitUniFFI/Package.swift
error: Found argument '--env-only' which wasn't expected, or isn't valid in this context
	Did you mean --env?

USAGE:
    tera <--toml <toml>|--json <json>|--yaml <yaml>|--env>

For more information try --help

```

https://github.com/livekit/rust-sdks/actions/runs/30532222462/job/90836993120

davidliu (commented):
👍

PR #1307, opened 2026-08-03
fix(CI): gate changeset action on unmanaged crates

lukasIO:
The changeset check derived its list of packages from `knope.toml` alone. A
crate that wasn't listed there was invisible to the whole release pipeline: no
changeset was ever required for it, and knope never released it. Nothing noticed
when a new publishable crate was added to the workspace but not wired into
`knope.toml`.

This adds a reconciliation gate between the Cargo workspace and `knope.toml`,
using the `publish` field from `cargo metadata` as the source of truth.

### Changes

- **`changeset_detect.py`**: new `reconcile_knope_config()` that reports two
  lists:
  - `unmanaged` — publishable crates (not `publish = false`) missing from
    `knope.toml`.
  - `stale` — `knope.toml` entries with no matching workspace crate (e.g. a
    renamed or removed crate).

  The rule is one-directional: a publishable crate must be knope-managed, but a
  knope-managed crate may set `publish = false` (e.g. `livekit-ffi` /
  `livekit-uniffi`, which CI publishes via wrapper packages).
- **`changeset-check.yml`**: fails the check with an explanatory PR comment when
  either list is non-empty. Each unmanaged crate must be fixed by adding a
  `[packages.<name>]` block, or by setting `publish = false`. Runs before the
  "no versioned packages affected" early-out so a brand-new crate can't slip
  through as "nothing affected".
- **`knope.toml`**: registers `livekit-runtime`, which the new gate surfaced as
  a pre-existing gap — it is publishable but was never knope-managed.
- **`test_changeset_detect.py`**: tests covering unmanaged, stale,
  `publish = false` exclusion, knope-managed `publish = false`, and restricted
  registries.

### DIFF
--- a/.github/scripts/changeset_detect.py
+++ b/.github/scripts/changeset_detect.py
@@ -35,6 +35,8 @@ Output (JSON on stdout):
     "present":           [[pkg, bump], ...], # bumps already in the changeset
     "missing":           [pkg, ...],         # required packages not yet covered
     "invalid":           [{...}, ...],       # bump lines in a format knope would ignore
+    "unmanaged":         [pkg, ...],         # publishable crates missing from knope.toml
+    "stale":             [pkg, ...],         # knope.toml entries with no workspace crate
     "changeset_content": str,               # a changeset prefilled with `missing`
   }
 """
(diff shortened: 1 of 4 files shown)


1egoman (approved):

ladvoc (approved):
LGTM ✅

PR #1312, opened 2026-08-04
Route unmatched H264 formats to platform decoder factories

MaxHeimbrock:
The SFU now includes H.264 Baseline (42001f) in its default codec set and selects the forwarded codec from the subscriber's SDP. Create() matched platform factories with IsSameCodec, which is strict on profile-level-id; VideoToolbox only advertises 42e01f/640c1f, so 42001f fell through to the internal FFmpeg decoder, which desktop prebuilts cannot initialize (no H.264 codec compiled in) — every frame failed to decode and subscribers saw black video in a PLI loop.

Hardware decoders handle Baseline/Main/High regardless of the exact fmtp they advertise, so when no exact or packetization-mode-relaxed match exists, create the decoder from the platform factory's own advertised H.264 format instead of falling through.

Solving [issue description](https://claude.ai/code/artifact/b7ae7e12-a6ff-4c5f-9856-62c0ee962070?via=auto_preview) client side

### DIFF
--- a/webrtc-sys/src/video_decoder_factory.cpp
+++ b/webrtc-sys/src/video_decoder_factory.cpp
@@ -125,6 +125,25 @@ std::unique_ptr<webrtc::VideoDecoder> VideoDecoderFactory::Create(
           return factory->Create(env, adjusted);
       }
     }
+
+    // Still no match: the profile-level-id differs from everything the
+    // platform factories advertise (e.g. the SFU sends Baseline 42001f but
+    // VideoToolbox only lists 42e01f/640c1f). Hardware decoders handle
+    // Baseline/Main/High regardless of the exact fmtp they advertise, so
+    // relax the profile too and create from the factory's own advertised
+    // format rather than falling through to the internal FFmpeg decoder,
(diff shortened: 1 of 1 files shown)


MaxHeimbrock:
From more digging:

The theoretical bad combination. The new fallback sits above the internal webrtc::H264Decoder::Create() path in Create(). So any H.264 format that doesn't strictly match a platform factory now goes to the platform factory — the internal decoder is never consulted for H.264 again on platforms that have one. 

That's harmless today because the internal decoder is broken on every desktop prebuilt (that's the bug). But picture a build where FFmpeg H.264 actually works (Option C, or some future prebuilt): Fix B's probe passes, so we advertise the internal list — which includes High 4:4:4 Predictive. 

If such a stream arrived, strict match against VideoToolbox fails, the fallback fires, and VideoToolbox gets a 4:4:4 stream it very likely can't decode — while a working FFmpeg sits unused. That's the one case where A regresses relative to no-A, and it requires a working internal decoder plus an exotic profile, neither of which exists in current desktop builds.

PR #1313, opened 2026-08-04
Only advertise internal H264 decode formats if the decoder works

MaxHeimbrock:
GetSupportedFormats() unconditionally appended SupportedH264DecoderCodecs(), so desktop clients claimed 42001f/42e01f decode support in their SDP even though the internal FFmpeg decoder cannot initialize on desktop prebuilts (H.264 codec not compiled in). The SFU then selected Baseline for those subscribers, producing undecodable streams.

H264Decoder::IsSupported() only reflects the WEBRTC_USE_H264 build flag — the missing FFmpeg codec surfaces first when Configure() fails at runtime. Probe Create()+Configure() once at factory construction and advertise the internal formats only when the probe succeeds. Platform-factory formats (VideoToolbox, MediaCodec, NVDEC) are unaffected.

Also gate the internal H264Decoder::Create() fallback in Create() on the same probe, so a format the platform factories reject yields a clear "No VideoDecoder found" error instead of a decoder that fails Configure() at runtime.

Solving [issue description](https://claude.ai/code/artifact/b7ae7e12-a6ff-4c5f-9856-62c0ee962070?via=auto_preview) client side

### DIFF
--- a/webrtc-sys/include/livekit/video_decoder_factory.h
+++ b/webrtc-sys/include/livekit/video_decoder_factory.h
@@ -35,5 +35,6 @@ class VideoDecoderFactory : public webrtc::VideoDecoderFactory {
 
  private:
   std::vector<std::unique_ptr<webrtc::VideoDecoderFactory>> factories_;
+  bool internal_h264_decoder_works_ = false;
 };
 }  // namespace livekit_ffi
--- a/webrtc-sys/include/livekit/video_decoder_factory.h
+++ b/webrtc-sys/include/livekit/video_decoder_factory.h
@@ -35,6 +35,6 @@ class VideoDecoderFactory : public webrtc::VideoDecoderFactory {
 
  private:
   std::vector<std::unique_ptr<webrtc::VideoDecoderFactory>> factories_;
-  bool internal_h264_decoder_works_ = false;
+  const bool internal_h264_decoder_works_;
 };
 }  // namespace livekit_ffi
--- a/webrtc-sys/src/video_decoder_factory.cpp
+++ b/webrtc-sys/src/video_decoder_factory.cpp
@@ -41,6 +41,24 @@
 
 namespace livekit_ffi {
 
+namespace {
+// H264Decoder::IsSupported() only reflects the WEBRTC_USE_H264 build flag;
+// desktop prebuilts link an FFmpeg without the H.264 codec, which only
+// surfaces when Configure() fails at runtime ("FFmpeg H.264 decoder not
+// found"). Probe once so the SDP does not advertise decode support the
+// internal decoder cannot deliver.
+bool InternalH264DecoderWorks() {
+  if (!webrtc::H264Decoder::IsSupported())
+    return false;
+  auto decoder = webrtc::H264Decoder::Create();
+  if (!decoder)
+    return false;
+  webrtc::VideoDecoder::Settings settings;
+  settings.set_codec_type(webrtc::kVideoCodecH264);
+  return decoder->Configure(settings);
+}
+}  // namespace
+
 VideoDecoderFactory::VideoDecoderFactory() {
 #ifdef __APPLE__
   factories_.push_back(livekit_ffi::CreateObjCVideoDecoderFactory());
@@ -55,6 +73,12 @@ VideoDecoderFactory::VideoDecoderFactory() {
     factories_.push_back(std::make_unique<webrtc::NvidiaVideoDecoderFactory>());
   }
 #endif
+
+  internal_h264_decoder_works_ = InternalH264DecoderWorks();
+  if (!internal_h264_decoder_works_) {
+    RTC_LOG(LS_WARNING) << "Internal H264 decoder is unavailable, "
+                           "not advertising its formats";
+  }
 }
 
 std::vector<webrtc::SdpVideoFormat> VideoDecoderFactory::GetSupportedFormats()
@@ -71,9 +95,11 @@ std::vector<webrtc::SdpVideoFormat> VideoDecoderFactory::GetSupportedFormats()
   for (const webrtc::SdpVideoFormat& format :
        webrtc::SupportedVP9DecoderCodecs())
     formats.push_back(format);
-  for (const webrtc::SdpVideoFormat& h264_format :
-       webrtc::SupportedH264DecoderCodecs())
-    formats.push_back(h264_format);
+  if (internal_h264_decoder_works_) {
+    for (const webrtc::SdpVideoFormat& h264_format :
+         webrtc::SupportedH264DecoderCodecs())
+      formats.push_back(h264_format);
+  }
 
   formats.push_back(webrtc::SdpVideoFormat(
       webrtc::SdpVideoFormat::AV1Profile0(),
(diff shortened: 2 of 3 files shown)


webrtc-sys/src/video_decoder_factory.cpp
```diff
@@ -55,6 +73,12 @@ VideoDecoderFactory::VideoDecoderFactory() {
     factories_.push_back(std::make_unique<webrtc::NvidiaVideoDecoderFactory>());
   }
 #endif
+
+  internal_h264_decoder_works_ = InternalH264DecoderWorks();
```
MaxHeimbrock: During construction of the factory check if at runtime there is a h264 decoder present from this lib at runtime.

webrtc-sys/src/video_decoder_factory.cpp
```diff
@@ -71,9 +95,11 @@ std::vector<webrtc::SdpVideoFormat> VideoDecoderFactory::GetSupportedFormats()
   for (const webrtc::SdpVideoFormat& format :
        webrtc::SupportedVP9DecoderCodecs())
     formats.push_back(format);
-  for (const webrtc::SdpVideoFormat& h264_format :
-       webrtc::SupportedH264DecoderCodecs())
-    formats.push_back(h264_format);
+  if (internal_h264_decoder_works_) {
```
MaxHeimbrock, line 108: Filter the supported formats based on if a working h264 decoder is present in this lib at runtime.

webrtc-sys/src/video_decoder_factory.cpp
```diff
@@ -131,7 +157,8 @@ std::unique_ptr<webrtc::VideoDecoder> VideoDecoderFactory::Create(
     return webrtc::CreateVp8Decoder(env);
   if (absl::EqualsIgnoreCase(format.name, webrtc::kVp9CodecName))
     return webrtc::VP9Decoder::Create();
-  if (absl::EqualsIgnoreCase(format.name, webrtc::kH264CodecName))
+  if (absl::EqualsIgnoreCase(format.name, webrtc::kH264CodecName) &&
```
MaxHeimbrock, line 171: If at runtime for whatever reason we try to create the decoder and it is not supported, error

webrtc-sys/src/video_decoder_factory.cpp
MaxHeimbrock: Runtime costs for this new code: 

- On the broken desktop builds (the case that matters): the H264DecoderImpl constructor is trivial member initialization. Configure() allocates one AVCodecContext (a few KB), sets fields, then avcodec_find_decoder() — a lookup in FFmpeg's registered-codec table — returns null, and the failure path immediately calls Release(), freeing the context. Microseconds of latency, a few KB of transient heap.
 
- On builds where FFmpeg does have H.264: it additionally runs avcodec_open2(), which initializes the decoder's private context and tables — roughly single-digit milliseconds and some transient allocations. Frame buffers are not allocated here; those come lazily via get_buffer2 during actual decoding. The probe's unique_ptr destructor then runs Release() and everything is freed before InternalH264DecoderWorks() returns.

webrtc-sys/include/livekit/video_decoder_factory.h
```diff
@@ -35,5 +35,6 @@ class VideoDecoderFactory : public webrtc::VideoDecoderFactory {
 
  private:
   std::vector<std::unique_ptr<webrtc::VideoDecoderFactory>> factories_;
+  bool internal_h264_decoder_works_ = false;
```
xianshijing-lk: nit, const bool internal_h264_decoder_works_;

webrtc-sys/src/video_decoder_factory.cpp
```diff
@@ -41,6 +41,24 @@
...
+namespace {
+// H264Decoder::IsSupported() only reflects the WEBRTC_USE_H264 build flag;
+// desktop prebuilts link an FFmpeg without the H.264 codec, which only
+// surfaces when Configure() fails at runtime ("FFmpeg H.264 decoder not
+// found"). Probe once so the SDP does not advertise decode support the
+// internal decoder cannot deliver.
+bool InternalH264DecoderWorks() {
+  if (!webrtc::H264Decoder::IsSupported())
+    return false;
+  auto decoder = webrtc::H264Decoder::Create();
```
xianshijing-lk, line 56: curiously, what is the return type of webrtc::H264Decoder::Create() ? is it a smart pointer holding a H264Decoder ?

webrtc-sys/src/video_decoder_factory.cpp
```diff
@@ -41,6 +41,24 @@
 
 namespace livekit_ffi {
 
+namespace {
+// H264Decoder::IsSupported() only reflects the WEBRTC_USE_H264 build flag;
+// desktop prebuilts link an FFmpeg without the H.264 codec, which only
+// surfaces when Configure() fails at runtime ("FFmpeg H.264 decoder not
+// found"). Probe once so the SDP does not advertise decode support the
+// internal decoder cannot deliver.
+bool InternalH264DecoderWorks() {
```
xianshijing-lk: nit, rename it to IsInternalH264DecoderAvailable() ?

webrtc-sys/src/video_decoder_factory.cpp
```diff
@@ -41,6 +41,24 @@
...
+  if (!webrtc::H264Decoder::IsSupported())
+    return false;
```
xianshijing-lk, line 54: I wonder if you should print out the warnings or error logs for vairous reasons

webrtc-sys/src/video_decoder_factory.cpp
```diff
@@ -71,9 +95,11 @@ std::vector<webrtc::SdpVideoFormat> VideoDecoderFactory::GetSupportedFormats()
   for (const webrtc::SdpVideoFormat& format :
        webrtc::SupportedVP9DecoderCodecs())
     formats.push_back(format);
-  for (const webrtc::SdpVideoFormat& h264_format :
-       webrtc::SupportedH264DecoderCodecs())
-    formats.push_back(h264_format);
+  if (internal_h264_decoder_works_) {
+    for (const webrtc::SdpVideoFormat& h264_format :
+         webrtc::SupportedH264DecoderCodecs())
+      formats.push_back(h264_format);
```
xianshijing-lk, line 111: nit, use {}
for (const webrtc::SdpVideoFormat& h264_format : webrtc::SupportedH264DecoderCodecs()) {
  formats.push_back(h264_format);
}

xianshijing-lk (approved):
lgtm, some minor comments.

webrtc-sys/src/video_decoder_factory.cpp
MaxHeimbrock: Yes, a smart pointer, so the probe decoder is destroyed (and Release()d) automatically when IsInternalH264DecoderAvailable() returns.

webrtc-sys/src/video_decoder_factory.cpp
MaxHeimbrock: Done

webrtc-sys/src/video_decoder_factory.cpp
MaxHeimbrock: Done

webrtc-sys/include/livekit/video_decoder_factory.h
MaxHeimbrock: Done

webrtc-sys/src/video_decoder_factory.cpp
MaxHeimbrock: Done

PR #1314, opened 2026-08-04
potential fix: av foundation take the duration from the matched range…

stephen-derosa:


ladvoc (approved):
Makes sense to me! What version of macOS did you test on?

stephen-derosa:
> Makes sense to me! What version of macOS did you test on?

26.5.2 (25F84)!

PR #1317, opened 2026-08-06
other_sdks

stephen-derosa:


livekit-api/src/signal_client/mod.rs
```diff
@@ -147,11 +147,15 @@ pub enum SignalError {
 pub struct SignalSdkOptions {
     pub sdk: String,
     pub sdk_version: Option<String>,
+    /// Comma separated list of additional LiveKit SDKs layered on top of this one, with
+    /// versions, e.g. `"components-js:1.2.3,track-processors-js:1.2.3"`. Sent to the server
+    /// as `ClientInfo.other_sdks`. Empty when there are none.
+    pub other_sdks: String,
```
ladvoc: suggestion: While this is in the public API by necessity, it isn't something that should appear in generated docs so it is clear it is for internal use; I would recommend annotating with `#[doc(hidden)]` and removing the doc comment.

ladvoc: suggestion: Worth making this an `Option<String>` to avoid an allocation when this isn't used.

ladvoc (approved):
LGTM after comments are addressed. I think this is a good solution to avoid a breaking C ABI change that would require updating all FFI clients. Let's wait for other reviewers to get eyes on this before merging.

xianshijing-lk:
could you please add some description to this PR to explain what it doe ?

livekit-api/src/signal_client/mod.rs
stephen-derosa: done!

stephen-derosa: done!

livekit-api/src/signal_client/mod.rs
```diff
@@ -147,11 +147,16 @@ pub enum SignalError {
 pub struct SignalSdkOptions {
     pub sdk: String,
     pub sdk_version: Option<String>,
+    /// Comma separated list of additional LiveKit SDKs layered on top of this one, with
+    /// versions, e.g. `"components-js:1.2.3,track-processors-js:1.2.3"`. Sent to the server
+    /// as `ClientInfo.other_sdks`. Empty when there are none.
+    /// #[doc(hidden)]
```
ladvoc: issue: This is an attribute, so it shouldn't be commented:
```rust
#[doc(hidden)]
pub other_sdks: Option<String>,
```

.changeset/other_sdks_propagation.md
```diff
@@ -0,0 +1,9 @@
+---
+livekit: minor
+livekit-api: minor
+livekit-ffi: minor
+---
+
+Add `other_sdks` field to propagate additional SDK metadata to the server. 
+
+Allows C++ host libraries and other FFI consumers to declare additional LiveKit SDKs layered on top (e.g., "components-js:1.2.3,track-processors-js:1.2.3"), which is forwarded to `ClientInfo.other_sdks` in both v1 and v0 signaling paths and visible to the data stack. The field is optional (defaults empty) and backward compatible with older servers.
```
ladvoc: suggestion: This is too detailed for a changelog entry.

livekit-api/src/signal_client/mod.rs
stephen-derosa: fixed

xianshijing-lk (approved):

PR #1318, opened 2026-08-06
Caching added similar to Swift

MaxHeimbrock:
## Background

Stacked cached token source in Rust, similar to swift: https://github.com/livekit/client-sdk-swift/blob/main/Sources/LiveKit/Token/CachingTokenSource.swift

### DIFF
--- a/livekit-token-source/src/caching.rs
+++ b/livekit-token-source/src/caching.rs
@@ -1,0 +1,140 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Mutex;
+
+use async_trait::async_trait;
+
+use crate::request::TokenSourceFetchOptions;
+use crate::response::{TokenSourceResponse, TokenSourceResult};
+use crate::token_source::TokenSourceConfigurable;
+
+/// Persistence backend for [`TokenSourceCached`].
+///
+/// Implement this trait to keep credentials in a custom location, e.g. the
+/// platform keychain or a database. The default is the process-lifetime
+/// [`TokenSourceInMemoryStore`].
+#[async_trait]
+pub trait TokenSourceStore: Send + Sync {
+    /// Stores the given credentials, replacing any stored previously.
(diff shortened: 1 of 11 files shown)


livekit-token-source/src/caching.rs
```diff
@@ -0,0 +1,142 @@
...
+        F: Fn(&TokenSourceFetchOptions, &TokenSourceResponse) -> bool + Send + Sync + 'static,
+    {
+        self.validator = Box::new(validator);
+        self
+    }
+
+    /// Removes the stored credentials, forcing the next fetch to hit the
+    /// underlying source.
+    pub async fn invalidate(&self) {
+        self.store.clear().await;
```
xianshijing-lk, line 119: what will happen if a fetch is ongoing here ?

livekit-token-source/src/caching.rs
```diff
@@ -0,0 +1,142 @@
...
+        &self,
+        options: &TokenSourceFetchOptions,
+    ) -> TokenSourceResult<TokenSourceResponse> {
+        if let Some((cached_options, cached_response)) = self.store.retrieve().await {
+            if cached_options == *options && (self.validator)(&cached_options, &cached_response) {
+                return Ok(cached_response);
+            }
+        }
+
+        let response = self.source.fetch(options).await?;
```
xianshijing-lk, line 141: curiously, can multiple fetch() be called at the same time ? I wonder if we will need to protect such corner case.

livekit-token-source/tests/caching.rs
```diff
@@ -0,0 +1,287 @@
...
+    ) -> (Self, Arc<AtomicUsize>, Arc<Mutex<Option<TokenSourceResponse>>>) {
+        let calls = Arc::new(AtomicUsize::new(0));
+        let next = Arc::new(Mutex::new(Some(response(token))));
+        (Self { calls: calls.clone(), next: next.clone() }, calls, next)
+    }
+}
+
+#[async_trait]
+impl TokenSourceConfigurable for CountingSource {
+    async fn fetch(
```
xianshijing-lk, line 80: nit, should this fetch() has a .await to mimic the real use case ?

xianshijing-lk (approved):
some nits and questions, lgtm in general

livekit-token-source/src/caching.rs
MaxHeimbrock: Added documentation:
```
/// A fetch already in flight is unaffected: it still resolves and stores
/// its response afterwards, repopulating the cache (last writer wins).
```

livekit-token-source/src/caching.rs
MaxHeimbrock: Yes, and we don't protect against this, same as in Swift btw. This would be a bigger design. I think as other SDKs are not that advanced either yet, maybe we keep it in mind as a follow up.

livekit-token-source/tests/caching.rs
MaxHeimbrock: In these tests it would not change the behaviour, it would still run in sequence.

I added tests though for a token source that can mimic parallel in flight requests to test the existing contract of `last writer wins`.

PR #1321, opened 2026-08-07
Setup trusted publishing

ladvoc:
The following crates are currently published to crates.io and have had [trusted publishing](https://crates.io/docs/trusted-publishing) configured:
- device-info
- imgproc 
- libwebrtc 
- livekit
- livekit-api 
- livekit-common
- livekit-data-stream 
- livekit-datatrack 
- livekit-net 
- livekit-protocol 
- livekit-runtime
- livekit-wakeword
- soxr-sys 
- webrtc-sys
- webrtc-sys-build 
- yuv-sys 

Note: when a new crate is added, it must also manually have trusted publishing enabled and published manually; subsequent releases will be handled by CI automatically.

Follow up: after the first successful release using trusted publishing, remove the `CARGO_TOKEN` from repo secrets and enable "Require trusted publishing for all new versions" for all crates.

Closes CLT-3231

### DIFF
--- a/.github/scripts/check_crates_published.sh
+++ b/.github/scripts/check_crates_published.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# Copyright 2026 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Make sure that every publishable crate in the workspace exists on crates.io.
+# Trusted publishing cannot create a new crate. A new crate needs one manual
+# publish before the release runs.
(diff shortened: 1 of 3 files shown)


lukasIO:
should we have a release gate that ensures all packages are already published to crates.io before proceeding to the publish step? 

we did this on the agents-js to ensure we don't fail mid release for new packages

1egoman:
> should we have a release gate that ensures all packages are already published to crates.io before proceeding to the publish step?

Yea, I think this is a really good idea. It would be potentially quite tricky to recover if the packages that make up a release are half rolled out and half not.

1egoman:
@ladvoc Two other suggestions:

1. I think it would be good before this is merged also to add some context to the "Client SDK Release Instructions" internal notion document outlining what one needs to do to add a new package to capture the context here. I know I am very likely to forget how this all works unless it is written down somewhere.

2. If this isn't hard, it might be good to have a callout in the release pull request that knope makes (maybe a github actions created comment?) that says something like "Here's a list of new packages, make sure you create these manually before merging this". That would also help ensure that a partial release isn't done accidentally.

ladvoc:
CI check for new, unpublished crates added in [b6e7d0b](https://github.com/livekit/rust-sdks/pull/1321/commits/b6e7d0b4c62cabba9adf9ca4448ae52b57f5771f). Internal docs updated to reflect how to manually publish new crates and enable trusted publishing.

lukasIO (approved):

1egoman (approved):

PR #1323, opened 2026-08-10
Automate the Dart/Flutter release for livekit-uniffi

hiroshihorie:
### Before you submit your PR

### PR description

Automates the Dart/Flutter half of the livekit-uniffi release (CLT-2872), the release-side prerequisite for livekit/client-sdk-flutter#1160.

- **Make the generated package publishable.** Release builds omit the dev dylib and `publish_to`, gain LICENSE/README/CHANGELOG and real version constraints, and fail closed for unknown profiles. A `dart-clean` task keeps stale host dylibs out of release packages, and a shipped `analysis_options.yaml` silences lints that only fire on generated code.
- **Re-enable the Dart tests** disabled since #1034. The uniffi-dart pin on main already carries the multi-crate codegen fixes from upstream, so this PR no longer changes the pin.
- **Add `uniffi-dart-publish.yml`**, a tag-push workflow (pub.dev only accepts OIDC tokens from tag-push runs). It resolves and validates the tag against the crate version, builds every `build-<triple>.zip` through the reusable `uniffi-cdylib.yml` and attaches them to the release, then builds the Dart package from a copy outside the work tree (`packages/` is gitignored and pub archives from git's file listing), validates it, and publishes. The publish job depends on the asset job, so there is nothing to poll and a package can never publish without its assets.
- **Harden `uniffi-cdylib.yml`.** It checks out the release tag rather than the dispatch ref, builds the Linux libraries inside manylinux_2_28 containers (glibc 2.28 floor, where ubuntu-latest gives 2.39 and breaks Debian 12 / Ubuntu 22.04 / AL2023), builds aarch64 Linux on a native arm64 runner, and installs cargo-ndk as a prebuilt binary.
- **Pin the Dart SDK** to 3.13.1 in both the test and publish workflows so PR CI validates the SDK the publish runs with.

**Publishing stays disabled** (`PUBLISH_ENABLED=false`, every run stops at `--dry-run`) until the first manual publish creates the package on pub.dev and automated publishing is configured. Runbook in the workflow header.

### Breaking changes

None. Dev-profile `cargo make dart-package` behaves as before.

### MSRV

No changes.

### Testing

- Dev build: all 5 Dart FFI tests pass, same flow now re-enabled in CI.
- Release build: `dart pub publish --dry-run` exits 0 with zero warnings and no dylib in the package.
- Full release matrix dry run from this branch against the v0.1.9 tag: all 12 cdylib targets plus the Swift and Android jobs passed with `dry_run=true` (https://github.com/livekit/rust-sdks/actions/runs/34203702564). The CI-built Linux libraries were downloaded and verified: bare `.so` at the zip root, sha256 sidecars match, highest required symbol version GLIBC_2.28 on both architectures.
- Consumer: livekit/client-sdk-flutter#1160's uniffi tests pass against the generated package.
- Not exercisable before merge: the publish workflow's own job wiring, because GitHub only registers dispatchable workflows from the default branch. Plan is to dispatch it from main with `dry_run=true` before the next livekit-uniffi release. Whether knope's API-created tag fires the push trigger is confirmed only by the first real release. Both cost nothing while publishing is disabled.

### Async

No async code added; the crate surface is untouched.


livekit-uniffi/support/dart/README.md
```diff
@@ -0,0 +1,40 @@
+# livekit_uniffi
+
+Dart bindings for the LiveKit Rust SDK core, generated from the
+[`livekit-uniffi`](https://github.com/livekit/rust-sdks/tree/main/livekit-uniffi)
+crate with [UniFFI](https://mozilla.github.io/uniffi-rs/) and
+[uniffi-dart](https://github.com/Uniffi-Dart/uniffi-dart).
+
+This is a low-level package. It is consumed by
```
ladvoc, line 8: Nice callout! We should make sure this is included in all the READMEs for the package repos (outside the scope of this PR).

.github/workflows/uniffi-dart-publish.yml
```diff
@@ -0,0 +1,203 @@
...
+      tag_name:
+        description: "Release tag (e.g. livekit-uniffi/v0.1.9)"
+        required: true
+        type: string
+
+env:
+  # Flip to "true" once the enablement steps in the header are done. Real
+  # publishing additionally requires a tag-push trigger; workflow_dispatch
+  # runs always stop at the dry run (pub.dev rejects their OIDC tokens).
+  PUBLISH_ENABLED: "false"
```
ladvoc, line 49: question: Might it make sense to turn this into a workflow input like some of the other workflows do?

.github/workflows/uniffi-dart-publish.yml
```diff
@@ -0,0 +1,203 @@
...
+          restore-keys: |
+            ${{ runner.os }}-cargo-target-dart-
+
+      - name: Install Common Deps for UniFFI Tasks
+        uses: ./.github/actions/uniffi-deps
+
+      - name: Setup Dart
+        uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260 # v1.7.2
+        with:
+          sdk: stable
```
ladvoc, line 131: question: Should this pin to a specific release?

ladvoc (approved):

.github/workflows/uniffi-dart-publish.yml
```diff
@@ -0,0 +1,203 @@
...
+      # this workflow's failed run (or dispatch it with the tag).
+      - name: Wait for cdylib release assets
+        timeout-minutes: 90
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ steps.tag.outputs.tag_name }}
+        run: |
+          expected=12
+          missing=0
+          while true; do
```
pblazej, line 173: Based on my fails with swift it's not a good idea to poll, maybe there's another way to trigger it _after_ dylib update? Have you tested that e2e?

livekit-uniffi/support/dart/README.md
hiroshihorie: Agreed, tracking that separately. Out of scope here.

.github/workflows/uniffi-dart-publish.yml
hiroshihorie: An input would be inert here. pub.dev only accepts OIDC tokens from tag-push runs, which have no inputs, and dispatch runs can never publish regardless. Kept the env constant as the gate.

.github/workflows/uniffi-dart-publish.yml
hiroshihorie: Done, pinned to 3.13.1 here and in uniffi-dart-test.yml so PR CI validates the same SDK the publish runs with. Bump both together.

.github/workflows/uniffi-dart-publish.yml
hiroshihorie: Fair, and no it was not tested e2e. Removed the polling instead: this workflow now calls uniffi-cdylib.yml itself and the publish job needs it, so assets and package come from one tag-push run. Full matrix dry run from this branch against v0.1.9 passed: https://github.com/livekit/rust-sdks/actions/runs/34203702564. The publish workflow itself can only be dispatched once it is on main (GitHub registers workflows from the default branch), so the plan is a dry_run dispatch from main before the 0.2.0 release.

PR #1324, opened 2026-08-10
add connection test cli tool

chenosaurus:
- Add a simple CLI tool to test connection to LK Cloud & report the method


examples/connection_test/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[[bin]]
+name = "livekit-connection-test"
+path = "src/main.rs"
+
+[dependencies]
+clap = { version = "4.5", features = ["derive"] }
```
ladvoc, line 12: issue: `livekit`, `clap`, and `tokio` are workspace deps, set `workspace = true` rather than hardcoding version or path.

PR #1328, opened 2026-08-13
Update schema metadata validation logic

ladvoc:
No longer allow `Other` to be used at publish time and allow custom encodings to be used with well-known ones.

### DIFF
--- a/livekit-datatrack/src/schema.rs
+++ b/livekit-datatrack/src/schema.rs
@@ -114,6 +114,8 @@ pub enum DataTrackSchemaEncoding {
     JsonSchema,
 
     /// Another well-known encoding not known to this client version.
+    ///
+    /// Only produced for received tracks; cannot be used when publishing.
     Other,
     /// An application-specific encoding identified by the contained string.
     ///
@@ -163,6 +165,8 @@ pub enum DataTrackFrameEncoding {
     Json,
 
     /// Another well-known encoding not known to this client version.
+    ///
+    /// Only produced for received tracks; cannot be used when publishing.
     Other,
     /// An application-specific encoding identified by the contained string.
     ///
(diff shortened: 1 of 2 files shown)


livekit-datatrack/src/schema.rs
```diff
@@ -186,14 +190,24 @@ pub enum DataTrackSchemaError {
     /// Specified schema and frame encodings are incompatible.
     #[error("Specified schema and frame encodings are incompatible")]
     Incompatible,
+
+    /// The 'other' encoding represents an unrecognized encoding on received tracks
+    /// and cannot be used when publishing.
+    #[error("The 'other' encoding cannot be used when publishing")]
+    OtherEncoding,
```
1egoman, line 197: nitpick(non-blocking): I wonder if there's a way to do this where you aren't exposing `::Other` in the publishing context by splitting up this `DataTrackSchemaEncoding` into a publishing variant and a subscribing variant. I don't think there would be a way to do this where it wouldn't be a breaking change though, so maybe disregard if you see to many problems with this idea.

livekit-datatrack/src/schema.rs
```diff
@@ -186,14 +190,24 @@ pub enum DataTrackSchemaError {
...
 }
 
 /// Validates that the given frame and schema encodings are compatible.
 pub(crate) fn validate_schema(
     frame_encoding: Option<&DataTrackFrameEncoding>,
     schema_encoding: Option<&DataTrackSchemaEncoding>,
 ) -> Result<(), DataTrackSchemaError> {
+    use DataTrackFrameEncoding as FrameEncoding;
+    use DataTrackSchemaEncoding as SchemaEncoding;
```
1egoman, line 206: nitpick: Why alias these here rather than just the fully qualified names below? IMO, it adds some confusion because it's not 100% clear that `FrameEncoding` and `DataTrackFrameEncoding` / etc mean the same thing if you enter the file from further down.

1egoman (approved):

PR #1332, opened 2026-08-16
fix(rtc_engine): escalate a PeerConnection that stays disconnected

xianshijing-lk:
### Before you submit your PR

> **Stacked on #1331.** Targets `sxian/CLT-3249/resume-requires-pc-recovery-evidence`; review that one first. It touches the same `RtcEvent::ConnectionChange` handler, so it does not apply cleanly to `main` on its own.

### PR description

Only `PeerConnectionState::Failed` drove recovery. libwebrtc does not reach `Failed` until ICE consent expires — tens of seconds after a transport actually stops working — while `Disconnected`, which it reports within a couple of seconds, was ignored entirely. Nothing acted in between, so a session whose media plane had died stayed "connected" and silently deaf for the whole consent window before anything began recovering.

The clearest case is **steady state**, with no signalling failure at all: a NAT rebind, a Wi-Fi→cellular handover, a route change. There is no resume in flight to notice, `Disconnected` is dropped on the floor, and recovery does not begin until `Failed` finally arrives.

It also lengthened node-failure recovery, though #1331 already removes the worst of that: with the resume no longer accepting a stale `Connected`, it rejects the dead transport and escalates at its own bound. This PR shortens that further and, more importantly, covers the case where nothing is waiting.

#### Approach

`Disconnected` is deliberately *not* treated as an immediate failure: brief disconnects during ordinary network disturbance are normal and self-healing, and tearing down a session for one would be worse than the disturbance itself. So a transport entering `Disconnected` starts a grace period (`PC_DISCONNECTED_GRACE`, 5s — comfortably past a transient blip, far short of consent expiry) and is judged on **its state when the countdown elapses**, not on the transition that started it.

Judging on the final state rather than tracking the transition has a nice property: a connection that recovered on its own needs no cancellation bookkeeping at all. It simply reads as connected when the countdown lapses, and nothing happens. Repeated `Disconnected` transitions collapse onto the countdown already running rather than each spawning their own, so a flapping transport cannot accumulate timers that all fire and re-report the same failure.

Escalation is delegated to the engine rather than decided locally, because the engine already interprets a transport failure in context: outside a reconnect it starts one; during a reconnect its existing sticky-escalation logic converts the cycle to a full reconnect instead of looping on resume. `Failed` is left to the existing branch, so it is still reported the instant it occurs and is not double-reported by the countdown.

#### Comparison with client-sdk-js

Worth recording, because it is not a straight port: **client-sdk-js does not act on `disconnected` either.** `PCTransportManager.updateState` has no branch for it — none of its conditions match, so `this.state` silently retains its previous value — which means `verifyTransport` keeps accepting the transport and the connection-reconcile loop cannot catch it.

JS's actual net for this case is server-driven: `ConnectionQuality::LOST` on the local participant → `scheduleLostQualityReconnect` (`RTCEngine.ts:1207`), which is the same arm/re-check/guard shape used here. That signal catches a **publisher** the server has stopped receiving, but not a **subscriber** that has stopped receiving the server. Acting locally on `disconnected` covers both directions.

Rust currently forwards `ConnectionQuality` to the application and does nothing with it (`rtc_engine/mod.rs`), so adopting the JS-style LOST trigger as well would be a reasonable complementary follow-up — the two are not redundant.

### Breaking changes

None. No public API change; `PC_DISCONNECTED_GRACE` is new and additive.

Behaviourally, a transport that stays disconnected for 5s now triggers recovery where previously it was ignored until `Failed`. That is the point of the change, but it does mean sessions on genuinely marginal links may now reconnect where they previously sat in a degraded-but-quiet state. The grace period is the tuning point if that proves too eager in the field.

### MSRV

Unchanged.

### Testing

`cargo test -p livekit --lib` — 88 passed.

The escalation decision is split into `should_escalate_after_grace(state)` so each branch is asserted directly, without waiting out a real countdown:

- `transport_still_down_after_grace_is_escalated` — the defect itself: `Disconnected` (and `Connecting`, i.e. never got back) at the end of the window must escalate rather than wait for `Failed`.
- `transport_that_recovered_during_grace_is_left_alone` — the reason this is a countdown and not an immediate reaction; guards against regressing to "escalate on `Disconnected`".
- `failed_is_not_double_reported_after_grace` — `Failed` is already reported by the existing branch.
- `closed_transport_is_not_escalated_after_grace` — a deliberate teardown (session close, or a full reconnect replacing this session) must not be fought.

Plus, against a real `PeerConnection`:

- `disconnect_grace_admits_one_countdown_at_a_time` — asserts repeated transitions collapse onto one countdown, **and** that a fresh one can be armed afterwards, so a transport that recovers and dies again is still escalated. The second half is the one that matters: a dedupe flag that is never released would silently disable all subsequent detection.

Not covered by unit tests: the wall-clock behaviour of the timer itself, and the engine-side escalation it triggers (already covered by the existing reconnect tests and the `__lk-e2e-test` fault injection). End-to-end confirmation needs a live multi-node deployment; **I have not run it** — I do not have the credentials.

### Async

Uses `livekit_runtime::spawn` and `livekit_runtime::sleep`, no direct runtime dependency. `on_rtc_event` takes `self: &Arc<Self>` so the countdown can keep the session alive for its duration; it checks the session's `closed` flag on wake and bails, so a closed session does not escalate.

The delay here is not waiting for state to "catch up" — it is the substance of the fix. Reacting to `Disconnected` without it would reconnect on every transient blip. The two new unit tests are synchronous; the existing `#[tokio::test]` convention in these modules is unchanged.

### DIFF
--- a/livekit/src/rtc_engine/rtc_session.rs
+++ b/livekit/src/rtc_engine/rtc_session.rs
@@ -77,6 +77,21 @@ pub struct PcGenerationSnapshot {
 }
 
 pub const ICE_CONNECT_TIMEOUT: Duration = Duration::from_secs(15);
+
+/// How long a PeerConnection may sit in `Disconnected` before it is treated as a failure.
+///
+/// `Disconnected` is not by itself a reason to reconnect: ICE reports it after a couple of
+/// seconds without incoming packets, and ordinary network disturbances — a brief handover, a
+/// congested link — resolve on their own well inside that. But it is also the *only* early
+/// signal that a transport has died: libwebrtc will not escalate to `Failed` until consent
+/// expires, tens of seconds later, and until this existed nothing acted in between. A
+/// session whose media plane had silently gone away therefore stayed "connected" and deaf
+/// for the whole consent window before anything began recovering.
+///
+/// So we start a countdown instead of reacting immediately, and act only if the transport is
+/// still not connected when it elapses. Long enough to ride out a transient blip, far short
+/// of consent expiry.
+pub const PC_DISCONNECTED_GRACE: Duration = Duration::from_secs(5);
 pub const TRACK_PUBLISH_TIMEOUT: Duration = Duration::from_secs(10);
 pub const LOSSY_DC_LABEL: &str = "_lossy";
 pub const RELIABLE_DC_LABEL: &str = "_reliable";
@@ -1585,7 +1600,9 @@ impl SessionInner {
         Ok(())
     }
 
-    async fn on_rtc_event(&self, event: RtcEvent) -> EngineResult<()> {
+    // `&Arc<Self>` rather than `&self`: a `Disconnected` transition arms a countdown that
+    // outlives this call and needs to keep the session alive while it runs.
+    async fn on_rtc_event(self: &Arc<Self>, event: RtcEvent) -> EngineResult<()> {
         match event {
             RtcEvent::IceCandidate { ice_candidate, target } => {
                 log::debug!("local ice_candidate {:?} {:?}", ice_candidate, target);
@@ -1627,6 +1644,9 @@ impl SessionInner {
                         proto::leave_request::Action::Resume,
                         false,
                     );
+                } else if state == PeerConnectionState::Disconnected {
+                    // Not actionable yet — see `PC_DISCONNECTED_GRACE`.
+                    self.arm_disconnected_grace(target);
                 }
             }
             RtcEvent::DataChannel { data_channel, target } => {
@@ -1988,6 +2008,67 @@ impl SessionInner {
         }
 
         Ok(transceiver)
+    }
+
+    /// The transport for `target`, or `None` when this session has none (there is no
+    /// separate subscriber in single-PC mode).
+    fn transport_for(&self, target: SignalTarget) -> Option<&PeerTransport> {
+        match target {
+            SignalTarget::Publisher => Some(&self.publisher_pc),
+            SignalTarget::Subscriber => self.subscriber_pc.as_ref(),
+        }
+    }
+
+    /// Start the [`PC_DISCONNECTED_GRACE`] countdown for a transport that has just entered
+    /// `Disconnected`, and escalate if it has not recovered by the time it elapses.
+    ///
+    /// The decision is deliberately made from the transport's state *when the countdown
+    /// elapses* rather than from the transition that started it: a connection that recovered
+    /// on its own — the common case for a transient disturbance — needs no cancellation
+    /// bookkeeping, it simply reads as connected and the countdown lapses silently. A
+    /// connection that flapped and is down again at that point is escalated, which is the
+    /// right call regardless of how many times it bounced in between.
+    ///
+    /// Escalation is left to the engine, which already knows how to interpret a transport
+    /// failure in context: outside a reconnect it starts one, and during a reconnect it
+    /// sticks a full-reconnect escalation onto the cycle rather than looping on resume.
+    fn arm_disconnected_grace(self: &Arc<Self>, target: SignalTarget) {
+        let Some(transport) = self.transport_for(target) else {
+            return;
+        };
+        if !transport.try_arm_disconnect_grace() {
+            return; // a countdown is already running for this transport
+        }
+
+        let session = self.clone();
+        livekit_runtime::spawn(async move {
+            livekit_runtime::sleep(PC_DISCONNECTED_GRACE).await;
+
+            let Some(transport) = session.transport_for(target) else {
+                return;
+            };
+            transport.disarm_disconnect_grace();
+
+            if session.closed.load(Ordering::Acquire) {
+                return;
+            }
+
+            let state = transport.peer_connection().connection_state();
+            if should_escalate_after_grace(state) {
+                log::warn!(
+                    "{:?} pc stayed in {:?} for {:?}; treating as a failed transport",
+                    target,
+                    state,
+                    PC_DISCONNECTED_GRACE,
+                );
+                session.on_session_disconnected(
+                    "pc_state disconnected beyond grace period",
+                    DisconnectReason::UnknownReason,
+                    proto::leave_request::Action::Resume,
+                    false,
+                );
+            }
+        });
     }
 
     /// Called when the SignalClient or one of the PeerConnection has lost the connection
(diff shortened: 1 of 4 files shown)


livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -2674,6 +2755,26 @@ pub fn handle_remote_dt_packets(dc: &DataChannel, emitter: WeakUnboundedSender<S
...
+    match state {
+        // Recovered on its own — the transient disturbance this grace period exists to
+        // absorb. Nothing to do.
+        PeerConnectionState::Connected => false,
+        // Torn down deliberately (close, or a full reconnect replacing this session).
+        PeerConnectionState::Closed => false,
+        // Already reported the moment it happened, by the `Failed` branch in the connection
+        // change handler; escalating again here would double-report the same failure.
+        PeerConnectionState::Failed => false,
+        // Still `Disconnected` — or never got back past `Connecting` — after a window that a
```
lukasIO: The `ICE_CONNECTION_TIMEOUT` is currently 15s. 
As this check here also acts on `connecting` state now, we would basically never allow for PC reconnections to take longer than 5s. Is that desired behaviour?

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -2695,6 +2776,26 @@ pub fn handle_remote_dt_packets(dc: &DataChannel, emitter: WeakUnboundedSender<S
...
+        // healthy blip would have recovered inside. Treat it as a dead transport.
+        _ => true,
+    }
```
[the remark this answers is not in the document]
lukasIO: that's the same issue I refered to in [this comment](https://github.com/livekit/rust-sdks/pull/1332#discussion_r3796120098)

cnderrauber:
SFU’s timeout:
```
iceDisconnectedTimeout = 10 * time.Second               
iceFailedTimeout       = 5 * time.Second                         
```
Can we use the same value in rust-sdk to rely on the ice connection state transition?

livekit/src/rtc_engine/rtc_session.rs
xianshijing-lk: Good catch, and you're right that Connecting shouldn't have been in there. Fixed in the latest push, but I want to separate two things, because the 5s and the 15s turn out not to be measuring the same thing.

 ICE_CONNECT_TIMEOUT isn't a steady-state clock. It's used in exactly two places (rtc_session.rs:2526, :2712), both inside an explicit wait : the reconnect PC wait and the publisher-connection wait. Neither runs when nothing is reconnecting. So on main, a PC stuck in Connecting with no reconnect in flight has no clock running at all; nothing happens at 15s, and nothing happens until libwebrtc reaches Failed at consent expiry, tens of seconds later. That gap is what the countdown was added to close.

The grace countdown also only arms on a Disconnected transition, so it never applies to a PC that's merely Connecting,  only to one that already dropped and is now retrying.

 But your underlying point holds, and it's sharper than the framing. 5s is a defensible bar for Disconnected, where nothing is visibly being done. It's the wrong bar for Connecting, where an ICE restart is actually in flight: gathering, connectivity checks and DTLS can easily exceed 5s on a relayed or lossy path — a single lost STUN binding request costs ~1.5s of retransmit before the reply lands. Cutting that off is exactly backwards on the networks this feature exists to help.

 So Connecting now returns Rearm rather than escalating, buying one further window (PC_DISCONNECTED_GRACE_WINDOWS = 2). Bounded, not open-ended — a repair that hasn't landed in two windows still gets escalated.

 Digging into this turned up a worse problem in the same area, also fixed here. When the countdown fired during an in-flight reconnect, it reported a failure that the engine treats as proof the resume isn't holding, latching a full reconnect onto the cycle. But wait_pc_reconnected_with_snapshot was already judging the same transport with better evidence — the generation counters, i.e. transition history rather than a single level read — and on its own 15s deadline, escalating correctly on expiry. So the countdown wasn't adding a second opinion, it was adding a worse-informed one on a shorter fuse, and it converted merely-slow resumes into track-republishing rebuilds. Worse, it fired preferentially on slow networks, since a healthy resume finishes before the window elapses. It now stands down entirely when a reconnect is in flight.

xianshijing-lk:
> SFU’s timeout:
> 
> ```
> iceDisconnectedTimeout = 10 * time.Second               
> iceFailedTimeout       = 5 * time.Second                         
> ```
> 
> Can we use the same value in rust-sdk to rely on the ice connection state transition?

From Claude, it tells they are different things and we can't completely rely on the ice connection: 

Good question — worth spelling out, because the client already has equivalents of both knobs and they're closer than expected.

libwebrtc's ICE timeouts (p2p/base/p2p_constants.h):

- ice_unwritable_timeout = 5s (+ 5 failed checks) → selected pair leaves writable → ICE kDisconnected → PeerConnectionState::Disconnected
- ice_inactive_timeout = 15s → pair inactive → ICE kFailed → PeerConnectionState::Failed
- receiving_timeout = 2.5s (not surfaced as a PC state)

So the failed deadlines already coincide: 15s on the client, 10 + 5 = 15s on the SFU. The only divergence is disconnected — client ~5s vs SFU 10s — and the client is the faster of the two. Adopting 10s there would make the SDK slower to notice a dead path, which widens exactly the staleness window this PR is closing. (The SFU's 10s also carries a specific reason in transport.go — ice-lite/Firefox interop — that doesn't transfer to a client-side agent.)

Beyond the values, matching them still wouldn't let the resume read PeerConnectionState as recovery evidence, for two reasons:

1. A healthy resume produces no transition at all. Measured against a dev server (recorded in the doc comment on PC_RECONNECT_SETTLE_DELAY): a same-node resume performs a genuine subscriber ICE restart — the server's offer carries a fresh ice-ufrag and both transports gather a new generation — yet neither IceConnectionState nor PeerConnectionState ever leaves Connected, because libwebrtc keeps the old candidate pair selected until the new one is ready. Requiring an observed transition would time out every signal-only resume into a full reconnect. No timeout value changes that.
2. The Failed edge is unobservable inside this wait by construction. It lands at 15s, and ICE_CONNECT_TIMEOUT is also 15s — the wait dies at exactly the moment the evidence would arrive.

Worth noting we are relying on ICE state transitions, just as edges rather than as a level: note_connection_state counts every entry/exit of Connected in the event handler, so a poll running later still sees a drop that has since been repaired. The level can't be polled after the fact; that's the whole defect being fixed.

One place the sizing question does bite: the "never dropped, never re-entered" branch is fail-open at 3s, which is under the client's own ~5s detection latency, so a dead subscriber can still slip through if the resume completes within ~2s of the media break. Raising the delay would close it but charges every signal-only resume, so I'd rather fix it with positive evidence — the selected candidate pair's responses_received / bytes_received / selected_candidate_pair_changes from getStats, which proves the far end is answering in milliseconds when media is flowing and removes the need for the delay to outlast any ICE timeout. Planning that as a follow-up rather than growing this PR.

FWIW the knobs are settable if we ever want them — ice_unwritable_timeout / ice_inactive_timeout exist on RTCConfiguration — but they aren't plumbed through webrtc-sys today (our RtcConfiguration carries only ice_servers, continual_gathering_policy, ice_transport_type).

xianshijing-lk:
Ping for review

livekit-api/src/signal_client/mod.rs
```diff
@@ -385,6 +385,21 @@ impl SignalClient {
...
+    ///
+    /// Set by [`Self::restart`] before it touches the stream and cleared by
+    /// [`Self::set_reconnected`] once the resume has fully recovered, so it spans the
+    /// whole resume — not just the moment the socket is down. Callers use it to tell
+    /// "something is already driving recovery, with its own deadline" from "nothing is
+    /// watching this", which are the two cases that want opposite reactions to a
+    /// transport that has stopped reporting `Connected`.
+    ///
+    /// Unlike [`Self::is_connected`], this says nothing about the socket itself: during a
+    /// healthy resume it stays `true` well after the new stream is in place.
```
lukasIO, line 399: What's your opinion on the usefulness of these kind of comments? 
I personally find them rather distracting and potentially confusing given the fact that they might quickly be out of date wrt the actual code base.

PR #1336, opened 2026-08-17
Bump libwebrtc to webrtc-b9233c3-2

### LINKED ISSUE #1301: TURN/TLS fails with unknown_ca when root is missing from WebRTC bundle
Hi! I think TURN/TLS certificate verification in the FFI is using an incomplete embedded root bundle.

### Environment

- `livekit` Python 1.1.13 / `livekit-ffi` 0.12.68 on macOS
- Also reproduced with Python SDK 0.22.0, 1.0.0, and 1.0.25
- TURN/TLS on port 443 behind an AWS NLB, using an ACM certificate

### Minimal reproduction

```python
from livekit import rtc

config = rtc.RtcConfiguration(
    ice_transport_type=rtc.IceTransportType.TRANSPORT_RELAY,
)
room = rtc.Room()
await room.connect(url, token, rtc.RoomOptions(rtc_config=config))
```

The signaling connection and `JoinResponse` succeed, but the RTC connection times out with ICE error 701. A browser connects to the same room successfully.

A packet capture shows the server sending `ServerHello` and its certificate chain, followed by the client sending TLS fatal alert 48 (`unknown_ca`). No TURN allocation is attempted.

The chain ends at `Starfield Services Root Certificate Authority - G2`. WebRTC M144's embedded root bundle does not contain that root. I reproduced this directly with its BoringSSL build: verification fails with the embedded bundle, then succeeds after adding the missing root (SHA-256 `56:8D:69:05:A2:C8:87:08:A4:B3:02:51:90:ED:CF:ED:B1:97:4A:60:6A:13:C6:E5:29:0F:CB:2A:E6:3E:DA:B5`).

Could the embedded roots be updated, or could TURN/TLS use the OS trust verifier?

pblazej:
Fixes #1301 — picks up webrtc-sdk/webrtc#277, which falls back to the operating system's trust store when the anchors compiled into `rtc_base/ssl_roots.h` yield no path for a TURN/TLS chain.

### Why it was broken

WebRTC validates TURN/TLS solely against the anchors in `rtc_base/ssl_roots.h` — 36 roots generated in 2023 from `https://pki.goog/roots.pem`, Google's own PKI list rather than a general CA store. It contains no Amazon Root CA, no Starfield Services Root G2 and no ISRG Root X1, so a relay-only connection to a TURN server fronted by AWS ACM or Let's Encrypt fails with `unknown_ca` even though the host OS trusts the chain and the server sent a complete one.

### Why no Rust-side change is needed

The fix lives in `rtc_base/openssl_adapter.cc`, below every SDK, and `webrtc-sys` binds the C++ API directly — so it is covered without touching this repo beyond the tag. Verified locally on macOS by driving `rtc_base`'s `SSLAdapter` with no ObjC and no PeerConnection in the picture, against a local coturn whose CA is outside the built-in set:

```
(openssl_adapter.cc:924): Invoking SSL Verify Callback.
(platform_certificate_verifier_apple.cc): rejected by the system trust store: ...
```

### Build

[webrtc-b9233c3-2](https://github.com/livekit/rust-sdks/releases/tag/webrtc-b9233c3-2) — all 11 platform jobs green, 11 assets published. Built from `9e3b4f8`, so it includes the patch repairs from #1329; confirmed in the job logs that `external_audio_source.patch` applied cleanly rather than being skipped by `|| true`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -10,12 +10,16 @@ rustflags = ["-C", "target-feature=+crt-static"]
 
 [target.x86_64-apple-darwin]
 rustflags = ["-C", "link-args=-ObjC"]
+rustdocflags = ["-C", "link-args=-ObjC"]
 
 [target.aarch64-apple-darwin]
 rustflags = ["-C", "link-args=-ObjC"]
+rustdocflags = ["-C", "link-args=-ObjC"]
 
 [target.aarch64-apple-ios]
 rustflags = ["-C", "link-args=-ObjC"]
+rustdocflags = ["-C", "link-args=-ObjC"]
 
 [target.aarch64-apple-ios-sim]
 rustflags = ["-C", "link-args=-ObjC"]
+rustdocflags = ["-C", "link-args=-ObjC"]
--- a/.changeset/bump_libwebrtc_to_webrtc_b9233c3_2.md
+++ b/.changeset/bump_libwebrtc_to_webrtc_b9233c3_2.md
@@ -1,0 +1,19 @@
+---
+webrtc-sys: patch
+libwebrtc: patch
+livekit: patch
+livekit-ffi: patch
+---
+
+fix: bump libwebrtc to webrtc-b9233c3-2 so TURN/TLS can use the OS trust store
+
+WebRTC validates TURN/TLS against a small set of anchors compiled into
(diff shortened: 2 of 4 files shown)


cloudwebrtc (approved):
lgtm!

PR #1339, opened 2026-08-18
Expose remote data track pipeline options over UniFFI

pblazej:
Exposes `RemoteDataTrack.set_pipeline_options` and a `RemoteDataTrackPipelineOptions` record (`max_partial_frames`, default 1) so Swift reaches parity with the JS, Python, and Rust SDKs, which all document this as the lever for reassembling large or high-frequency frames under packet reordering. Also defaults `DataTrackSubscribeOptions.buffer_size` to 16 to match the documented default.

### DIFF
--- a/.changeset/expose_remote_data_track_pipeline_options.md
+++ b/.changeset/expose_remote_data_track_pipeline_options.md
@@ -1,0 +1,5 @@
+---
+livekit-uniffi: patch
+---
+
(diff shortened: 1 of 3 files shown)


pblazej:
@ladvoc I discarded the idea of `flush()` for now, it seems possible but will require integration with the consuming SDK.

ladvoc:
> @ladvoc I discarded the idea of `flush()` for now, it seems possible but will require integration with the consuming SDK.

Ye, flush was not applicable here because ordering is already guaranteed by the async channel.

ladvoc (approved):
LGTM ✅

Thanks for adding this, apologies for the oversight.

PR #1343, opened 2026-08-20
fix(uniffi): register the Bytes custom type once, in livekit-common

pblazej:
`bytes::Bytes` was registered with `custom_type!` in both `livekit-uniffi` and `livekit-datatrack`. A remote type can only be registered once per UniFFI component: each emits its own `public typealias Bytes` and `FfiConverterTypeBytes`, and cargo-swift compiles every component file into one Swift module — `invalid redeclaration of 'Bytes'`. UniFFI's own duplicated helpers coexist because they're `fileprivate`; `custom_type!` emits `public`, so they don't. Patched until now by a perl regex in `swift-workarounds`, which is deleted here.

`livekit-common` now owns the registration and every component borrows it:

```rust
uniffi::use_remote_type!(livekit_common::Bytes);
```

Owning a registration means being a component — `custom_type!` resolves `crate::UniFfiTag`, and bindgen rejects metadata belonging to no namespace (`Unknown namespace for CustomType`). So `livekit-common` gains an optional `uniffi` feature, enabled transitively by `livekit-uniffi`, and a third generated Swift file. **That file had to be added to `uniffi-swift.yml`** — the publish step enumerates sources rather than globbing, so an unlisted component silently ships a package that can't build. Costs 4,160 B on the cdylib (1,107,856 → 1,112,016), leaving 66 KiB under the iOS gate.

Verified: `livekit-common` and `livekit-datatrack` still build without the feature; the three-file Swift module compiles; `client-sdk-swift` `main` builds clean against the generated package, tests included; the xcframework carries all three headers in one framework modulemap.

## Alternatives for the collision

| Option | Verdict |
| --- | --- |
| Keep the perl regex | Fragile — matches literal generated formatting, so it silently no-ops when uniffi's Swift templates shift — and Swift-only, so Kotlin and Python kept shipping two declarations. |
| Let `livekit-datatrack` own it | Free (80 B, no new component), but an unintuitive home for a generic byte buffer, and any future component would have to depend on `livekit-datatrack` to borrow it. Was the first version of this PR. |
| Collapse to one component — drop `uniffi` from `livekit-datatrack` and mirror its types with `#[uniffi::remote]` | Removes the class outright and returns the duplication below. But ~150 lines of mirrored Rust, hand-written adapters for the two `with_foreign` traits, and dropping `#[non_exhaustive]` from two enums. Separate PR. |

Upstream: [#2933](https://github.com/mozilla/uniffi-rs/issues/2933) (open) is this bug, and this PR takes the external-type route it proposes. [#2126](https://github.com/mozilla/uniffi-rs/issues/2126) (open) — no `#[uniffi::remote(Trait)]` in 0.31 or 0.32 — is why collapsing needs trait adapters. [#2802](https://github.com/mozilla/uniffi-rs/issues/2802) (open) is the same conflict across separate UniFFI *libraries*.

## Alternatives for the duplication

Not addressed here. Each component re-emits the whole FFI runtime: ≈19 KB across three of them — 8.3 KB native and ~10.6 KB of Swift code+data, from two measured per-component deltas of ~4.2 KB and 5,287 B.

| Option | Verdict |
| --- | --- |
| Hoist the shared Swift runtime into one `internal` file | Worse — prototyped at **+14,592 B** on the linked binary, because it destroys the per-file specializations that make the duplicate copies cheap. Buys ~5% compile time. |
| Wait for upstream to emit the common part once | Asked for twice, declined twice. Not in flight. |
| Collapse to one component | The only thing that actually returns it. |

Upstream: [#408](https://github.com/mozilla/uniffi-rs/issues/408) / [#409](https://github.com/mozilla/uniffi-rs/pull/409) (closed) asked to eliminate re-declared Swift helpers and were resolved with the `UNIFFI_SHARED_H` guard plus `fileprivate` — duplicates left to coexist rather than removed. [#2257](https://github.com/mozilla/uniffi-rs/issues/2257) (closed) asked to combine the per-crate outputs; closed without a feature. [#2153](https://github.com/mozilla/uniffi-rs/issues/2153) (open) is the only live proposal that would share the runtime, and also documents the Python breakage: bindgen writes `from . import livekit_datatrack` into a flat out-dir with no `__init__.py`, giving `ImportError: attempted relative import with no known parent package`. [#2930](https://github.com/mozilla/uniffi-rs/issues/2930) (open) is the Swift → bindings-pipeline migration such a change would land on.

### DIFF
--- a/livekit-common/Cargo.toml
+++ b/livekit-common/Cargo.toml
@@ -9,10 +9,10 @@ repository.workspace = true
 
 [dependencies]
 livekit-protocol = { workspace = true }
-bytes = { workspace = true }
+bytes = { workspace = true, optional = true }
 uniffi = { workspace = true, features = ["scaffolding-ffi-buffer-fns"], optional = true }
 
 [features]
 # Exposes this crate's shared FFI type registrations. Enabled transitively by
 # livekit-uniffi so every component borrows one converter per type.
-uniffi = ["dep:uniffi"]
+uniffi = ["dep:uniffi", "dep:bytes"]
--- a/livekit-common/Cargo.toml
+++ b/livekit-common/Cargo.toml
@@ -9,3 +9,10 @@ repository.workspace = true
 
 [dependencies]
 livekit-protocol = { workspace = true }
+bytes = { workspace = true }
+uniffi = { workspace = true, features = ["scaffolding-ffi-buffer-fns"], optional = true }
+
+[features]
+# Exposes this crate's shared FFI type registrations. Enabled transitively by
+# livekit-uniffi so every component borrows one converter per type.
+uniffi = ["dep:uniffi"]
--- a/livekit-uniffi/src/common.rs
+++ b/livekit-uniffi/src/common.rs
@@ -14,10 +14,8 @@
 
 use bytes::Bytes;
 
-// `Bytes` is a remote type, so each UniFFI component that registers it with
-// `custom_type!` emits its own converter — and the two component Swift files are
-// compiled into one module, so a second `typealias Bytes` fails to build
-// ("invalid redeclaration of 'Bytes'"). Reuse livekit-datatrack's registration
-// instead of adding a second one; the type is then emitted once, in the file that
-// owns it. Upstream: https://github.com/mozilla/uniffi-rs/issues/2933
-uniffi::use_remote_type!(livekit_datatrack::Bytes);
+// `Bytes` is registered once in livekit-common so every component that needs it borrows the
+// same converter; registering it here as well would emit a second `public typealias Bytes`
+// into this component's Swift file, and both files compile into one module.
+// Upstream: https://github.com/mozilla/uniffi-rs/issues/2933
+uniffi::use_remote_type!(livekit_common::Bytes);
--- a/livekit-uniffi/src/common.rs
+++ b/livekit-uniffi/src/common.rs
@@ -14,4 +14,10 @@
 
 use bytes::Bytes;
 
-uniffi::custom_type!(Bytes, Vec<u8>, { remote });
+// `Bytes` is a remote type, so each UniFFI component that registers it with
+// `custom_type!` emits its own converter — and the two component Swift files are
+// compiled into one module, so a second `typealias Bytes` fails to build
+// ("invalid redeclaration of 'Bytes'"). Reuse livekit-datatrack's registration
+// instead of adding a second one; the type is then emitted once, in the file that
+// owns it. Upstream: https://github.com/mozilla/uniffi-rs/issues/2933
+uniffi::use_remote_type!(livekit_datatrack::Bytes);
--- a/.changeset/fix_uniffi_register_the_bytes_custom_type_once_in_livekit_co.md
+++ b/.changeset/fix_uniffi_register_the_bytes_custom_type_once_in_livekit_co.md
@@ -1,0 +1,8 @@
+---
+livekit: patch
+livekit-api: patch
+livekit-data-stream: patch
+livekit-ffi: patch
+---
+
+fix(uniffi): register the Bytes custom type once, in livekit-common - #1343 (@pblazej)
--- a/.changeset/reuse_datatrack_bytes_converter.md
+++ b/.changeset/reuse_datatrack_bytes_converter.md
@@ -1,0 +1,5 @@
+---
+livekit-uniffi: patch
+---
+
+Reuse `livekit-datatrack`'s `Bytes` converter instead of registering a second one, so UniFFI emits it once and the post-generation Swift workaround is no longer needed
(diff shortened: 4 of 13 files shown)


livekit-uniffi/src/common.rs
```diff
@@ -14,4 +14,10 @@
...
 use bytes::Bytes;
 
-uniffi::custom_type!(Bytes, Vec<u8>, { remote });
+// `Bytes` is a remote type, so each UniFFI component that registers it with
+// `custom_type!` emits its own converter — and the two component Swift files are
+// compiled into one module, so a second `typealias Bytes` fails to build
+// ("invalid redeclaration of 'Bytes'"). Reuse livekit-datatrack's registration
+// instead of adding a second one; the type is then emitted once, in the file that
+// owns it. Upstream: https://github.com/mozilla/uniffi-rs/issues/2933
+uniffi::use_remote_type!(livekit_datatrack::Bytes);
```
1egoman: My main concern with putting this in `livekit-datatrack` is that it's a somewhat unintuitive place for this definition to go. I'm also not sure how this would impact data streams v2 given the uniffi bindings will have the same problem (and a dependency between `livekit-datatrack` and `livekit-data-stream` sounds like a bad idea).

I realize it's a bit more work, but what do you think about doing something like defining this `custom_type!` in some third package which `livekit-uniffi`, `livekit-datatrack`, and [data streams v2 uniffi](https://github.com/livekit/rust-sdks/pull/1286) once that is is merged can all pull in via `uniffi::use_remote_type!(...)`? I'm not sure what this third package would be called, but my initial thought is maybe it could be in `livekit-common` - that package could gain a `uniffi` feature which gets set through each layer when building `livekit-uniffi`.

Thoughts on this?

ladvoc:
Wait, what happens if you simply do this:
```rust
uniffi::use_remote_type!(bytes::Bytes);
```
At least the docs for [this macro](https://mozilla.github.io/uniffi-rs/latest/types/remote_ext_types.html) imply this would work.

pblazej:
@ladvoc looks like it won't work:

```
error[E0425]: cannot find type `UniFfiTag` in crate `bytes`

The first path segment isn't the crate that defines the type — it's the crate whose registration you're borrowing. The macro expands to <Bytes as FfiConverter<bytes::UniFfiTag>>, and bytes isn't a UniFFI crate, so it has no UniFfiTag. The docs are genuinely misleading here: their example borrows from another uniffi crate, so the path reads like a plain type path. You need a crate that called setup_scaffolding!().
```

livekit-uniffi/src/common.rs
pblazej: Prototyped it — works, `Bytes` lands once in `livekit_common.swift`. But it has to be a full *component*: a hand-written `UniFfiTag` compiles, then bindgen refuses with `Unknown namespace for CustomType`. Cost is a third namespace — `livekit_common.swift` at 558 lines, 553 of them duplicated runtime, publishing one `typealias Bytes = Data` — plus +4,160 B on the cdylib.

#1286 doesn't hit this either: no new `setup_scaffolding!()`, it reuses this crate's registration. And it answers the general question the other way, hand-mirroring `common::EncryptionType`/`ClientCapability` into `livekit-uniffi`.

So it turns on whether `livekit-common` becomes a component — bigger than this PR, and a two-line follow-up either way.

pblazej: Done — pushed b46be142, `livekit-common` owns it now.

Verified: `livekit-common` and `livekit-datatrack` still build without the feature (the dep is optional, gated on `uniffi`), the three-file module compiles, `client-sdk-swift` `main` builds clean against the generated package with tests, and the xcframework carries all three headers in one modulemap. Cost is 4,160 B on the cdylib, 66 KiB under the iOS gate.

livekit-common/Cargo.toml
```diff
@@ -9,3 +9,10 @@ repository.workspace = true
 
 [dependencies]
 livekit-protocol = { workspace = true }
+bytes = { workspace = true }
+uniffi = { workspace = true, features = ["scaffolding-ffi-buffer-fns"], optional = true }
+
+[features]
+# Exposes this crate's shared FFI type registrations. Enabled transitively by
+# livekit-uniffi so every component borrows one converter per type.
+uniffi = ["dep:uniffi"]
```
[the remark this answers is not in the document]
pblazej: Good catch, applied in 4f3e5120 — `bytes = { workspace = true, optional = true }`, enabled by `uniffi = ["dep:uniffi", "dep:bytes"]`.

Worth noting the impact was nil in practice: `livekit`, `livekit-api`, `livekit-data-stream`, `livekit-datatrack` and `livekit-uniffi` all declare `bytes` directly already, so nothing gained a dependency — but the edge shouldn't have been there. `cargo tree -p livekit-common --depth 1` is back to just `livekit-protocol` without the feature, and lists `bytes` and `uniffi` with it.

1egoman (approved):
Nice, I'm a fan of the `livekit-common` approach!

PR #1344, opened 2026-08-21
Extract `livekit-token` from `livekit-api`

1egoman:
### Project Overview
Right now, `livekit-api` is a crate that is an amalgamation of many things:
- Signaling code
- Server focused apis which allow users to issue commands via twirp / etc
- Access token generation
- Webhook management / smaller server focused utilities / etc

We've discussed on the client team that it would be good to start to split this up. Larger, monolithic crates are harder to build and require larger numbers of features (which can interact in unexpected ways) than smaller ones, and ideally once this migration is complete, `livekit` will be able to depend on the smaller constituent crates rather than the monolithic `livekit-api`.

### Description
This initial pull request is focused on extracting out a new `livekit-token` crate from `livekit-api`. This new crate will contain all access token generation code and be depended on by `livekit-uniffi` (for exposing via uniffi) and `livekit-api` (so it can be re-exported under its old path for downstream consumption).

This should NOT be a breaking change and is largely scoped as a lift and shift effort right now. The expectation is that from this point forward, all livekit packages will consume `livekit-token` directly, and any external packages will be expected to consume `livekit-api`.

## TODO
- [x] Publish initial `livekit-token` to crates.io


livekit-api/src/lib.rs
```diff
@@ -14,11 +14,12 @@
...
 
+// The token implementation lives in the livekit-token crate. This alias keeps
+// the historical `livekit_api::access_token::*` paths working, and remains the
+// documented, supported way to reach these types.
 #[cfg(feature = "access-token")]
-pub mod access_token;
-
-#[cfg(feature = "access-token")]
-mod jwt_provider;
+#[doc(inline)]
```
ladvoc, line 21: praise: Nice, did not know about this attribute.

ladvoc (approved):
LGTM, a few optional follow-ups that might be good to include here but feel free to ignore:
- Add an initial change log entry for `livekit-token` documenting that it used to live in `livekit-api`
- Update examples that depend on `livekit-api` just for token generation to depend on `livekit-token` instead

ladvoc:
Note: since this PR adds a new crate, it will need to be published manually the first time (due to trusted publishing). This should be done before the release since, in order to publish a new version of `livekit-api`, it will need to already exist on crates.io.

1egoman:
@ladvoc In regards to:

> Update examples that depend on livekit-api just for token generation to depend on livekit-token instead

I was thinking that these crates would be an internal abstraction only - ie, examples would still keep using `livekit-api`, docs would still recommend this pattern, etc. The only difference would be organization and the resulting build time / feature flag simplification for internal uses.

If you are thinking otherwise, definitely interested in hearing more, but I think this is a fairly important point to be agreement on since it possibly influences further refactoring efforts. For now I'll keep moving under the assumption this is an internal only refactor.

1egoman:
Created a new `livekit-token` stub release on crates.io: https://crates.io/crates/livekit-token

PR #1345, opened 2026-08-21
Extract `livekit-signaling` (and an unexpected side effect, `livekit-region`) from `livekit-api`

1egoman:
### Project Overview
Right now, `livekit-api` is a crate that is an amalgamation of many things:
- Signaling code
- Server focused apis which allow users to issue commands via twirp / etc
- Access token generation
- Webhook management / smaller server focused utilities / etc

We've discussed on the client team that it would be good to start to split this up. Larger, monolithic crates are harder to build and require larger numbers of features (which can interact in unexpected ways) than smaller ones, and ideally once this migration is complete, `livekit` will be able to depend on a handful of relevant smaller constituent crates rather than the monolithic `livekit-api`.

### Description
This pull request builds on #1344 and extracts out a new `livekit-signaling` crate from `livekit-api`, which contains all signaling code. Arguably this being in `livekit-api` in the first place (given `livekit-api` is meant to be "serverside apis") was a little weird, though that was where the http infrastructure lived so I guess it made a certain amount of sense.

Now though, all the client focused http infrastructure has been moved to `livekit-net` in #1258, so there's very little forcing them to be in the same crate.

One small hiccup I did encounter: both existing serverside apis and signaling code both depend on [region selection code](https://github.com/livekit/rust-sdks/blob/main/livekit-api/src/region.rs), so in order to break out signaling, this needs to get moved to a "third thing" - I briefly considered `livekit-common`, but as `livekit-api` and `livekit-signaling` are likely to be the only dependents, I made a new `livekit-region` crate which I relocated this region selection code into.

This pull request will likely conflict with the signaling refactoring / state machine work lukas is working on. The actual code changes here are a "lift and shift", so lukas's updates should be able to be rebased on top of this into place without a ton of effort.

## TODO
- [x] Publish initial `livekit-signaling` and `livekit-region` crates to crates.io


.github/workflows/test-api.yml
```diff
@@ -40,10 +44,10 @@ jobs:
...
-          - name: signal-client (tokio)
-            cmd: cargo test -p livekit-api --no-default-features --features signal-client-tokio --lib signal_client -- --nocapture
-          - name: signal-client (async)
-            cmd: cargo test -p livekit-api --no-default-features --features signal-client-async --lib signal_client -- --nocapture
+          - name: livekit-signaling (tokio)
+            cmd: cargo test -p livekit-signaling --no-default-features --features tokio -- --nocapture
+          - name: livekit-signaling (async)
+            cmd: cargo test -p livekit-signaling --no-default-features --features async -- --nocapture
     services:
       mock-server:
```
1egoman: I'm not convinced this is a good idea - a LLM did this. Should there instead be a `livekit-signaling` specific ci job instead? Or maybe there's an argument I should keep things as they are and let lukas take over refining this as part of the signaling work more generally.

.github/workflows/test-api.yml
```diff
@@ -27,10 +27,13 @@ on:
 jobs:
   # Exercise every runtime backend so a regression in one (e.g. the async/isahc
   # server API silently failing to compile) is caught. Each leg pins a single
-  # runtime via --no-default-features, and every leg reaches the mock server: the
-  # services-* legs over HTTP, the signal-client legs over the WebSocket, selecting
-  # server behaviour with the token's `lk.mock` attribute. So the service container is
-  # a prerequisite for all of them, not a convenience for some.
+  # runtime via --no-default-features; the mock server backs the legs that make
+  # real requests (services-*), and is a harmless no-op for the livekit-signaling
+  # legs, which spin up their own ephemeral listeners.
```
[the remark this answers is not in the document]
1egoman: Fixed in 062230d

ladvoc (approved):
LGTM ✅, same suggestion as in #1344 to add change log entries to the new crates in order to document the history a bit.

lukasIO (approved):

.github/workflows/test-api.yml
```diff
@@ -41,10 +53,10 @@ jobs:
...
             cmd: cargo test -p livekit-api --no-default-features --features services-async,access-token --test services_async -- --nocapture
-          - name: signal-client (tokio)
-            cmd: cargo test -p livekit-api --no-default-features --features signal-client-tokio,access-token --lib --test signal_tokio --test signal_unreachable -- --nocapture
-          - name: signal-client (async)
-            cmd: cargo test -p livekit-api --no-default-features --features signal-client-async,access-token --lib --test signal_async -- --nocapture
+          - name: livekit-signaling (tokio)
+            cmd: cargo test -p livekit-signaling --no-default-features --features native-tokio --lib --test signal_tokio --test signal_unreachable -- --nocapture
+          - name: livekit-signaling (async)
+            cmd: cargo test -p livekit-signaling --no-default-features --features native-async --lib --test signal_async -- --nocapture
     services:
```
1egoman, line 60: I'm still not fully pleased with how these tests are organized now. I've tried a few different approaches and don't really love anything I've come to (moving these to `tests.yml`, breaking out into a seperate job in this file, etc) so I think I'm going to make a follow up task to clean this up and think about it a bit more.

1egoman:
Note: I have created `livekit-signaling` and `livekit-region` on crates.io.

PR #1347, opened 2026-08-22
yuv-sys: actionable error when libyuv submodule is not initialized

zacha7214:
## Problem

A fresh `git clone` without `--recursive` leaves `yuv-sys/libyuv` empty. The build script copies the empty directory and fails several calls later with:

thread 'main' panicked at yuv-sys/build.rs:140:10:
called Result::unwrap() on an Err value: Os { code: 2, kind: NotFound, message: "No such file or directory" }

Nothing in the error points at the submodule as the cause, and the README build instructions don't mention `--recursive`. CI/container builds don't hit this because the Dockerfile initializes submodules explicitly — it only affects local builds from a fresh clone.

## Fix

Check for `libyuv/include/libyuv.h` before the copy and panic with the remedy.

## Verification

- `git submodule deinit -f yuv-sys/libyuv` → build fails with the new message
- `git submodule update --init` → clean build succeeds (macOS, aarch64)

### DIFF
--- a/.changeset/yuv_sys_submodule_guard_error.md
+++ b/.changeset/yuv_sys_submodule_guard_error.md
@@ -1,0 +1,7 @@
+---
+yuv-sys: patch
+livekit-ffi: patch
+imgproc: patch
+---
+
+yuv-sys: fail with an actionable error when the libyuv submodule is not initialized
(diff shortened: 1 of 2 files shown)


ladvoc (approved):
LGTM ✅, thank you for the contribution.

PR #1351, opened 2026-08-24
Capture: clock source

ladvoc:
Add a capture source that renders a wall clock using the GPU.

This is useful for video latency measurements.

Closes: BOT-549

### DIFF
--- a/livekit-capture/src/sources/clock.rs
+++ b/livekit-capture/src/sources/clock.rs
@@ -1,0 +1,272 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


livekit-capture/src/sources/clock.rs
```diff
@@ -0,0 +1,272 @@
...
+        minute % 10,
+        COLON,
+        (second / 10) % 10,
+        second % 10,
+        DOT,
+        (millisecond / 100) % 10,
+        (millisecond / 10) % 10,
+        millisecond % 10,
+    ]
+}
```
alan-george-lk, line 193: This is some fun Rust syntax

stephen-derosa (commented):
nice, makes sense -- adding to the other shaders

alan-george-lk (approved):

PR #1354, opened 2026-08-24
Capture: device source

ladvoc:
Add a capture source for camera devices.

This PR defines the API for device source (which is the same across platforms); backend implementations (e.g., AVFoundation for macOS, V4L2 for Linux, etc.) are left to subsequent PRs).

Closes BOT-528


livekit-capture/src/sources/device/mod.rs
```diff
@@ -0,0 +1,601 @@
...
+#[non_exhaustive]
+pub enum DeviceFrameFormat {
+    /// Planar I420/YUV420P.
+    I420,
+    /// Biplanar NV12.
+    Nv12,
+    /// Packed BGRA.
+    Bgra,
+    /// Packed RGB24.
+    Rgb24,
```
alan-george-lk, line 79: Had this feedback with @stephen-derosa on the C++ side, any reason to not make these all-caps matching the comment/acronym?

livekit-capture/src/sources/device/mod.rs
```diff
@@ -0,0 +1,601 @@
...
+
+impl DeviceFrameFormat {
+    /// Returns a stable lower-case frame-format name.
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::I420 => "i420",
+            Self::Nv12 => "nv12",
+            Self::Bgra => "bgra",
+            Self::Rgb24 => "rgb24",
+            Self::Bgr24 => "bgr24",
```
alan-george-lk, line 100: Do these need to be lowercase?

livekit-capture/src/sources/device/mod.rs
```diff
@@ -0,0 +1,601 @@
...
+    livekit::webrtc::video_frame::FrameMetadata {
+        user_timestamp: Some(capture_wall_time_us),
+        frame_id: None,
+        user_data: None,
+    }
+}
+
+/// Validates the platform-neutral parts of a configuration; `supported`
+/// reports whether the backend can deliver a frame format.
+#[allow(dead_code)]
```
alan-george-lk, line 412: Similar to other PRs, why allow dead code on new stuff?

livekit-capture/src/sources/device/mod.rs
```diff
@@ -0,0 +1,601 @@
...
+//! ([`devices`]), and errors use one platform-neutral vocabulary. On
+//! platforms without a backend the module still compiles, and construction
+//! and enumeration fail with
+//! [`DeviceVideoSourceError::UnsupportedPlatform`].
+//!
+//! Where the platform supports it, frames reach the RTC track as
+//! platform-native buffers without a CPU copy. Otherwise they are converted
+//! to I420.
+
+use unsupported as backend;
```
alan-george-lk, line 28: I wasn't sure on this so I asked GPT, so take with a grain of salt:

> Device capture is not implemented on any platform. [mod.rs:28](https://github.com/livekit/rust-sdks/blob/5beef1332599643c44d8a92e8a9a18fe7ac36b21/livekit-capture/src/sources/device/mod.rs#L28) unconditionally aliases the backend to unsupported, and [source-device = []](https://github.com/livekit/rust-sdks/blob/5beef1332599643c44d8a92e8a9a18fe7ac36b21/livekit-capture/Cargo.toml#L55) adds no platform backend or dependencies. Consequently, devices(), devices_blocking(), DeviceVideoSource::new(), and the new FFI requests will always return UnsupportedPlatform, including on macOS/Linux/Windows.
`use unsupported as backend;` is not a runtime fallback decision—it selects that module at compile time. The selected module’s Session::open() always returns an error. So this PR exposes a polished public API, but no code path can ever open a camera.

Is there a way to make a unit test or some sort of test that exercises this? Is this actually an issue for this PR?

alan-george-lk (approved):
Approved with some minor stuff

PR #1355, opened 2026-08-24
Capture: AVFoundation backend for device source

ladvoc:
Add an AVFoundation backend for device source.

Closes BOT-412


livekit-capture/src/sources/device/avfoundation.rs
```diff
@@ -0,0 +1,1542 @@
...
+    output: &AVCaptureVideoDataOutput,
+) -> Option<Retained<NSDictionary<NSString, AnyObject>>> {
+    let preferred = [
+        // WebRTC's VideoToolbox H.264 encoder allocates full-range NV12
+        // buffers for its CPU upload path. Prefer the same CoreVideo
+        // format for direct CVPixelBuffer input so the native path does
+        // not have to reset VideoToolbox into a separate video-range pool.
+        kCVPixelFormatType_420YpCbCr8BiPlanarFullRange,
+        kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange,
+    ];
```
alan-george-lk, line 383: Not confident in this, but does this NV12-only preferred list limit the ability to use non-NV12? i.e. not using `frame_format` here to match requested frame format

alan-george-lk (approved):

PR #1356, opened 2026-08-24
Capture: V4L2 backend for device source

ladvoc:
Add a V4L2 backend for device source.

Closes BOT-413


livekit-capture/src/sources/device/v4l2.rs
```diff
@@ -0,0 +1,1106 @@
...
+fn backend_error(error: io::Error) -> DeviceVideoSourceError {
+    DeviceVideoSourceError::Backend(error.to_string())
+}
+
+/// Returns the ordered source frame formats to try for a request.
+///
+/// The request's own frame format (already validated as supported) is tried
+/// first; an explicit constraint on the highest-* requests pins the list to
+/// that one format.
+fn frame_formats_for_request(request: &DeviceFormatRequest) -> Vec<DeviceFrameFormat> {
```
alan-george-lk, line 278: I'm a little out of my depths here Rust-wise and V4L2-wise, so I thought it might be useful to have another model check for errors while I learn more about the Rust code and V4L2. Take this with a grain of salt, and apologies in advance if low-usefulness:

> P1 — Exact requests can silently fall back to a different pixel format.
[frame_formats_for_request](https://github.com/livekit/rust-sdks/blob/70d36276ad4014faf4d5f17016aeb735f85c4838/livekit-capture/src/sources/device/v4l2.rs#L278) puts the requested format first but appends defaults. Then [apply_ordered_format_request](https://github.com/livekit/rust-sdks/blob/70d36276ad4014faf4d5f17016aeb735f85c4838/livekit-capture/src/sources/device/v4l2.rs#L327) tries each one until one succeeds.
Example: Exact(1280×720, 30, Mjpeg) on a camera that cannot provide MJPEG but can provide YUYV will succeed as YUYV. Exact should instead return UnsupportedFormat.
Suggested comment:
P1: Exact requests are currently allowed to fall back to the default FourCC list. If the requested format is unavailable but a later fallback succeeds, construction returns a source in a different format instead of UnsupportedFormat. Could Exact pass only its requested frame_format to apply_ordered_format_request (or skip that fallback path entirely)? Please add a regression test where the requested FourCC is unavailable but a fallback FourCC is available. Rust angle: the code is type-safe, but its control flow changes a value the caller marked as “exact.” The compiler cannot infer that fallback is semantically forbidden.

> P2 — stepwise V4L2 sizes are reported as a complete list even though only the endpoints are retained.
[resolutions_from_frame_size](https://github.com/livekit/rust-sdks/blob/70d36276ad4014faf4d5f17016aeb735f85c4838/livekit-capture/src/sources/device/v4l2.rs#L637) converts a Stepwise range into only its minimum and maximum dimensions. Yet [devices()](https://github.com/livekit/rust-sdks/blob/70d36276ad4014faf4d5f17016aeb735f85c4838/livekit-capture/src/sources/device/v4l2.rs#L232) marks that list formats_complete: true.
Stepwise means the driver supports a grid/range of intermediate sizes. A client asking for the “closest” resolution may be given an endpoint even when a much better intermediate size is supported; device enumeration also claims those valid sizes do not exist.
Suggested comment:
P2: FrameSizeEnum::Stepwise is reduced to min/max while formats_complete is set to true. This makes Closest choose poor endpoint-only results and misreports the device’s formats as complete. Could we retain enough stepwise metadata to negotiate the nearest aligned size, or set formats_complete to false whenever any format/interval is stepwise?

alan-george-lk (approved):
LGTM assuming AI flagged issues are fixed

PR #1357, opened 2026-08-24
Capture: binding crate for libargus

ladvoc:
Add a `libargus-sys` crate with bindings to the NVIDIA Jetson libargus camera API.

Closes BOT-553


libargus-sys/src/lk_argus.h
```diff
@@ -0,0 +1,213 @@
...
+#define LK_ARGUS_ABI_VERSION 1
+
+/* Maximum supported DMA buffer ring size. */
+#define LK_ARGUS_MAX_DMA_BUFS 16
+
+/* Status codes. 0 is success; negative values are errors. */
+enum {
+  LK_ARGUS_OK = 0,
+  /* Invalid arguments crossed the FFI boundary (caller bug). */
+  LK_ARGUS_ERR_INVALID_ARG = -1,
```
alan-george-lk, line 52: It makes sense for error-codes, but typically I don't see negative values used for enums like this (I'm not a C developer so maybe it's common there)

libargus-sys/src/lk_argus.h
```diff
@@ -0,0 +1,213 @@
...
+#endif
+
+/* ABI version of this shim. Incremented on breaking changes. */
+#define LK_ARGUS_ABI_VERSION 1
+
+/* Maximum supported DMA buffer ring size. */
+#define LK_ARGUS_MAX_DMA_BUFS 16
+
+/* Status codes. 0 is success; negative values are errors. */
+enum {
```
alan-george-lk, line 49: `enum class` is a little more modern. This code looks like C though which doesn't have that, but the `.cpp` has lots of C++ in it, so a bit puzzled

libargus-sys/src/lk_argus.cpp
```diff
@@ -0,0 +1,939 @@
...
+// Logging
+
+std::mutex g_log_mutex;
+LkArgusLogFn g_log_fn = nullptr;
+void* g_log_user_data = nullptr;
+
+#if defined(__GNUC__)
+__attribute__((format(printf, 2, 3)))
+#endif
+void lk_log(int32_t level, const char* fmt, ...) {
```
alan-george-lk, line 54: Do we actually need another logger vs. just a `std::cout/cerr` or `fprintf`? This one is simple enough, just clarifying

libargus-sys/src/lk_argus.cpp
```diff
@@ -0,0 +1,939 @@
...
+#include "lk_argus.h"
+
+#include <atomic>
+#include <chrono>
+#include <cstdarg>
+#include <cstdio>
+#include <cstring>
+#include <mutex>
+#include <thread>
+#include <vector>
```
alan-george-lk, line 27: Says `C shim` but lots of C++ in here

libargus-sys/src/lk_argus.cpp
```diff
@@ -0,0 +1,939 @@
...
+
+  ret = NvBufSurfaceUnMap(surface, 0, -1);
+  if (ret != 0) {
+    lk_log(LK_ARGUS_LOG_ERROR, "NvBufSurfaceUnMap failed: %d", ret);
+    return LK_ARGUS_ERR_NVBUF;
+  }
+  return LK_ARGUS_OK;
+}
+
+void lk_argus_session_destroy(LkArgusSession* s) {
```
alan-george-lk, line 897: From AI-pairing, take with grain of salt:

> P1: lk_argus_session_destroy can delete s while another thread is blocked in lk_argus_frame_acquire; when acquireFrame returns, that function continues dereferencing s. This contradicts the documented “interrupts any pending acquire” shutdown behavior and is a use-after-free. Please add an in-flight-operation count plus condition variable (destroy sets interrupted, stops capture, then waits for acquire/copy operations to exit before freeing surfaces and s), or change the API contract to require callers to interrupt and join the acquisition thread before destruction.

alan-george-lk (approved):

PR #1374, opened 2026-08-28
Switch node support from uniffi-bindgen-node to uniffi-bindgen-react-native.

jhugman:
This is largely derived from the AIC work, which publishes to npmjs.


.changeset/switch_node_bindings_to_ubrn.md
```diff
@@ -0,0 +1,14 @@
+---
+livekit-uniffi: patch
+---
+
+# Build the node package with uniffi-bindgen-react-native
+
+The node bindings now come from ubrn's NAPI backend and ship in the layout the
+node ecosystem already uses: `@livekit/uniffi` carries the TypeScript and
+declares one optional dependency per platform, and `@livekit/uniffi-<triple>`
+carries just the native library. The right one is resolved at load time, so
```
1egoman, line 10: question: I am curious why exactly you opted to publish this as a separate set of packages. Today [@livekit/rtc-ffi-bindings](https://www.npmjs.com/package/@livekit/rtc-ffi-bindings) and the corresponding     `@livekit/rtc-ffi-bindings-linux-arm64-gnu` / `@livekit/rtc-ffi-bindings-darwin-arm64` / etc packages are already being published, each containing the dylib/so/dll which is included in each of these packages.

[Here's an example](https://www.npmjs.com/package/@livekit/rtc-ffi-bindings-linux-arm64-gnu?activeTab=code) of what the existing platform specific packages look like:

Image

IMO it would be good to combine these - ideally keep the `@livekit/rtc-ffi-bindings-` prefix on the package names, and name the library files the same as the existing files, but also include the node bindings alongside, so there is only one set of library containing packages which can be used for the legacy c abi fii and for uniffi.

.github/workflows/uniffi-cdylib.yml
```diff
@@ -92,21 +92,20 @@ jobs:
...
-            sudo apt-get install -y gcc-aarch64-linux-gnu
-            export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
-            # .cargo/config.toml forces -fuse-ld=lld for this target to link
-            # libwebrtc; livekit-uniffi doesn't use libwebrtc. A non-empty global
-            # RUSTFLAGS overrides that per-target config, selecting the GNU bfd
-            # linker the cross binutils provides.
-            export RUSTFLAGS="-C link-arg=-fuse-ld=bfd"
-          fi
-          cargo build --release --target ${{ matrix.target }} -p livekit-uniffi
+        run: cargo zigbuild --release --target ${{ matrix.target }}.2.28 -p livekit-uniffi
```
1egoman, line 108: question: Can you explain what `cargo-zigbuild` is and why it is needed here? I looked it up online and it seems to be related to cross compilation? IMO if what was here before worked I think it would be better to keep it as is.

livekit-uniffi/support/node/test/index.ts
```diff
@@ -54,8 +55,28 @@ async function main() {
   const decodedGrants = tokenVerify(jwt, credentials);
   console.log("Verified generated JWT:", decodedGrants);
```
1egoman, line 56: Ah, this is a fairly important point that I'm not sure we had the same shared understanding on - the old node implementation was in `livekit-uniffi`, not `livekit-ffi`. But for node to interact with the existing ffi interface over `livekit-ffi` the uniffi implementation needs to be in `livekit-ffi` package instead, and the old `livekit-uniffi` node implementation (what you ported on top of here) can be dropped.

I think seeing all the work you have done here, it's making me think the easiest path to get from the current state to the desired end state (ie, node bindings in existing `livekit-ffi` built node packages) might be to couple this PR with the imminent `livekit-ffi` / `livekit-uniffi` merge.

What comes to mind is something like:
1. Let this sit for a few weeks (IMO probably has to happen anyway so @lukasIO can review once he is back from vacation, he's the subject matter expert on node sdk builds in this repo) while the `livekit-ffi` / `livekit-uniffi` merge takes place
2. Once that is done rebase it on top of the merged `livekit-ffi`
3. While doing this rebase, move over the changes made here to `livekit-ffi` and merge with the existing packages (ie, publish these new packages under the old names, with the old dylib names, etc so they are fully backwards compatible)

Thoughts on this?

1egoman (commented):
I think as I started to review this I realized that what you had done here (add uniffi bindings to `livekit-uniffi`) and what I had in mind (add uniffi bindings to livekit-ffi) were sufficiently far apart where I think it makes sense to put this on pause for a bit until after the upcoming livekit-ffi / livekit-uniffi merge, and then as part of rebasing this PR onto the updated main, address some of the comments I left at that point.

Also that dovetails nicely into when @lukasIO will be back from vacation and I would really like this to be reviewed by him before being merged as well.

livekit-uniffi/support/node/test/index.ts
lukasIO: unless I'm missing something this ^ sounds like the right approach to me, too!

jhugman:
Yeah, I figured there were two migrations here: one migrating `livekit-uniffi` from `uniffii-bindgen-node` to ubrn (this PR), and the other migrating all the bindgens from `livekit-uniffi` to `livekit-ffi`.

Your suggestion is doing this in the opposite order: build out the node bindings for `livekit-ffi`, then delete the node bindings for `livekit-uniffi`. I think let's park this PR, and either re-do it _de novo_ in `livekit-ffi` or rebase after the move to `livekit-ffi`.

jhugman:
I've tidied this PR (removed the cross compiling) in readiness for the other migration landing.

PR #1375, opened 2026-08-28
Remove livekit-runtime

1egoman:
## Background
The `livekit-runtime` crate was [introduced in ~2024 by the folks at zed](https://github.com/livekit/rust-sdks/pull/310) so they could use `livekit` in zed, which wasn't using a tokio runtime (my understanding is, a completely custom runtime!). The general design patterns haven't really changed since then.

A few pain points that have been encountered recently touch on `livekit-runtime`:
1. The migration wasn't done completely, and there are quite a few tokio runtime specific calls which were missed (or introduced later inadvertently). One example I know of offhand is [the `tokio::fs` usage in the `send_file` data stream method](https://github.com/livekit/rust-sdks/blob/6078e6aed5df72c5e6613e76729f096964fbe894/livekit-data-stream/src/outgoing/manager.rs#L290). These could be fixed piecemeal, but we'd need to extend the abstractions further to deal with a bunch of different async operations which is fairly burdensome.
2. Because of 1, today, the `async` and `dispatcher` features (what zed was using specifically) don't actually build on `main` properly 😞 .
3. We should be testing to ensure that all feature combinations build in ci, but in order to do this with the current design of `livekit-runtime` (which isn't super well encapsulated - higher level features in the `livekit` crate sometimes imply certain runtimes), it would be challenging. We'd have to effectively do a build for each `platform` x `runtime` x `ssl configuration` x any ad hoc features which is impractical.
4. `async-std` (which the `async` feature enables) has recently been deprecated in favor of `smol`, which we don't support today.

From what I can tell looking through zed's code, they seem to have since migrated from the custom runtime approach they were using to a [more typical looking tokio backed approach](https://github.com/zed-industries/zed/blob/1e2e422c7b33b28eab0a948e442334198f3b6d27/Cargo.toml#L674). It also looks like they [might be using](https://github.com/zed-industries/zed/blob/1e2e422c7b33b28eab0a948e442334198f3b6d27/Cargo.toml#L979) a [rust-sdks fork, not actually mainline](https://github.com/zed-industries/livekit-rust-sdks). I've done a search across all of github, and the only places that `dispatcher` seems to be used is by stale zed forks, so I'm fairly confident this can be dropped with little to no consequence.

## Summary

Given this context and socializing this decision with the client team, we have decided to **drop** support for non tokio async runtimes. `livekit-runtime` will be deprecated and all non tokio shim code will be removed. This will address all points above (fewer features making testing easier, no need to support `async-std`, and ci validates the crate will always compile on tokio) and lead to a much simpler to maintain project.

If we decide eventually that it's worth reintroducing this behavior, then https://github.com/livekit/rust-sdks/pull/1369 would be a good place to start, along with taking some learnings from https://webrtc.rs/blog/2026/01/31/async-friendly-webrtc-architecture.html.

## TODO
- [ ] Publish a final `livekit-runtime` release marking it as deprecated.


livekit-api/Cargo.toml
```diff
@@ -9,21 +9,17 @@ readme = "README.md"
...
-signal-client-async      = ["signal-client", "livekit-signaling/native-async"]
-signal-client-dispatcher = ["signal-client", "livekit-signaling/native-dispatcher"]
+# deprecated.
+signal-client        = ["dep:livekit-signaling"]
+signal-client-native = ["signal-client", "livekit-signaling/native"]
+
+services = ["access-token", "dep:livekit-region", "dep:reqwest", "dep:tokio", "tokio/time"]
 
-services-tokio = ["access-token", "dep:livekit-region", "dep:reqwest", "dep:tokio", "tokio/time", "dep:livekit-runtime", "livekit-runtime/tokio"]
-services-async = ["access-token", "dep:livekit-region", "dep:isahc", "dep:livekit-runtime", "livekit-runtime/async"]
```
1egoman, line 26: Note to reviewers - there is a small breaking change here in a public crate. `signal-client-tokio` / `signal-client-async` / `signal-client-dispatcher` are no longer exposed, but these were never advertised publicly and acted as features that `livekit` used to bring in a subset of `livekit-api` prior to breaking out relevant bits of functionality into `livekit-signaling`.

livekit/Cargo.toml
```diff
@@ -9,11 +9,12 @@ readme = "README.md"
 
 [features]
 # By default ws TLS is not enabled
-default = ["tokio"]
+default = ["native"]
 
-async = ["livekit-signaling/native-async", "livekit-runtime/async"]
-tokio = ["livekit-signaling/native-tokio", "livekit-runtime/tokio"]
-dispatcher = ["livekit-signaling/native-dispatcher", "livekit-runtime/dispatcher"]
```
1egoman, line 16: Note to reviewers - another small feature level change in `livekit`. `async` / `dispatcher` no longer exist and `tokio` was kept around as an alias for `native` as one probably implies the other in hte vast majority of uses (the big exception would be `livekit-ffi` / `livekit-uniffi`)

.changeset/remove_livekit_runtime.md
```diff
@@ -0,0 +1,13 @@
+---
+libwebrtc: patch
+livekit: patch
+livekit-api: patch
```
xianshijing-lk: should it be a minor ? assuming that is an api change

xianshijing-lk (commented):
lgtm

.changeset/remove_livekit_runtime.md
1egoman: I think this makes sense, yea.

theomonnom (approved):

ladvoc (approved):
Generally looks good to me, been a while since I've reviewed a PR that isn't an addition! A few minor questions/suggestions:
- I would consider this a minor breaking change since crate feature are being removed/renamed.
- Remove "[...] however we plan to make the async executor runtime agnostic" from the top-level README.

.changeset/remove_livekit_runtime.md
1egoman: Done in 780b828

1egoman:
(I will wait until this is released and add a TODO item later to `Publish a final livekit-runtime release marking it as deprecated`)

PR #1376, opened 2026-08-28
Break out `livekit-rpc` into its own crate

1egoman:
## Summary
In the [original rpc v2 change](https://github.com/livekit/rust-sdks/pull/1013), I kept all of the newly introduced rpc logic in the `livekit` crate. At the time this made sense because `livekit` was the only consumer. However, we're rapidly moving towards a world where much of the client logic in each sdk will be shared rust code, and as part of that, rpc will need to be exposed via `livekit-uniffi`.

So, in preparation for this, relocate all rpc code out of `livekit` and into a new `livekit-rpc` crate. This new crate matches quite closely with how `livekit-datatrack` and `livekit-data-stream` are organized - there's an `api` namespace with all of the symbols to re-export publicly via `livekit`, and a `backend` namespace with all symbols which are required to wire up the `RpcClientManager` and `RpcServerManager` into a client.

To keep this small / isolated, I've opted to save the actual uniffi integration and wiring up into `livekit-uniffi` for a follow up pull request. It should be pretty straightforward though, largely just add `livekit-rpc` as a dependency to `livekit-uniffi`, then expose everything in `api` and `backend` via the ffi for other sdks to consume. It's not super high priority though since all sdks already support rpc v2 via their own downstream language implementations.

> [!WARNING]
> This pull request is probably most easily reviewed going commit by commit. I have structured it so that each commit is fairly atomic and easy to understand, but because of the file moves the diff of all commits merged together is fairly complex to reason about.

#### Drive by bug fixes

As part of this change, a LLM found a few bugs in the existing rpc v2 code, which I've opted to fix at the same time (these are each in their own commit for ease of review):

- An RPC call to a participant who disconnects mid-call now fails promptly with `RecipientDisconnected`. Previously, pending calls were never purged on disconnect, so the caller waited out its full response timeout (15s by default) and got `ResponseTimeout` instead.
- A server reporting a version that is not valid semver no longer panics the calling task. In practice, this should never happen but it's fairly easy to catch so I think it makes sense to add. An unparseable version is no longer treated as evidence that the server is too old.
- A v1 `RpcResponse` carrying the unused [compressed payload field](https://github.com/livekit/protocol/blob/11ccacca2a47a949e43f5055b01cfb874d20866a/protobufs/livekit_models.proto#L572), or no value at all, now fails with an `ApplicationError` instead of resolving the caller with an empty successful response.
- Removed an unguarded `unwrap` when building a v1 response packet, by giving the function a signature that cannot represent the invalid state.

## TODO
- [x] Switch base back to `main` before merging
- [x] Add `livekit-rpc` crate to crates.io with trusted publishing enabled


livekit-rpc/src/client.rs
```diff
@@ -165,15 +169,15 @@ impl RpcClientManager {
...
-    pub(crate) async fn send_v1_request(
+    async fn send_v1_request(
         &self,
         transport: &impl RpcTransport,
         destination_identity: &str,
         id: &str,
         method: &str,
         payload: &str,
         response_timeout: Duration,
-    ) -> Result<(), crate::room::RoomError> {
```
1egoman, line 176: Note to reviewers - This new `RpcTransportError` properly encapsulates the rpc errors for this new `livekit-rpc` crate. They are converted back to `RoomError` on the `livekit` crate end.

livekit/src/room/rpc.rs
```diff
@@ -0,0 +1,26 @@
...
+//! This module exists purely for source compatibility — it has no callers inside this crate.
+//! Code within `livekit` imports from `livekit_rpc` directly; the production transport lives
+//! in [`super::rpc_transport`].
+
+pub use livekit_rpc::api::*;
+
+// Historically public at `livekit::rpc::*`, but not usable without a transport, which is
+// internal. Kept reachable so existing code compiles, hidden from the docs.
+#[doc(hidden)]
+pub use livekit_rpc::backend::{HandleRequestOptions, RpcClientManager, RpcServerManager};
```
1egoman: Note to reviewers - I've kept all old symbols still reachable under `livekit::room::rpc::*` for backwards compatibility. This arguably is a bit overkill (given these symbols are all reexported from the crate root) and it might be worth just dropping this file outright. Curious what others think here.

livekit-rpc/src/client.rs
```diff
@@ -32,9 +31,24 @@ use tokio::sync::oneshot;
 ///
 /// Tracks pending ACKs and responses, handles v1 packet and v2 data stream
 /// transport selection based on the remote participant's client protocol.
+/// The oldest server version that supports RPC.
+const MIN_RPC_SERVER_VERSION: Version = Version::new(1, 8, 0);
```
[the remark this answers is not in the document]
1egoman: Fixed in 73327b5c0

.changeset/remove_livekit_runtime.md
```diff
@@ -0,0 +1,13 @@
+---
+libwebrtc: patch
+livekit: patch
+livekit-api: patch
+livekit-datatrack: patch
+livekit-ffi: patch
+livekit-net: patch
+livekit-signaling: patch
+livekit-token-source: patch
+livekit-uniffi: patch
```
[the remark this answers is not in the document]
1egoman: Out of date / not relevant

livekit-rpc/src/client.rs
ladvoc: Nice, I think we should be moving towards more contextual errors in more places anyway.

livekit/src/room/rpc.rs
ladvoc: I'm in favor of dropping this file outright and making this a minor breaking change.

livekit-rpc/AGENTS.md
```diff
@@ -0,0 +1,46 @@
+# AGENTS.md
```
ladvoc, line 1: praise: Thank you for adding the scoped AGENTS.md!

livekit-rpc/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+semver = "1.0"
+tokio = { workspace = true, default-features = false, features = ["sync", "time", "rt"] }
+uuid = { version = "1", features = ["v4"] }
+
+[dev-dependencies]
+# Enables `TextStreamReader::new_for_test`, used by the mock transport in `src/tests.rs`.
+livekit-data-stream = { workspace = true, features = ["test-utils"] }
+tokio = { workspace = true, default-features = false, features = ["macros", "rt", "rt-multi-thread", "time"] }
+bytes = { workspace = true }
+chrono = "0.4.38"
```
ladvoc: nitpick: This is used for a few crates now, so probably time to make it a workspace dep.

livekit-rpc/src/types.rs
```diff
@@ -12,106 +12,11 @@
...
-
-pub use client::RpcClientManager;
-pub use server::{HandleRequestOptions, RpcServerManager};
-
-use crate::data_stream::api::{StreamResult, StreamTextOptions, TextStreamInfo};
-use crate::room::id::ParticipantIdentity;
-use crate::room::participant::ClientCapability;
-use livekit_common::RemoteParticipantRegistry;
+use livekit_common::ParticipantIdentity;
 use livekit_protocol::RpcError as RpcError_Proto;
```
ladvoc: nitpick: The established pattern is to do `use livekit_protocol as proto;` at the top and then reference proto versions of structs under that alias (e.g., `proto::RpcError`).

livekit-rpc/src/client.rs
```diff
@@ -62,10 +65,16 @@ impl RpcClientManager {
         let min_effective_timeout = Duration::from_millis(1000);
 
         if let Some(version_str) = transport.server_version() {
-            let server_version = Version::parse(&version_str).unwrap();
-            let min_required_version = Version::parse("1.8.0").unwrap();
-            if server_version < min_required_version {
-                return Err(RpcError::built_in(RpcErrorCode::UnsupportedServer, None));
+            match Version::parse(&version_str) {
```
ladvoc, line 75: praise: Nice catch!

livekit-rpc/src/tests.rs
```diff
@@ -796,6 +796,30 @@ async fn test_server_below_minimum_version_rejected() {
     assert!(transport.texts().is_empty());
 }
 
+/// A server version we cannot parse must not panic, and must not block the call: we fail open
+/// and let the request go out rather than rejecting every RPC on a cosmetic version string.
+#[tokio::test]
+async fn test_unparseable_server_version_allows_call() {
```
ladvoc, line 861: question: I wonder if we should defer testing these kinds of cases until there is a larger refactor? In the data tracks implementation, Protobuf messages are converted into trusted messages with a `TryFrom` implementation (e.g., `proto::DataTrackSubscriberHandles` → `SfuSubscriberHandles`). One advantage of this pattern is you can unit test your `try_from` implementation against malformed Protobuf messages in isolation.

ladvoc (approved):

livekit-rpc/Cargo.toml
1egoman: Fixed in 6c3a03220

livekit-rpc/src/types.rs
1egoman: Fixed in b5522d54e

livekit/src/room/rpc.rs
1egoman: Fixed in 5a8113f8d

.changeset/extract_livekit_rpc_crate.md
```diff
@@ -0,0 +1,40 @@
+---
+livekit-rpc: patch
+livekit: minor
+livekit-ffi: patch
+livekit-data-stream: patch
+livekit-uniffi: patch
```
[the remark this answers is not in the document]
1egoman: The knope bot said I needed to add this, 🤷

livekit-rpc/src/tests.rs
1egoman: IMO, I'd rather have the test coverage than not have it.

That being said, I agree though that this could probably be made more elegant - I think in this case (due to the `RpcTransport` interface in between) the most natural place to put that `Result<...>` actually would be in the [return of server_version here](https://github.com/livekit/rust-sdks/blob/e36598d85c51e5241b90fd08845417de7da09e7d/livekit-rpc/src/tests.rs#L149) rather than using `TryFrom` - so something like:

```rust
    fn server_version(&self) -> Result<Option<Version>, semver::Error> {
        Version::parse(self.server_ver.clone())
    }
```

Thoughts? Is this the better place for this to live, and then the test can be reformed to fit this `RpcTransport` interface instead?

ladvoc: Makes sense, but I'd say that could be done as a follow up task. Let's keep the test.

livekit-rpc/src/client.rs
```diff
@@ -88,8 +108,13 @@ impl RpcClientManager {
...
             let mut pending_responses = self.pending_responses.lock();
-            pending_acks.insert(id.clone(), ack_tx);
-            pending_responses.insert(id.clone(), response_tx);
+            let destination = ParticipantIdentity(data.destination_identity.clone());
+            pending_acks.insert(
+                id.clone(),
+                Pending { destination_identity: destination.clone(), tx: ack_tx },
+            );
+            pending_responses
+                .insert(id.clone(), Pending { destination_identity: destination, tx: response_tx });
```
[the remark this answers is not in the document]
1egoman: Going to do this as a follow up item, I want to get this PR merged.

PR #1377, opened 2026-08-30
Fix Two Memory Leaks in RoomSession and DataChannel

raparanjpe:
This PR fixes two leaks, one with an ArcCycle between a `Room` and itself due to passing a strong reference to the `RoomSession` into a closure (`on_state_changed` contained by the `RoomSession`). For this we simply downgrade the Arc in the same way we do above on line 758 in `mod.rs` for the `local_participant`.

For the second leak we noticed a DataChannel cycle triggered via the `forward_dc_events` call in `rtc_events.rs`, specifically `dc.on_buffered_amount_change(Some(on_buffered_amount_change(rtc_emitter, dc.clone(), kind)));` (i.e. we're cloning the dc and passing it into itself, creating the strong reference cycle).

Initially we were going to just pass the amount from dc directly, but noticed we don't actually need that in the struct, so opted to remove the field directly. If it's needed for other reasons we can add it back without cloning the data channel arc.

The leak is proportional to room joins, and so we set up a simple rig to test joining and leaving rooms to measure the leak amount via jemalloc's `stats.allocated` gauge and got these results:

| Build | Per Room Join (kb) |
|-------|--------------------|
| main  | 162                |
| PR    | 20                 |

Although this is a bit minimal, over many calls a day the leak accumulates quite a bit.

### DIFF
--- a/livekit/src/rtc_engine/rtc_session.rs
+++ b/livekit/src/rtc_engine/rtc_session.rs
@@ -1677,7 +1677,7 @@ impl SessionInner {
                     );
                 }
             }
-            RtcEvent::DataChannelBufferedAmountChange { sent, amount: _, kind } => {
+            RtcEvent::DataChannelBufferedAmountChange { sent, kind } => {
                 let ev = DataChannelEvent {
                     kind,
                     detail: DataChannelEventDetail::BufferedAmountChange(sent),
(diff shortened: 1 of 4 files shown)


livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -1677,7 +1677,7 @@ impl SessionInner {
                     );
                 }
             }
-            RtcEvent::DataChannelBufferedAmountChange { sent, amount: _, kind } => {
```
raparanjpe, line 1680: This was the only place I was able to see amount being used and it's being ignored, so I opted to remove amount instead of wiring it through from the data channel.

PR #1382, opened 2026-08-31
Unpublish from remote tracks in RoomSession::close()

### LINKED ISSUE #1378: Memory Leak with Multiple Remote Publishers
We've identified a memory leak when multiple publishers share a single room. We can reproduce this against `livekit 0.7.49` and `livekit-server v1.9.0`:
1. Run `livekit-server` and have two publishers `A` and `B` join the room *as separate processes*, both publishing an audio track.
2. Have publisher `A` close their connection first with `B` still publishing and have `B` close a few seconds later.
3. Repeat this for many iterations.

You'll observe `proc-A` leaking memory on the live heap (check `stats.allocated` on jemalloc), roughly 70kb per join. This behavior is flipped if we have `B` close the connection first (`proc-B` will leak in that case).

However, we have noticed that if each side explicitly calls `unpublish` on its own tracks before closing the connection, the leak disappears entirely. This feels like a workaround however, instead of the correct way to address the core issue (i.e. the remote publisher does not get cleaned up on `close()` for some reason).

I assume there are some deeper intricacies in the connection closing process we are not privy to, and I'm happy to provide a simple gist or self-contained repro if useful. Notably, this does *not* reproduce if `A` and `B` are in a single process.

Note: The linked PR does resolve the leak when trying to locally reproduce, but not sure if this is the best approach.

raparanjpe:
Closes #1378 

The idea here is to follow a similar procedure to `fn handle_participant_disconnect` where we handle participant disconnects, but we should also do so when closing out a room.


raparanjpe:
Hi @ladvoc, let me know if I can provide more detail here or on the underlying issue, but we've been using this patch to resolve this memory leak in particular for some time now and it would be great to have it merged upstream!

PR #1388, opened 2026-09-01
bugfix: PeerConnectionInterface::AddIceCandidate

stephen-derosa:
# Overview

`PeerConnection::add_ice_candidate` handed libwebrtc a `[&]` lambda that captured `ctx` (the
`rust::Box<PeerContext>`) and `on_complete` by reference — both stack parameters of the shim.
`PeerConnectionInterface::AddIceCandidate` completes asynchronously, so whenever libwebrtc's
operations chain deferred the completion, the shim returned and destroyed the captures first.
Dropping the context dropped the pending `oneshot::Sender`, so the Rust future resolved with
`add_ice_candidate cancelled`; the later callback then read freed memory (`SIGSEGV`/`SIGBUS`).

The callback state now lives in `AddIceCandidateCompletion`, held behind a `shared_ptr` by
`make_add_ice_candidate_callback`. It outlives the call, is safe for the copies libwebrtc makes
(`AddIceCandidate` takes a copyable `std::function`, and `OperationsChain` copies it), and an
atomic guard moves the context back to Rust exactly once.

## Reproducing failure (before merge)

The bug only shows when libwebrtc's operations chain is non-empty: `ChainOperation` runs an
operation immediately when the chain is idle, so the callback normally fires synchronously inside
the blocking proxy marshal and the dangling captures are still alive. Stacking a `CreateOffer` on a
freshly created peer connection (its DTLS certificate is still generating) keeps the chain busy and
defers the completion:

```
res=Err(RtcError { error_type: Internal, message: "add_ice_candidate cancelled" })
signal: 10, SIGBUS: access to undefined memory
```

# Tests

**`webrtc-sys`** — four unit tests over a test-only seam that drives the *production* callback
factory the way libwebrtc does (build it in a frame that returns, copy it, drop the original, then
invoke): survives-the-frame, error forwarding, exactly-once across repeated invocations,
release-on-drop.

**`libwebrtc`** — three integration tests on the existing peer-connection fixture: deferred
completion resolves, close during a deferred completion, and 25 repeated deferred completions. Each
asserts the chain is genuinely occupied first, so they cannot pass vacuously.

Against the old `[&]` implementation: 3/4 `webrtc-sys` tests SIGBUS, all 3 `libwebrtc` tests SIGSEGV.

| Command | Result |
|---|---|
| `cargo fmt -- --check` | clean |
| `clang-format` on new C++ regions | clean |
| `cargo test -p libwebrtc -p webrtc-sys -- --test-threads=1` | 29 passed, 0 failed |
| `cargo clippy -p webrtc-sys -p libwebrtc --all-targets` | no warnings in changed files |
| `cargo check --workspace --all-targets` | no errors |
| 20x repeat under `MallocScribble`/`MallocPreScribble`/`MallocGuardEdges` | 0 failures |

ASan was not run: `-Zsanitizer=address` needs nightly and the repo pins stable `1.97.1`; the
prebuilt libwebrtc is not instrumented either. macOS malloc guards were used instead.

# API breaking changes

None. `PeerConnection::add_ice_candidate` keeps its signature and its success/error results; the
change is confined to how the native shim owns the completion state.

### DIFF
--- a/.changeset/add_ice_candidate_completion_ownership.md
+++ b/.changeset/add_ice_candidate_completion_ownership.md
@@ -1,0 +1,19 @@
+---
+webrtc-sys: patch
+libwebrtc: patch
+---
+
+# Keep `add_ice_candidate` completion state alive until libwebrtc completes
+
+The native `PeerConnection::add_ice_candidate` shim passed libwebrtc a lambda that captured
+the Rust context and completion function by reference, even though
+`PeerConnectionInterface::AddIceCandidate` completes asynchronously. Whenever libwebrtc's
+operations chain deferred the completion — for example when a `CreateOffer` or
+`SetRemoteDescription` was still in flight — the shim returned first and destroyed both
+captures. Dropping the context dropped the pending `oneshot::Sender`, so the Rust future
+resolved with `add_ice_candidate cancelled`, and the later callback read freed memory,
+crashing with `SIGSEGV`/`SIGBUS`.
+
+The completion state now lives in a shared, single-use object that outlives the call, is
+safe for the copies libwebrtc makes of the callback, and hands the context back to Rust
+exactly once.
--- a/libwebrtc/src/peer_connection.rs
+++ b/libwebrtc/src/peer_connection.rs
@@ -273,6 +273,8 @@ impl Debug for PeerConnection {
 
 #[cfg(test)]
 mod tests {
+    use std::{future::Future, pin::Pin, task};
+
     use log::trace;
     use tokio::sync::mpsc;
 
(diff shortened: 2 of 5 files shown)


.changeset/add_ice_candidate_completion_ownership.md
```diff
@@ -0,0 +1,19 @@
...
+The native `PeerConnection::add_ice_candidate` shim passed libwebrtc a lambda that captured
+the Rust context and completion function by reference, even though
+`PeerConnectionInterface::AddIceCandidate` completes asynchronously. Whenever libwebrtc's
+operations chain deferred the completion — for example when a `CreateOffer` or
+`SetRemoteDescription` was still in flight — the shim returned first and destroyed both
+captures. Dropping the context dropped the pending `oneshot::Sender`, so the Rust future
+resolved with `add_ice_candidate cancelled`, and the later callback read freed memory,
+crashing with `SIGSEGV`/`SIGBUS`.
+
+The completion state now lives in a shared, single-use object that outlives the call, is
```
stephen-derosa, line 17: this is sort of the main item to highlight in the whole PR

cliqer:
Read through this one properly since it overlaps with #1392; the completion ownership and the single-use claim look right to me, and the three `libwebrtc` tests reproduce the crash on the unpatched shim (verified locally on macOS arm64: `add_ice_candidate cancelled` followed by the test process dying, then 3/3 green with the fix). Two things blocking or worth a look before merge:

1. **Changeset Check fails.** The changeset bumps only `webrtc-sys` and `libwebrtc`, but the check requires the dependents as well, so it needs `livekit: patch` and `livekit-ffi: patch` too. Same failure #1392 hit; that front matter is at https://github.com/livekit/rust-sdks/blob/fix/add-ice-candidate-dangling-capture/.changeset/add_ice_candidate_capture_by_value.md if you want to copy it.

2. **The test seam is compiled into every build.** `webrtc-sys/build.rs` defines `LIVEKIT_TEST` unconditionally (there is a `TODO(theomonnom)` on that line saying it should be tests only), so `complete_add_ice_candidate_for_test` ships in release builds. Harmless dead code, but the `#ifdef LIVEKIT_TEST` guard reads as test-only when it is not. Fine to leave for a follow-up, just flagging it so the comment does not mislead.

Happy to close #1392 once this lands.

stephen-derosa:
consumed by https://github.com/livekit/rust-sdks/pull/1392

PR #1392, opened 2026-09-02
fix(webrtc-sys): capture add_ice_candidate completion state by value

cliqer:
## Problem

`PeerConnection::add_ice_candidate` (webrtc-sys/src/peer_connection.cpp) passes libwebrtc a completion lambda that captures `ctx` and `on_complete` by reference:

```cpp
peer_connection_->AddIceCandidate(
    candidate->release(), [&](const webrtc::RTCError& err) {
      on_complete(std::move(ctx), to_error(err));
    });
```

`PeerConnectionInterface::AddIceCandidate` with a callback completes asynchronously on the signaling thread, and libwebrtc defers it behind the operations chain whenever the chain is busy (typically while a `SetRemoteDescription` is still in flight). In that case the lambda runs after `add_ice_candidate` has returned, so `ctx` and `on_complete` are dangling references into a dead stack frame. The deferred call then invokes freed memory.

Observed in production as a SIGSEGV on the `signaling_thread` with PC=0 / LR=0 (a null function pointer call) on the first remote ICE candidate that arrived while the answer was being applied. Reproduces reliably when candidates are forwarded immediately after `set_remote_description` without waiting for it to settle.

## Fix

Move `ctx` into a `shared_ptr` owned by the lambda and capture `on_complete` by value, so the callback holds its own state for as long as libwebrtc keeps it. No API change.

We have shipped this as a vendored patch for several months on macOS desktop builds and the crash has not recurred.


cliqer:
The `Test (x86_64-pc-windows-msvc)` failure is `test_e2ee` (livekit integration test), which is unrelated to this change: webrtc-sys builds and every other test passes on Windows.

The same test fails identically on `main` on the Windows runner, for example run 33550195134 (2026-09-01, 14 passed / 1 failed / test_e2ee) and run 33463303270, while other recent main runs pass, so it looks like a flaky test on that runner rather than a regression from this PR. I do not have permission to rerun the job from a fork.

stephen-derosa (commented):
@cliqer thanks for opening this fix! We have a PR for i believe the same bug here:
https://github.com/livekit/rust-sdks/pull/1388. #1388 does guard against a second invocation of completion, but to be frank I dont think this is a reachable code path is reachable. It also adds tests for the explicit dereferencing of freed stack memory.

Would you be open to pulling in the relevant tests from #1388 into this PR?

Im glad you have been proving out this fix for the last few months!

cliqer:
@stephen-derosa happy to. Pulled the three `libwebrtc` regression tests from #1388 into this branch as-is (deferred completion resolves, close during a deferred completion, 25 repeats of the deferred path). Ran them locally on macOS arm64 against the current prebuilt:

- with this fix: 3 passed, 0 failed
- with the unpatched `add_ice_candidate`: the first test fails with `add_ice_candidate cancelled` and the test process dies on the freed callback state, so they do reproduce the bug

I left out the `webrtc-sys` seam tests, since they exercise the double-invocation guard of #1388's `AddIceCandidateCompletion`, which as you say is not a reachable path with libwebrtc; this PR keeps the shim to the shared context capture only. If you would rather land #1388 with its guard and tests, I am fine closing this one, whichever gets the crash out of the next release faster.

webrtc-sys/src/peer_connection.cpp
```diff
@@ -170,9 +170,10 @@ void PeerConnection::add_ice_candidate(
     std::shared_ptr<IceCandidate> candidate,
     rust::Box<PeerContext> ctx,
     rust::Fn<void(rust::Box<PeerContext>, RtcError)> on_complete) const {
+  auto owned_ctx = std::make_shared<rust::Box<PeerContext>>(std::move(ctx));
   peer_connection_->AddIceCandidate(
-      candidate->release(), [&](const webrtc::RTCError& err) {
-        on_complete(std::move(ctx), to_error(err));
+      candidate->release(), [owned_ctx, on_complete](const webrtc::RTCError& err) {
+        on_complete(std::move(*owned_ctx), to_error(err));
```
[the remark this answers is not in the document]
cliqer: Addressed: the context now lives in a shared `std::optional` and is claimed with `std::exchange` before the Rust callback runs, so any further invocation of a callback copy is a no-op. The three deferred-completion tests still pass (3/3 locally).

stephen-derosa (approved):
approving since this is a bit cleaner than #1388 . Please wait for @lukasIO review before merging since he has a better understanding than i do on the intricacies here

webrtc-sys/src/peer_connection.cpp
```diff
@@ -170,9 +172,12 @@ void PeerConnection::add_ice_candidate(
     std::shared_ptr<IceCandidate> candidate,
     rust::Box<PeerContext> ctx,
     rust::Fn<void(rust::Box<PeerContext>, RtcError)> on_complete) const {
+  auto owned_ctx = std::make_shared<std::optional<rust::Box<PeerContext>>>(
```
stephen-derosa: I think you can revert `01466f356eb8327d939fdb8ea9fcedef069ca941` to guard against the duplicate calls of callbacks by webrtc, then you are good to merge!

stephen-derosa: also drop the respective tests

cliqer:
Done: reverted `01466f3` and dropped the three libwebrtc tests, so the PR is back to the shared-context capture only. CI red on `Test API` is `sip_busy` hitting the live server (fails on every PR today, unrelated).

lukasIO (approved):

stephen-derosa:
@cliqer after you rebase, tests should pass and we can merge in

cliqer:
Rebased onto main (`4dfca31`), now includes #1398 so the `sip_busy` check should go green. Net diff unchanged: the shared-context capture in `peer_connection.cpp` plus the changeset.

Thanks @stephen-derosa and @lukasIO for the quick reviews and for sorting the test-server drift. Happy to see this land; it lets us drop the vendored `webrtc-sys` patch on our side.

stephen-derosa:
@cliqer this will be in the next release!

PR #1394, opened 2026-09-02
Fix Objective-C linking

ladvoc:
The ObjC categories in _libwebrtc.a_ (e.g. `NSString (StdString)`) export no symbols, so the linker drops them and macOS/iOS binaries abort at startup:
```
+[NSString stringForStdString:]: unrecognized selector sent to class
```

The usual fix, `-ObjC` (currently documented in the README as required for consumers), can't be applied at the crate level since Cargo doesn't propagate link args to dependent crates. Instead, merge all `obj/sdk` objects into one archive member when assembling _libwebrtc.a_, so the categories load together with the referenced SDK symbols and consumers need no flag as of the next prebuilt release.

After this merges, [_livekit-examples/rust-dev-client_](https://github.com/livekit-examples/rust-dev-client) should be updated to remove the workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,14 +203,6 @@ cargo build
   
 ```
 
-### MacOS
-
-When building on MacOS, `-ObjC` linker flag is needed. LiveKit's WebRTC implementation make use of ObjectiveC libraries on the Mac. You may get the following error if the app isn't linked with ObjC:
-
-```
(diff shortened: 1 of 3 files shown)


pblazej (approved):
Works fine locally

PR #1395, opened 2026-09-02
CudaContext lifecycle memory fixes

alan-george-lk:
This PR addresses two issues related to memory lifecycle/safety: `CudaContext` and `FfiServer` cleanup on dispose.

CudaContext:

- Prior to this PR, users could inadvertently re-initialize memory via `CudaContext` multiple times
- Similarly, the decoders that initialized the context never destroyed it during their destructors
- This addresses both issues and adds thread safety and debug logs

FfiServer:

- Adds proper cleanup to dispose()
- Adds tests around this

### DIFF
--- a/livekit-ffi/src/server/mod.rs
+++ b/livekit-ffi/src/server/mod.rs
@@ -187,10 +187,14 @@ impl FfiServer {
         self.logger.set_capture_logs(false);
 
         // Closing rooms does not release resources owned by FFI handles.
-        // Cancel handle watchers before dropping all stored native resources.
+        // Cancel handle watchers first, then drop handles one by one so Drop
+        // impls can call drop_handle without re-entering DashMap::clear().
         *self.config.lock() = None;
         self.handle_dropped_txs.clear();
-        self.ffi_handles.clear();
+        let leftover: Vec<_> = self.ffi_handles.iter().map(|entry| *entry.key()).collect();
+        for id in leftover {
+            let _ = self.ffi_handles.remove(&id);
+        }
     }
 
     pub fn send_event(&self, message: proto::ffi_event::Message) -> FfiResult<()> {
--- a/livekit-ffi/src/server/mod.rs
+++ b/livekit-ffi/src/server/mod.rs
@@ -170,6 +170,10 @@ impl FfiServer {
             .collect()
     }
 
+    /// Closes active rooms and releases every resource owned by an FFI handle.
+    ///
+    /// Clearing the handle map is required even after rooms are closed because
+    /// room handles retain their RTC engines and peer connection factories.
     pub async fn dispose(&'static self) {
         log::debug!("disposing ffi server");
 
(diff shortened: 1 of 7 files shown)


livekit-ffi/src/server/mod.rs
```diff
@@ -262,10 +273,10 @@ impl FfiServer {
     pub fn drop_handle(&self, id: FfiHandleId) -> bool {
         let existed = self.ffi_handles.remove(&id).is_some();
         self.handle_dropped_txs.remove(&id);
-        if !existed {
+        if !existed && self.is_setup() {
```
alan-george-lk, line 276: Double-check this is worth it

livekit-ffi/src/server/mod.rs
```diff
@@ -182,8 +186,15 @@ impl FfiServer {
...
         self.logger.set_capture_logs(false);
 
-        // Drop all handles
-        *self.config.lock() = None; // Invalidate the config
+        // Closing rooms does not release resources owned by FFI handles.
+        // Cancel handle watchers first, then drop handles one by one so Drop
+        // impls can call drop_handle without re-entering DashMap::clear().
+        *self.config.lock() = None;
+        self.handle_dropped_txs.clear();
+        let leftover: Vec<_> = self.ffi_handles.iter().map(|entry| *entry.key()).collect();
```
alan-george-lk, line 194: Rust novice question: is this the cleanest way to grab remaining handles before iterating? Feels a bit complex

webrtc-sys/src/nvidia/cuda_context.cpp
```diff
@@ -36,6 +37,13 @@ namespace livekit_ffi {
 static void* s_module_ptr = nullptr;
 static const int kRequiredDriverVersion = 11000;
 
+// Serializes access to the singleton context, its reference count, and the
+// dynamically loaded CUDA module.
+static std::mutex& cudaMutex() {
```
alan-george-lk, line 42: Not really sure the value of this singleton helper vs. just a `static mutex` up above. Either way static memory sucks

webrtc-sys/src/nvidia/cuda_context.cpp
```diff
@@ -183,7 +201,16 @@ CUcontext CudaContext::GetContext() const {
 }
 
 void CudaContext::Shutdown() {
-  // Shutdown CUDA context
+  std::lock_guard<std::mutex> lock(cudaMutex());
+  if (ref_count_ == 0) {
```
alan-george-lk, line 211: The refcount kind of sucks, but is needed because both the encoder/decoder factories reference this.

livekit-ffi/src/server/mod.rs
ladvoc: This works, but you should be able to avoid the call to `collect` (which allocates) and just consume the iterator directly using the for loop.

ladvoc (approved):

livekit-ffi/src/server/mod.rs
xianshijing-lk, line 276: curiously, why self.is_setup() is needed here?

xianshijing-lk (approved):
lgtm, one question

livekit-ffi/src/server/mod.rs
alan-george-lk: There's a misleading warning log that emits during shutdown sometimes “Attempted to drop unknown FFI handle”, where this code can be re-entered (valid code path). This was just an easy fix to no-op / prevent that log in that case

livekit-ffi/src/server/mod.rs
alan-george-lk: Appreciate the suggestion, although I changed it and the test that covers this now deadlocks. I'll keep the original code since it seems to be stable.

PR #1398, opened 2026-09-03
sim_busy: failed_precondition -> failed_precondition

stephen-derosa:
Updates test to match expected results from here: https://github.com/livekit/protocol/pull/1766

### DIFF
--- a/.changeset/update_test_to_use_new_sip_grpc_code.md
+++ b/.changeset/update_test_to_use_new_sip_grpc_code.md
@@ -1,0 +1,5 @@
+---
+livekit-api: patch
+---
+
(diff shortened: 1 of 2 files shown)


ChelseaBradbury:
I think we should update this docstring also:
https://github.com/livekit/rust-sdks/blob/c5959abaf507727eb3bff26d4161d3255f163a26/livekit-api/src/services/mod.rs#L145

It looks like each sdk has roughly the same docstring somewhere

nishadmusthafa (approved):

ChelseaBradbury (approved):
Not going to block over the docs; I can take care of that if we don't add it here

PR #1400, opened 2026-09-04
Video source lifecycle memory fixes

alan-george-lk:
## Summary

* Stop the raw-video keepalive task from cloning NativeVideoSource. It now uses a Weak liveness check so dropping the source actually releases the native handle and black I420 keepalive buffer.
* Previously, a source that never received captureFrame kept that task alive forever (~1.32 MiB per unused 720p source). 
* Also pair CUDA context init/shutdown on NVIDIA encoder/decoder factories, drop leftover FFI handles in dispose(), and fail closed if nvEncInitializeEncoder throws.

## Testing

* Adds regression tests directly
* 1,000 create/drop cycles peaked at ~1.35 GiB; after the fix they peak at ~41 MiB.

### DIFF
--- a/libwebrtc/src/native/video_source.rs
+++ b/libwebrtc/src/native/video_source.rs
@@ -15,7 +15,7 @@
 use std::{
     sync::{
         atomic::{AtomicUsize, Ordering},
-        Arc,
+        Arc, Weak,
     },
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
@@ -48,6 +48,12 @@ pub struct NativeVideoSource {
     captured_frames: Arc<AtomicUsize>,
 }
 
+fn keepalive_should_continue(captured_frames: &Weak<AtomicUsize>) -> bool {
+    captured_frames
+        .upgrade()
+        .is_some_and(|captured_frames| captured_frames.load(Ordering::Relaxed) == 0)
+}
+
 impl NativeVideoSource {
     pub fn new(resolution: VideoResolution, is_screencast: bool) -> NativeVideoSource {
         Self::new_inner(resolution, is_screencast, true)
@@ -77,8 +83,9 @@ impl NativeVideoSource {
         };
 
         if raw_keepalive {
+            let captured_frames = Arc::downgrade(&source.captured_frames);
+            let sys_handle = source.sys_handle.clone();
             tokio::spawn({
-                let source = source.clone();
                 // This buffer reaches the encoder without any plane ever being
                 // written, so it must be black-initialized: `I420Buffer::new`
                 // leaves the pixel data uninitialized and would leak recycled
@@ -90,7 +97,7 @@ impl NativeVideoSource {
                     loop {
                         interval.tick().await;
 
-                        if source.captured_frames.load(Ordering::Relaxed) > 0 {
+                        if !keepalive_should_continue(&captured_frames) {
                             break;
                         }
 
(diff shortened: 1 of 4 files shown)


libwebrtc/src/native/video_source.rs
```diff
@@ -77,42 +126,9 @@ impl NativeVideoSource {
...
-                                frame_id: 0,
-                                user_data: Vec::new(),
-                            },
-                        );
-                    }
-                }
-            });
+            let captured_frames = Arc::downgrade(&source.captured_frames);
+            let sys_handle = source.sys_handle.clone();
+            drop(spawn_raw_keepalive(resolution, captured_frames, sys_handle));
```
ladvoc: nitpick: Wrapping with `drop` is redundant; drop is implicit at end of scope.

libwebrtc/src/native/video_source.rs
```diff
@@ -48,6 +49,54 @@ pub struct NativeVideoSource {
     captured_frames: Arc<AtomicUsize>,
 }
 
+fn keepalive_should_continue(captured_frames: &Weak<AtomicUsize>) -> bool {
+    captured_frames
+        .upgrade()
+        .is_some_and(|captured_frames| captured_frames.load(Ordering::Relaxed) == 0)
+}
+
+fn spawn_raw_keepalive(
```
ladvoc: nitpick: Consider marking this function async, renaming to `raw_keepalive_task`, and moving `tokio::spawn` inline. If you want to get fancier, you could use this pattern which I particularly like since it avoids positional arguments, but might be overkill in this case:
```rust
struct RawKeepaliveTask {
    resolution: VideoResolution,
    captured_frames: Weak<AtomicUsize>,
    sys_handle: SharedPtr<vt_sys::ffi::VideoTrackSource>,
}

impl RawKeepaliveTask {
    async fn run(self) {
       // notice this consumes self
       // same as current impl.
   }
}
```

libwebrtc/src/native/video_source.rs
```diff
@@ -48,6 +49,54 @@ pub struct NativeVideoSource {
...
+        // leaves the pixel data uninitialized and would leak recycled
+        // heap memory to subscribers in the first keyframes.
+        let i420 = I420Buffer::new_black(resolution.width, resolution.height);
+        async move {
+            let mut interval = interval(Duration::from_millis(100)); // 10 fps
+
+            loop {
+                interval.tick().await;
+
+                if !keepalive_should_continue(&captured_frames) {
```
ladvoc: note: Not related to changes in this PR, but I wonder what happens if you initially publish some frames (which would kill this task) but later stop publishing frames (e.g., camera temporally goes offline so there are no frames available). This might be the correct logic, but worth verifying with WebRTC.

ladvoc (approved):

libwebrtc/src/native/video_source.rs
```diff
@@ -48,6 +49,54 @@ pub struct NativeVideoSource {
...
+                    break;
+                }
+
+                let mut builder = vf_sys::ffi::new_video_frame_builder();
```
ladvoc: note: Also not related to this PR, but this is something we should work to eliminate over time; FFI details between Rust and libwerbtc should not be public API to this layer.

libwebrtc/src/native/video_source.rs
alan-george-lk: Fixed

libwebrtc/src/native/video_source.rs
alan-george-lk: Implemented first suggestion

PR #1402, opened 2026-09-04
refactor(signaling): explicit signal lifecycle state machine

lukasIO:
SignalInner tracked its lifecycle in two fields, a stream slot and a `reconnecting` AtomicBool, that had to be kept in step by hand across restart, set_reconnected, close and send. It now holds one SignalState that owns the transport: Connected, Reconnecting, Offline, Disconnecting, Closed.

Every change goes through SignalState::transition, one match that is the whole table and also says which transport each move releases. An input a state does not accept is logged and refused; a resume from a state that cannot accept it fails with SignalError::InvalidState. The resume gate stays a state until the engine confirms, a stale transport cannot report in because SignalClient::restart awaits the old task first, and ReconnectFailed always lands in Offline.

The held-signal queue moves to a sync parking_lot Mutex that is never held across an await. The old async queue lock was taken in both orders relative to the stream lock, which with tokio's fair RwLock could deadlock against a pending restart writer. A send that fails with any transport error is now held like a SendError was.


livekit-signaling/src/lib.rs
```diff
@@ -205,11 +210,96 @@ pub enum SignalEvent {
...
+
+impl SignalState {
+    /// The transition table.
+    /// `Ok`: the transport the old state owned and the new one does not - close it.
+    /// `Err`: refused, logged, nothing moved.
+    fn transition(&mut self, input: SignalInput) -> Result<Option<SignalStream>, SignalInput> {
+        use SignalInput as In;
+        use SignalState::*;
+        // `Closed` is a placeholder while the old state is moved out
+        let (next_state, released_stream) = match (std::mem::replace(self, Closed), input) {
```
1egoman: nitpick: I brought this up in our 1:1 but it might be slightly cleaner if SignalState::Closed was the `Default` value for `SignalState` and then this could become `std::mem::take(self)` given the replaced value is bogus anyway.


```suggestion
        let (next_state, released_stream) = match (std::mem::take(self), input) {
```

livekit-signaling/src/lib.rs
```diff
@@ -542,25 +632,19 @@ impl SignalInner {
...
     /// Restart is called when trying to resume the room (RtcSession resume).
     ///
-    /// Leaves `reconnecting=true` on success — the engine is expected to call
-    /// [`Self::set_reconnected`] once the full resume has succeeded. On failure
-    /// resets `reconnecting=false` so subsequent retries can re-enter cleanly.
-    /// The stream slot is held under a write lock for the entire close + new
-    /// connect, so concurrent senders block on the read side until the new
-    /// stream is in place.
+    /// The write lock is held for the entire close + connect, so concurrent senders wait on
+    /// the read side and land on the new transport instead of in a transport-less gap.
```
1egoman, line 672: thought: This is a pretty clever way to prioritize locks during the close and connect phases. That being said I also think it is potentially somewhat unintuitive. Maybe it makes sense to move this comment to the `AsyncRwLock` field definition in the struct?

livekit-signaling/src/lib.rs
```diff
@@ -577,105 +661,99 @@ impl SignalInner {
...
+        // Already closing or closed: whoever owns that close finishes it.
+        let Ok(stream) = self.state.write().await.transition(SignalInput::Close) else {
+            return;
+        };
+        if let Some(stream) = stream {
             stream.close(notify_close).await;
         }
+        let _ = self.state.write().await.transition(SignalInput::CloseComplete);
+    }
+
```
1egoman, line 742: thought: I'm not sure about this idea or not, but I wonder if there would be benefit to driving the state machine side effects by subscribing to a stream of state changes from the state machine, not just running the side effects alongside at every `.transition(...)` call sizte. What this would look like is some handler consuming the stream of state changes and that handler containing a big `match` not unlike how `.transition(...)` works today.

I will say one nice thing about that is it means that places which all issue the same `.transition(...)` call would be guaranteed to run the same side effects, which could be nice.

.changeset/signal_state_enum.md
```diff
@@ -0,0 +1,16 @@
+---
+livekit-signaling: patch
+---
+
+# Model the signal connection lifecycle as an explicit state machine
+
+`SignalInner` tracked its lifecycle in two fields, a stream slot and a `reconnecting` flag,
+that had to be kept in step by hand. It now holds one `SignalState` that owns the transport
+(`Connected`, `Reconnecting`, `Offline`, `Disconnecting`, `Closed`). Every change goes through one
```
1egoman: nitpick: this file and `.changeset/refactor_signaling_explicit_signal_lifecycle_state_machine.md` look to be semantically identical?

1egoman (approved):
Generally makes sense to me. I focused more on the infrastructure and general approach, and less on the exact states and all the exact transitions between them.

One thing missing which I was expecting to see: some sort of mechanism that a visualizer could subscribe to state changes and use them to build a visualization. Was this expected to be part of a follow up change or did I miss it in here somewhere?

livekit-signaling/src/lib.rs
lukasIO: Yeah, agree, I would like to move into that direction eventually.
I think it should be rather easy to transition (pun intended) to that design later on once we also represent the PC connection as more of a state machine pattern?

livekit-signaling/src/lib.rs
lukasIO: The thing that doesn't sit quite right with me about this is that it's not really the default state. 
The way the signal connection is set up (this has been the case before this PR) is that the signal inner only gets constructed after the JoinResponse has been received. 

While thinking about this I realised that ideally there's another small restructure so that we at least get `Connecting` as an explicit initial state. I added this and also made connecting the default now, so we can use `take` here.

.changeset/signal_state_enum.md
lukasIO: thanks, yeah, deleted it and added the missing crates to the other file

lukasIO:
> Was this expected to be part of a follow up change or did I miss it in here somewhere?

yeah, that's intended as follow up work

livekit-signaling/src/lib.rs
lukasIO: I feel like the context makes sense on restart, as this is where the lock aquisition actually happens. 

I also added a short comment additionally on the struct like you suggested.

livekit-signaling/src/lib.rs
1egoman: Nice, yes, `Connecting` makes even more sense! What you did looks good to me.

livekit-signaling/src/lib.rs
1egoman: Makes sense, cool, let's revisit it later then.

PR #1405, opened 2026-09-04
Room lifecycle memory fixes

alan-george-lk:
## Summary

- Break the `RoomSession` ownership cycle by capturing it weakly from the E2EE callback
- Clear data-channel callbacks during RTC teardown, breaking the observer/callback self-cycle
- Add regression tests verifying both room-session destruction and data-channel callback cleanup

## Memory impact

Previously, repeated room joins retained room sessions, SDP/codec allocations, peer connections, and WebRTC threads. After these fixes, 1,000 connect/disconnect cycles dropped from approximately 488 MiB RSS to 44 MiB, with stable thread and file-descriptor counts.

## Testing

- New room lifecycle E2E regression passed
- C++ 1,000-cycle memory lifecycle tester (representative of robotics use case) passed

### DIFF
--- a/livekit/src/rtc_engine/rtc_session.rs
+++ b/livekit/src/rtc_engine/rtc_session.rs
@@ -1998,6 +1998,10 @@ impl SessionInner {
         // closing here — before the future can suspend — is what stops a cancelled
         // `close()` leaving the ICE sockets bound for the process's lifetime. The
         // signalling socket is unaffected, so the Leave below still goes out.
+        self.lossy_dc.on_message(None);
+        self.lossy_dc.on_buffered_amount_change(None);
+        self.reliable_dc.on_message(None);
+        self.reliable_dc.on_buffered_amount_change(None);
         self.publisher_pc.close();
         if let Some(ref sub_pc) = self.subscriber_pc {
             sub_pc.close();
--- a/livekit/src/rtc_engine/rtc_session.rs
+++ b/livekit/src/rtc_engine/rtc_session.rs
@@ -1675,7 +1675,7 @@ impl SessionInner {
                     );
                 }
             }
-            RtcEvent::DataChannelBufferedAmountChange { sent, amount: _, kind } => {
+            RtcEvent::DataChannelBufferedAmountChange { sent, kind } => {
                 let ev = DataChannelEvent {
                     kind,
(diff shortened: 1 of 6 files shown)


raparanjpe:
Hi @alan-george-lk! Thank you for proposing this fix; I believe I followed a similar approach to weakly capturing the RoomSession in #1377. For the data channel I was going to follow a similar approach to you here, but noticed we can avoid capturing the data channel in its own closure altogether by changing the interface slightly. Feel free to use that implementation if it's easier than manually unsetting the closures.

ladvoc (approved):

alan-george-lk:
> Hi @alan-george-lk! Thank you for proposing this fix; I believe I followed a similar approach to weakly capturing the RoomSession in #1377. For the data channel I was going to follow a similar approach to you here, but noticed we can avoid capturing the data channel in its own closure altogether by changing the interface slightly. Feel free to use that implementation if it's easier than manually unsetting the closures.

Hey @raparanjpe, great minds think alike -- I just reviewed your PR. I think removing `amount` is cleaner, but there are test items/a PR stack here addressing a few related changes.

~~Can you retarget your PR to this one in the next day or two, keeping my changes except for the data channel related ones (taking your approach instead)? Then I can re-test everything with a memory tracker util I have on a separate SDK that brings in these changes.~~

I have pulled in your commit such that you get credit. We may merge this before you have a chance to review, but I'll make sure your name is in the author credits.

raparanjpe:
@alan-george-lk no problem at all!

livekit/src/rtc_engine/rtc_session.rs
```diff
@@ -1998,6 +1998,10 @@ impl SessionInner {
         // closing here — before the future can suspend — is what stops a cancelled
         // `close()` leaving the ICE sockets bound for the process's lifetime. The
         // signalling socket is unaffected, so the Leave below still goes out.
+        self.lossy_dc.on_message(None);
```
raparanjpe: You don't need to set these to None anymore as the closure isn't capturing the arc reference on these data channels!

alan-george-lk: Good catch -- meant to remove these, thanks. Done now.

xianshijing-lk (approved):

PR #1406, opened 2026-09-04
Agent guidance around memory lifecycle edges

alan-george-lk:
Introduce `AGENTS.md` guidance around some of the fixes done in this PR stack, to hopefully avoid/catch future regressions.


ladvoc (approved):
Nice! These all seem like good suggestion. My only question is whether we should move the more specific points (e.g. FFI, C++) to separate AGENTS.md files within the related crates to keep the top-level one lean.

PR #1418, opened 2026-09-09
Add pre-encoded video ingest FFI

stephen-derosa:
# Overview
Expose Pre-Encoded Ingest through ffi


livekit-ffi/src/server/video_source.rs
```diff
@@ -16,9 +16,13 @@ use super::{colorcvt, FfiHandle};
 use crate::{proto, server, FfiError, FfiHandleId, FfiResult};
 use livekit::webrtc::{
     prelude::*,
-    video_frame::{FrameMetadata, VideoFrame},
+    video_frame::{
+        EncodedFrameType, EncodedVideoCodec, EncodedVideoFrame, FrameMetadata, VideoFrame,
+    },
 };
 
+const MAX_ENCODED_VIDEO_FRAME_SIZE: usize = 64 * 1024 * 1024;
```
ladvoc: question: Where does this value come from?

livekit-ffi/src/server/video_source.rs
```diff
@@ -94,12 +127,100 @@ impl FfiVideoSource {
         }
         Ok(())
     }
+
+    /// # Safety
+    ///
+    /// `capture.data_ptr` must address `capture.data_len` readable bytes for
+    /// the duration of this call.
+    pub unsafe fn capture_encoded_frame(
```
ladvoc: suggestion: Instead of marking this whole function unsafe, consider doing this instead:
- In the protobuf, encapsulate pointer/length fields in their own message
- Create unsafe helper method to convert this to Rust slice
- Use this in `capture_encoded_frame`

This keeps unsafe code scoped and auditable.

ladvoc (approved):

livekit-ffi/src/server/video_source.rs
stephen-derosa: arbitrary and removed -- it was a guard against frames that are large, granted this is way too small.

livekit-ffi/src/server/video_source.rs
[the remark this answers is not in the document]
stephen-derosa: this is the case for all video tracks, we shouldnt implement a limit here.

PR #1419, opened 2026-09-09
livekit-token: the agent grant, the kind claim, and no default claims

theomonnom:
`VideoGrants` gains the `agent` grant and `Claims` gains the `kind` claim, with `AccessToken::with_kind`. The Go, Python and JS SDKs already carry both. `livekit-uniffi` exposes them too: `TokenOptions.kind`, `Claims.kind`, and `agent` on its `VideoGrants` record.

**Why:** an agent worker authenticates to `/agent` with `VideoGrants(agent=True)`, and a simulated job's participant token is `kind: "agent"`. Neither could be minted from Rust, so livekit/agents-rust was about to hand-roll these claims over jsonwebtoken instead of depending on this crate.

**The inferred grants become `Option<bool>` (breaking).** The server infers four grants when absent (`protocol/auth/grants.go`): `canPublish` and `canSubscribe` → true, `canUpdateOwnMetadata` → false, and `canPublishData` → whatever `canPublish` is. The Go and JS SDKs model these as optional; this crate had them as `bool` with a `default_true` on the read side, which cannot say "unset" and read an absent `canPublishData` as `true` rather than as `canPublish`. Now `None` leaves the decision to the server and the getters (`can_publish()`, `can_subscribe()`, `can_publish_data()`, `can_update_own_metadata()`) read a token the way the server does. Every other default checked against the server's getters matches. No crate in this workspace sets or reads those four fields directly.

**No default claims.** With the four as `Option`, `VideoGrants` derives `Default`, and nothing at its default is written into the token: every claim and grant carries `skip_serializing_if = "is_default"`, one predicate, no hand-written `Serialize`. Unset claims (`sub`, `name`, `kind`, `metadata`, `sha256`, `attributes`, `roomConfig`) and grants at their default (`video`, `sip`, and every field in them) are omitted, as the server's own `omitempty` grants are. Verification of existing tokens is unchanged.

**Changeset:** `major` for `livekit-token` (pre-1.0, so the actual bump is semver-minor), `livekit-uniffi` (its record shape changed) and `livekit-api` (it re-exports the crate as `livekit_api::access_token`, so `^0.1.2` would otherwise miss the release); `patch` for the internal users `livekit`, `livekit-ffi`, `livekit-signaling`.

Tests: the two new claims round-trip through `to_jwt`/`verify`; a token with nothing set carries exactly `exp`, `iss`, `nbf`; `{}` and `VideoGrants::default()`/`SIPGrants::default()`/a bare `Claims` are the same thing in both directions; an absent `canPublishData` reads as `canPublish`. `livekit-uniffi`, `livekit-api` and `livekit-token-source` compile.

### DIFF
--- a/.changeset/token_agent_grant_and_kind.md
+++ b/.changeset/token_agent_grant_and_kind.md
@@ -1,0 +1,30 @@
+---
+livekit-token: major
+livekit-uniffi: major
+livekit-api: major
+livekit: patch
+livekit-ffi: patch
+livekit-signaling: patch
+---
+
+`VideoGrants` gains the `agent` grant and `Claims` the `kind` claim (with
+`AccessToken::with_kind`), which the Go, Python and JS SDKs already carry. An
+agent worker's token is `VideoGrants { agent: true }` and a simulated job's
+participant token is `kind: "agent"`; neither could be minted from Rust before.
+`livekit-uniffi` exposes both: `TokenOptions.kind`, `Claims.kind`, and `agent`
+on its `VideoGrants` record.
+
+**Breaking:** the four grants the server infers when absent -- `can_publish`,
+`can_subscribe`, `can_publish_data`, `can_update_own_metadata` -- are now
+`Option<bool>`, as in the Go and JS SDKs. `None` leaves the decision to the
+server, and the new getters (`can_publish()`, `can_subscribe()`,
+`can_publish_data()`, `can_update_own_metadata()`) read a token the way the
+server does, `can_publish_data` falling back to `can_publish` included. Code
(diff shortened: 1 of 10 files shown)


ladvoc (changes requested):
Generally looks good, a few points to address:
- Since we expose token generation/verification through `livekit-uniffi` now, you'll need to update there to use the new API.
- Changeset should specify major as the release type (this is necessary because it is a pre-1.0 release; the actual bump will be semver minor)
- Changeset also needs to list all crates that depend on this crate (release manager is not dependency aware right now unfortunately)

ladvoc (approved):

PR #1421, opened 2026-09-10
Migrate the SoxResampler to be exposed via Uniffi as well as by protobuf

jhugman:
This PR mechanically refactors the `SoxResampler` FFI, to introduce uniffi bindings, and interoperability at the foreign language side to be able to get the same objects from both FFI and uniffi generated code.

This allows a gradual migration, going at different speeds in different SDKs.

The commits are well ordered:
- infrastructure needed for this type of migration
- the mechanical exposure of the existing `SoxResampler` to uniffi (barely touching the FFI `requests.rs` file
- then adding another constructor to match the existing `NewSoxResampler` message's params.


livekit-ffi/src/server/resampler.rs
```diff
@@ -15,37 +15,93 @@
...
+    pub flags: u32,
 }
 
+#[derive(uniffi::Record)]
 pub struct RuntimeSpec {
     pub num_threads: u32,
 }
 
+#[derive(uniffi::Object)]
+/// New resampler using SoX (much better quality)
```
jhugman, line 44: This duplicates the comment in the `.proto` file. It gets copied into the generated foreign bindings.

livekit-ffi/src/server/resampler.rs
```diff
@@ -15,37 +15,93 @@
...
     out_buf: Vec<i16>,
     input_rate: f64,
     output_rate: f64,
     num_channels: u32,
 }
 
-unsafe impl Send for SoxResampler {}
-
-impl SoxResampler {
+impl SoxResamplerInner {
```
jhugman, line 139: The existing implementation is almost completely untouched.

livekit-ffi/src/server/resampler.rs
```diff
@@ -159,17 +213,45 @@ impl SoxResampler {
...
+}
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum SoxResamplerError {
+    #[error("{0}")]
+    NewError(String),
+    #[error("{0}")]
+    PushError(String),
+    #[error("{0}")]
+    FlushError(String),
```
jhugman, line 321: Error messages are bare, just in case there are existing tests which rely on the current behaviour.

livekit-ffi/src/server/resampler.rs
```diff
@@ -159,17 +213,45 @@ impl SoxResampler {
...
         }
     }
 }
 
-fn to_soxr_datatype(datatype: proto::SoxResamplerDataType) -> soxr_sys::soxr_datatype_t {
-    match datatype {
-        proto::SoxResamplerDataType::SoxrDatatypeInt16i => soxr_sys::soxr_datatype_t_SOXR_INT16_I,
-        proto::SoxResamplerDataType::SoxrDatatypeInt16s => soxr_sys::soxr_datatype_t_SOXR_INT16_S,
+#[derive(uniffi::Enum)]
+// TODO(theomonnom): support other datatypes (shouldn't really be needed)
```
jhugman, line 260: This comment is copied from the proto file.

livekit-ffi/src/server/resampler.rs
```diff
@@ -159,17 +213,45 @@ impl SoxResampler {
...
-impl Drop for SoxResampler {
+impl Drop for SoxResamplerInner {
     fn drop(&mut self) {
         unsafe {
             soxr_sys::soxr_delete(self.soxr_ptr);
         }
     }
 }
 
-fn to_soxr_datatype(datatype: proto::SoxResamplerDataType) -> soxr_sys::soxr_datatype_t {
```
jhugman, line 170: This is moved to a conversion file, which implements `From<> for soxr_sys::soxr_datatype_t`

livekit-ffi/src/server/resampler.rs
```diff
@@ -159,17 +233,45 @@ impl SoxResampler {
...
-    match datatype {
-        proto::SoxResamplerDataType::SoxrDatatypeInt16i => soxr_sys::soxr_datatype_t_SOXR_INT16_I,
-        proto::SoxResamplerDataType::SoxrDatatypeInt16s => soxr_sys::soxr_datatype_t_SOXR_INT16_S,
+#[derive(uniffi::Enum)]
+// TODO(theomonnom): support other datatypes (shouldn't really be needed)
```
jhugman, line 260: This comment is copied from the proto definition.

livekit-ffi/src/migration.rs
```diff
@@ -0,0 +1,109 @@
...
+    ($T:ty) => {
+        impl crate::server::FfiHandle for ::std::sync::Arc<$T> {}
+
+        impl crate::migration::HasFfiHandleId for $T {
+            fn arc_from_self(mut self) -> ::std::sync::Arc<Self> {
+                let handle_id = self.handle_id.unwrap_or_else(|| crate::FFI_SERVER.next_id());
+                self.handle_id = Some(handle_id);
+                let arc = ::std::sync::Arc::new(self);
+                crate::FFI_SERVER.store_handle(handle_id, arc.clone());
+                arc
```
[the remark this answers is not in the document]
jhugman: Changed so the saving in the dashmap only happens when a migration from uniffi foreign side to ffi/proto on foreign side is needed. Once the foreign side has move it to the ffi/proto system, it should be responsible for calling `drop_ffi_handle()` or calling `take_ffi_handle_id()`.

livekit-ffi/src/server/resampler.rs
```diff
@@ -159,17 +233,45 @@ impl SoxResampler {
...
+pub enum SoxResamplerDataType {
+    Interleaved,
+    Split,
```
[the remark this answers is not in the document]
jhugman: This is an issue that pre-dates this migration, and should be fixed, but not as part of this migration.

livekit-ffi/src/migration_tests.rs
```diff
@@ -0,0 +1,192 @@
...
+//!
+//! The FFI request surface is the contract that must not move while the types
+//! behind it grow uniffi annotations, so these tests drive it the way the
+//! bindings do — encoded requests in, handles and raw pointers out — and check
+//! the result against the same work done through the Rust API directly. The
+//! direct-call half of each pair lives next to the code it exercises, e.g.
+//! `server::resampler::migration_tests`.
+//!
+//! They are written to compile and pass either side of the migration, so they
+//! can be replayed on the commit before it.
```
1egoman, line 25: question: Why are these in its own file and not in `migration.rs`? Are you seeing these as more of integration tests?

livekit-ffi/src/migration_tests.rs
```diff
@@ -0,0 +1,192 @@
...
+/// pushed through the FFI request surface and through the Rust API must come
+/// back as the same audio, for each rate, channel count and quality recipe.
+#[test]
+fn proto_push_matches_direct_push() {
+    let cases = [
+        (48000.0, 16000.0, 1, proto::SoxQualityRecipe::SoxrQualityQuick),
+        (48000.0, 16000.0, 1, proto::SoxQualityRecipe::SoxrQualityVeryhigh),
+        (16000.0, 48000.0, 1, proto::SoxQualityRecipe::SoxrQualityMedium),
+        (48000.0, 24000.0, 2, proto::SoxQualityRecipe::SoxrQualityHigh),
+    ];
```
1egoman, line 120: suggestion: I think it would be good to have a test which explictly exercises both the ffi method and the uniffi method on a single object, handling off from one to the other. So that might look like:
1. Create sox resampler via uniffi
2. Call `.push()`, verify it worked
3. "take" the ffi handle from the uniffi version of the sox resampler via `.take_ffi_handle_id()`
4. Call "push" via the c ffi on the same object (with the ffi handle) and make sure it works
5. Convert the ffi handle back to a uniffi object via `::from_ffi_handle_id(...)`
6. Call `.push()` via uniffi one more time and make sure it still works.

IMO this doesn't necesarily need to be done with the sox resampler, but a test which does like this should live somewhere to make sure that this core infrastructure has at least some semblance of test coverage.

1egoman (approved):
Generally looks good to me, looking forward to seeing this rolled out across the rest of the `livekit-ffi` surface area!

livekit-ffi/src/migration_tests.rs
jhugman: `migration.rs` is supposed to be infra common to all of this work.
`migration_tests.rs` is supposed to accumulate proto specific tests for each of these PRs. The point is to keep `migration.rs` small and common, and both `migration_*.rs` files as easily deletable.

PR #1425, opened 2026-09-11
Publish the Dart package without the token on dependency resolution

hiroshihorie:
### PR description

The 0.1.12 release run failed in `Validate package`: `dart pub get` got a 403 from pub.dev and the publish step was skipped. Cause is a pub.dev regression, dart-lang/pub-dev#9576: package metadata endpoints now return 403 for any request carrying an `Authorization: Bearer` header (reproducible with `curl -H 'Authorization: Bearer abc' https://pub.dev/api/packages/test`), and setup-dart registers the OIDC token for the whole host, so `pub get` sends it on every request.

Workaround, as documented in that issue:

- `Validate package` removes the token first, so `pub get` and `--dry-run` resolve without it.
- `Publish to pub.dev` re-adds the token from `PUB_TOKEN` and runs `dart pub publish --force --skip-validation`. `--skip-validation` skips dependency resolution, so the upload is the only credentialed request.

Includes a `livekit-uniffi: patch` changeset so the next release exercises the fixed path. The comment notes to collapse back to the plain commands once pub.dev fixes the issue.

### Breaking changes

None.

### Testing

actionlint clean. The 403 was reproduced locally against pub.dev with and without the header. The publish path itself can only be exercised by the next tag push.

### DIFF
--- a/.changeset/dart_publish_token_workaround.md
+++ b/.changeset/dart_publish_token_workaround.md
@@ -2,7 +2,7 @@
 livekit-uniffi: patch
 ---
 
-# Publish the Dart package without sending the publishing token on dependency resolution
+Publish the Dart package without sending the publishing token on dependency resolution
 
 pub.dev has started answering 403 to package metadata requests that carry a
 bearer token, and the publishing token was registered for the whole host, so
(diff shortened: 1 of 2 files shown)


1egoman (approved):
Interesting that this workaround is required!

hiroshihorie:
Not needed after all. pub.dev fixed the 403 on their side (dart-lang/pub-dev#9576), and re-running the 0.1.12 tag job published livekit_uniffi 0.1.12 with the unmodified workflow.

PR #1427, opened 2026-09-11
Automatic capture thread join on drop

ladvoc:
When `RunningPump` is dropped, automatically join its thread for additional safety. Prefer `RunningPump::stop_and_join` to join and get exit stats.

### DIFF
--- a/livekit-capture/src/pixel/pump.rs
+++ b/livekit-capture/src/pixel/pump.rs
@@ -239,28 +239,33 @@ mod tests {
         );
     }
 
-    #[tokio::test]
-    async fn running_pump_stops_on_signal() {
-        struct EndlessSource;
+    struct EndlessSource;
 
-        impl PixelVideoSource for EndlessSource {
-            fn resolution(&self) -> VideoResolution {
-                RESOLUTION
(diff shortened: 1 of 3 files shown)


stephen-derosa (approved):

livekit-capture/src/pixel/pump.rs
```diff
@@ -239,28 +239,33 @@ mod tests {
...
-    #[tokio::test]
-    async fn running_pump_stops_on_signal() {
-        struct EndlessSource;
+    struct EndlessSource;
 
-        impl PixelVideoSource for EndlessSource {
-            fn resolution(&self) -> VideoResolution {
-                RESOLUTION
-            }
+    impl PixelVideoSource for EndlessSource {
```
stephen-derosa, line 244: i this not just for tests anymore?

stephen-derosa (approved):

livekit-capture/src/pixel/pump.rs
ladvoc: This is still inside `mod tests`, indentation was just wrong.

PR #1434, opened 2026-09-18
fix(rtc_engine): hold a weak reference to the engine from its own task

LautaroPetaccio:
## The problem

`engine_task` took an `Arc<EngineInner>`:

```rust
tokio::spawn(Self::engine_task(inner.clone(), session_events, close_rx));
```

But the task's `JoinHandle` and the `oneshot::Sender` that stops it both live in `EngineHandle` (`rtc_engine/mod.rs:266`), which lives in `running_handle` inside that same `EngineInner`. The task therefore kept alive the only thing that could stop it — its own stop signal — and `session_events`' emitter is likewise reachable only through the session it pins.

An engine dropped without an explicit `close()` released nothing: the session, both peer connections, the `Arc<LkRuntime>` (and so the `PeerConnectionFactory` and its WebRTC threads), the signal client with its open websocket, and the task itself all stayed resident for the lifetime of the process. The SFU also kept its half of the session, since the socket was never closed.

There is no `Drop` on `RtcEngine`, `EngineInner`, `RtcSession` or `SessionInner`, so nothing else could recover it.

## The fix

`RtcSession` already avoids exactly this by keeping `close_tx` in `SessionHandle`, *outside* the `Arc`'d inner, so dropping it stops its tasks. Rather than restructure `EngineHandle`, the task now takes a `Weak` and upgrades per event.

This is what AGENTS.md already requires:

> If an owner transitively stores the callback/task, the callback/task must not strongly capture that owner.

A side benefit is that the task now stops on its own once the engine is gone, instead of depending solely on the close signal.

## Verification

`test_drop_without_close_releases_engine` connects an engine, takes a drop probe, drops it without `close()`, and waits for the internals to be released.

- Before: fails — still alive after the full 10s window.
- After: passes, in 0.03s.

Full e2e suite against `livekit-server --dev` is green, including all six reconnection tests. The only delta from the baseline is the added test.

The test targets `RtcEngine` directly rather than going through `Room`, because the room layer has the same structural issue with its own `close_tx` and would mask an engine-level fix. `test_engine()` is added to the e2e harness for that purpose.


LautaroPetaccio:
Devin reports one flag here but withheld its contents from the PR ("Not posted on this PR by your GitHub settings"), and the review comments endpoint returns none, so I cannot see what it found. If someone with access to the Devin Review dashboard pastes it, I will address it.

For what it is worth, the three flags that *were* posted on the sibling PRs in this series were all legitimate, and two of them were defects I had introduced rather than pre-existing ones — so this one is worth reading rather than dismissing.

.changeset/engine_task_weak_self_reference.md
```diff
@@ -0,0 +1,15 @@
+---
+livekit: patch
+livekit-capture: patch
```
ladvoc, line 3: ~~issue: `livekit-capture` does not need a release based on this change. Did the bot say you needed to add it?~~

Nevermind, this is correct.

ladvoc (approved):
Thanks for the contribution! Looks good to me other than one small issue with the changeset.

PR #1442, opened 2026-09-18
Cleanup unused dependencies

ladvoc:


1egoman (approved):
Looks good to me, though it might be worth verifying that you didn't miss any packages in the changeset? Based on the comment left, it seems like knope maybe needs not just dependencies but also dependencies of dependencies / etc as well?

